PR #6, opened 2018-11-20
feat: messages "info" level instead of "error"

dylandevs:
The Unity SDK has no capacity for sending anything other than an error, so I've modified the `CaptureMessage` flow to be a less serious level.


bruno-garcia:
Thanks @dylandevs

PR #12, opened 2019-01-14
feat: fixed issue where Time.time was called on non-main threads

dylandevs:
We encountered an issue where Unity would throw the error `UnityException: get_time can only be called from the main thread.` in `HandleLogCallback`. Turns out that even with the lock object, `Time.time` can't be called off the main thread. Switching `logMessageReceivedThreaded ` with `logMessageReceived` fixes the issue.


bruno-garcia:
Thanks @dylandevs !

PR #28, opened 2020-03-05
Handle LogType other than Exception

Leinnan:
It will help fix the case when we try to log error.
For example when we call ScheduleException with "Error test" as condition it will call IndexOutOfBoundsException here(SendrySdk.cs:177):
`        var exc = condition.Split(new char[] { ':' }, 2); `
`        var excType = exc[0];`
`        var excValue = exc[1].Substring(1); // strip the space`

### DIFF
--- a/Assets/Sentry/Scripts/SentrySdk.cs
+++ b/Assets/Sentry/Scripts/SentrySdk.cs
@@ -306,7 +306,7 @@ public class SentrySdk : MonoBehaviour
 #else
             string message = type.ToString() + ": " + condition;
 #endif
-            ScheduleException( message, stackTrace );
+            ScheduleException(message, stackTrace);
         }
     }
 
--- a/Assets/Sentry/Scripts/SentrySdk.cs
+++ b/Assets/Sentry/Scripts/SentrySdk.cs
@@ -297,7 +297,7 @@ public class SentrySdk : MonoBehaviour
         _timeLastError = Time.time;
         if (type == LogType.Exception)
         {
-            ScheduleException( condition, stackTrace );
+            ScheduleException (condition, stackTrace);
         }
         else
         {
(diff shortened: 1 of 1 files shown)


Assets/Sentry/Scripts/SentrySdk.cs
```diff
@@ -295,7 +295,19 @@ public void OnLogMessageReceived(string condition, string stackTrace, LogType ty
             return; // silently drop the event on the floor
         }
         _timeLastError = Time.time;
-        ScheduleException(condition, stackTrace);
+        if(type == LogType.Exception)
```
bruno-garcia, line 298:
```suggestion
        if (type == LogType.Exception)
```

Assets/Sentry/Scripts/SentrySdk.cs
```diff
@@ -295,7 +295,19 @@ public void OnLogMessageReceived(string condition, string stackTrace, LogType ty
...
+        {
+            ScheduleException( condition, stackTrace );
```
bruno-garcia, line 300:
```suggestion
            ScheduleException (condition, stackTrace);
```

Assets/Sentry/Scripts/SentrySdk.cs
```diff
@@ -295,7 +295,19 @@ public void OnLogMessageReceived(string condition, string stackTrace, LogType ty
...
+        }
+        else
+        {
+#if NET_4_6
+            string message = $"{type.ToString()}: {condition}";
+#else
+            string message = type.ToString() + ": " + condition;
+#endif
+            ScheduleException( message, stackTrace );
```
bruno-garcia, line 309:
```suggestion
            ScheduleException (message, stackTrace);
```

Assets/Sentry/Scripts/SentrySdk.cs
```diff
@@ -297,7 +297,7 @@ public void OnLogMessageReceived(string condition, string stackTrace, LogType ty
         _timeLastError = Time.time;
         if (type == LogType.Exception)
         {
-            ScheduleException( condition, stackTrace );
+            ScheduleException (condition, stackTrace);
```
bruno-garcia, line 300:
```suggestion
            ScheduleException(condition, stackTrace);
```

PR #34, opened 2020-09-28
Allow user to capture any kind of event

oktomus:
With this change, we can now capture `SentryExceptionEvent` proceduraly. Which allow better error handling.


oktomus:
I forgot something :D

bruno-garcia:
@oktomus please note we plan to use the .NET SDK for Unity.
So we likely won't be making new releases of the `lite` SDK.

PR #38, opened 2021-01-03
use relative URL for sentry-dotnet submodule

BHSPitMonkey:
Allows either HTTPS or SSH recursive cloning to work without switching protocols.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "sentry-dotnet"]
 	path = src/sentry-dotnet-unity/sentry-dotnet
-	url = ../sentry-dotnet.git
+	url = ../../sentry-dotnet.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "sentry-dotnet"]
 	path = src/sentry-dotnet-unity/sentry-dotnet
-	url = git@github.com:getsentry/sentry-dotnet.git
+	url = ../sentry-dotnet.git


bruno-garcia:
@BHSPitMonkey would changing it to http help?
The approach on this PR mean you're on your own to clone the repo sibling to this one, right? Won't work on our just `git submodule update --init` ?

BHSPitMonkey:
This should allow using either protocol, and the submodule is fetched using whichever protocol the parent repo uses. Hardcoding to HTTP probably wouldn't hurt anything, but I'm not sure it's strictly the better option

johanhelsing:
Yeah, cloning or submodule init should be no problem for getsentry/sentry-unity on either git or https.

However, this will have the implication that if you fork sentry-unity, you will also have to fork sentry-dotnet.

So if I clone my fork with `git clone --recursive https://github.com/johanhelsing/getsentry` it will fail if I don't maintain my own version of sentry-dotnet that is up-to-date enough.

Not really saying it's an issue, just thought I'd let you know. Guess it depends on how closely coupled you think the two projects are. And whether requiring people to fork both if they want to contribute using github forks (or provide instructions for how to override the url in their local clone) is worth it.

johanhelsing:
I think doing `../../getsentry/sentry-dotnet` is an option, which means I will get the getsentry version of dotnet when cloning my own sentry-unity fork.

bruno-garcia:
> I think doing `../../getsentry/sentry-dotnet` is an option, which means I will get the getsentry version of sentry-dotnet when cloning my own sentry-unity fork (as long as it's on github).

Nice!
Works:

```
➜  sentry-unity git:(main) ✗ git submodule update --init --recursive
Submodule 'sentry-dotnet' (git@github.com:getsentry/sentry-dotnet.git) registered for path 'src/sentry-dotnet-unity/sentry-dotnet'
Cloning into 'sentry-dotnet-unity/sentry-dotnet'...
Submodule path 'src/sentry-dotnet-unity/sentry-dotnet': checked out 'bbdc0280065487665561800e2e43e1ed255a9506'
➜  sentry-unity git:(main) ✗ cat .gitmodules
[submodule "sentry-dotnet"]
	path = src/sentry-dotnet-unity/sentry-dotnet
	url = ../../getsentry/sentry-dotnet.git
➜  sentry-unity git:(main) ✗
```

@johanhelsing Thanks for your input.

@BHSPitMonkey would you please change it to `../../getsentry/sentry-dotnet` then?

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "sentry-dotnet"]
 	path = src/sentry-dotnet-unity/sentry-dotnet
-	url = git@github.com:getsentry/sentry-dotnet.git
+	url = ../sentry-dotnet.git
```
bruno-garcia, line 3:
```suggestion
	url = ../../sentry-dotnet.git
```

bruno-garcia (approved):

bruno-garcia:
Done! Thanks folks!

Issue #54, opened 2021-02-10
Define testing strategy\structure


semuserable:
We need to have a testing suite for Unity Sentry package.

* check Unity instruments for testing - [Unity Test Framework](https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/index.html)
* add tests to the package\project - [Adding tests to a package](https://docs.unity3d.com/Manual/cus-tests.html)

Tyrrrz:
I have briefly looked at it, so I might be missing something. Do you think this approach is sound?

1. Create a test scene in `Sentry.Unity.Tests` that contains a single button that throws exception on click
2. Instrument it via UTF like so:

```csharp
[UnityTest]
 public IEnumerator SetupButtonTestSetupButton()
 {
     yield return PlayModeSetup();
     yield return null;
     var button = GameObject.Find("TheButton");
     button.onClick.Invoke();
}
```

3. Run UTF tests from command line like so ([reference](https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/reference-command-line.html)):

```
Unity.exe -runTests -batchmode -projectPath PATH_TO_YOUR_PROJECT -testResults C:\temp\results.xml -testPlatform PS4
```

4. On CI, use https://github.com/marketplace/actions/unity-test-runner to run the tests

Tyrrrz:
There is also [AltUnity](https://altom.gitlab.io/altunity/altunitytester/) that allows you to instrument UI tests on top of Unity, which might actually be easier than running tests as part of the game.

semuserable:
@bruno-garcia can we close this one? Seems no longer relevant as we merged both `Runtime` and `Editor` examples.

bruno-garcia:
@semuserable feel free to close issues you believe are done.

PR #56, opened 2021-02-12
Libs for coroutine support in Unity tests

semuserable:
Related to #54 

We need to support `Unity` [coroutines](https://docs.unity3d.com/Manual/Coroutines.html) in `nunit` tests. This is done via `UnityTestAttribute` which is NOT included in `nunit`. The needed infra for the attribute (and the attribute itself) is located inside `UnityEngine.TestRunner.dll`.

`UnityEngine.TestRunner.dll` is not located in convenient manner as `UnityEngine.dll`. It comes from `UPM` testing package. We need to have a reference in our `Sentry.Unity.Test` project in order to write `coroutine` tests. 

The proposed solution is to place this lib (+ `UnityEditor.TestRunner.dll` lib) inside `libs` folder of the project. The setup is inside `Sentry.Unity.Tests.csproj`.

Additionally, I added a `coroutine` test and checked it. Everything works as expected :)


src/Sentry.Unity.Tests/Sentry.Unity.Tests.csproj
```diff
@@ -8,6 +8,14 @@
         <HintPath>$(UnityPath)\UnityEngine.dll</HintPath>
         <Private>false</Private>
       </Reference>
+      <Reference Include="UnityEditor.TestRunner, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null">
+        <HintPath>libs\UnityEditor.TestRunner.dll</HintPath>
+        <Private>false</Private>
+      </Reference>
+      <Reference Include="UnityEngine.TestRunner, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null">
+        <HintPath>libs\UnityEngine.TestRunner.dll</HintPath>
```
bruno-garcia, line 16: Can't we use UPM to "restore" the package?

Or use this instead? https://www.nuget.org/packages/Troikatorz.UnityEngine.TestRunner/

semuserable: Yes, I saw this package, but I'm not sure we want to use it.

* Consistency - there is also a `UnityEditor.TestRunner` which is not on `nuget` currently.
* As far as I can tell, the given `nuget`  is just "there" and "not supported" from outside. `Unity` somehow changes\update `UnityEngine.TestRunner.dll`. We'll be adjusting to it, can't rely on the provided  `nuget` package.

One possible solution I can see for now is to "publish" both of the packages into `nuget` by ourselves. Otherwise, place them in `libs` and update when needed.

What do you think?

src/Directory.Build.props
```diff
@@ -23,4 +23,8 @@
     <UnityPath Condition="$(UnityPath) == '' AND Exists('C:\Program Files\Unity\Editor\Data\Managed\UnityEngine.dll')">C:\Program Files\Unity\Editor\Data\Managed</UnityPath>
     <UnityPath Condition="$(UnityPath) == '' AND Exists('/Applications/Unity/Unity.app/Contents/Managed/UnityEngine.dll')">/Applications/Unity/Unity.app/Contents/Managed</UnityPath>
   </PropertyGroup>
+  <PropertyGroup>
+    <UnityTestPath Condition="Exists('C:\Projects\Unity\sentry-unity\samples\unity-of-bugs\Library\ScriptAssemblies\UnityEngine.TestRunner.dll')">C:\Projects\Unity\sentry-unity\samples\unity-of-bugs\Library\ScriptAssemblies</UnityTestPath>
```
bruno-garcia, line 27: This one can go?

bruno-garcia (commented):
quick pass

src/Directory.Build.props
semuserable: It works for me, if the question about that. Sorry, didn't get it.

bruno-garcia: this assumes `C:\Projects` though? can we use a relative path instead

src/Sentry.Unity.Tests/SimpleTests.cs
```diff
@@ -1,16 +1,26 @@
...
             var unitySentryOptions = ScriptableObject.CreateInstance<UnitySentryOptions>();
             Assert.NotNull(unitySentryOptions);
         }
+
+        [UnityTest]
+        public IEnumerator UnityCoroutine_CreatedGameObjectWithDelay_Found()
+        {
+            var cube = GameObject.CreatePrimitive(PrimitiveType.Cube);
+            cube.name = "test cube";
+            yield return new WaitForSeconds(2);
```
bruno-garcia, line 22: Is the test going to take 2 seconds to pass?
If so is there another way we can test this then? WaitForMilliseconds(1) or something?

src/Directory.Build.props
semuserable: Oh, I see. I'll check.

src/Sentry.Unity.Tests/SimpleTests.cs
semuserable: We can use `yield return null` to skip a frame, but there is no `WaitForMilliseconds`. It's just a test to show that coroutines work, we probably won't need it.

bruno-garcia: I'd go with the fastest approach that still validates your test.

semuserable: Got you, will do `yield return null` then.

bruno-garcia (approved):
Nice!

PR #59, opened 2021-02-25
Editor tests

semuserable:
We need to discuss the approach I came up with.

Some highlights

* `OnInspectorUpdate` method which is called less frequently then `OnGUI`. We can potentially place all validation logic there, but if we `Debug.Log|Errror|` there we will be seeing error messages quickly emerging in the console log. It may be desired behavior for editor use.
* Unity `preprocessor directives` like `UNITY_EDITOR` are **not** available in `Sentry` projects.


src/Sentry.Unity.Editor.Tests/EditModeTests.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using NUnit.Framework;
+using UnityEditor;
+using UnityEditor.SceneManagement;
+using UnityEngine.TestTools;
+
+namespace Sentry.Unity.Editor.Tests
+{
+    public sealed class EditModeTests
```
bruno-garcia, line 10: File name too:

```suggestion
    public sealed class EditorModeTests

```

src/Sentry.Unity.Editor.Tests/EditModeTests.cs
```diff
@@ -0,0 +1,84 @@
...
+using UnityEngine.TestTools;
+
+namespace Sentry.Unity.Editor.Tests
+{
+    public sealed class EditModeTests
+    {
+        [UnityTest]
+        public IEnumerator OptionsReleaseField_LengthExceeds_CreatesError()
+        {
+            EditorHelpers.SetupScene("BugFarmScene");
```
bruno-garcia, line 15: Do we need a scene to test a Window?

src/Sentry.Unity.Editor.Tests/EditModeTests.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            var validationErrors = new List<ValidationError>();
+            window.OnValidationError += error => validationErrors.Add(error);
+
+            // act
+            window.Options.Dsn = "qwerty";
+            yield return null; // mandatory, waiting for Unity to properly redraw and update inner states
+
+            // assert
+            Assert.AreEqual(1, validationErrors.Count);
```
bruno-garcia, line 28: Ideally we'd validate the actual error too because it could easily have 1 error but be unrelated and test would pass.

src/Sentry.Unity.Editor.Tests/Sentry.Unity.Editor.Tests.csproj
```diff
@@ -0,0 +1,39 @@
...
+      <Private>false</Private>
+    </Reference>
+    <Reference Include="UnityEngine.TestRunner">
+      <HintPath>$(UnityTestPath)\UnityEngine.TestRunner.dll</HintPath>
+      <Private>false</Private>
+    </Reference>
+  </ItemGroup>
+  <ItemGroup>
+    <!-- Unity uses exactly this (3.5.0) version as of now -->
+    <PackageReference Include="NUnit" Version="3.5.0" />
```
bruno-garcia, line 26: We could control this version on the main `Directory.Build.Props`.

Or ideally like we do in .NET where the test projects are under `test` so there's a D.B.P there:

https://github.com/getsentry/sentry-dotnet/blob/main/test/Directory.Build.props

bruno-garcia (commented):
Besides the stuff we discussed on our call I left a couple small points

src/Sentry.Unity.Editor.Tests/EditModeTests.cs
semuserable: +

src/Sentry.Unity.Editor.Tests/EditModeTests.cs
semuserable: No, I checked, will be removed.

src/Sentry.Unity.Editor.Tests/EditModeTests.cs
semuserable: I agree.

src/Sentry.Unity.Editor.Tests/Sentry.Unity.Editor.Tests.csproj
semuserable: Ok, I'll check!

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -138,4 +156,20 @@ private void OnGUI()
             // EditorGUILayout.EndToggleGroup();
         }
     }
+
+    public readonly struct ValidationError
+    {
+        public readonly string PropertyName;
```
bruno-garcia, line 162: 10+ years of brainwashing that public fields, even of readonly are evil make me have a hard time looking at this. Probably never to be a problem as it's not a public API that would break consumers if we needed to encapsulate later but curious, why not throw in a getter?
 ```suggestion
        public readonly string PropertyName { get; };
```

bruno-garcia (approved):
Looks awesome! Thanks!

src/Sentry.Unity.Editor/SentryWindow.cs
semuserable: It's `readonly struct`. From my perspective, it doesn't need to have a getter.

I see that you have a very strong opinion about that. If you want, I'll change it to a getter.

bruno-garcia: Oh I don't have a very strong opinion, I even merged the PR :)
I was trying to word it even in a joke (10 years of brainwashin).

I can't argue why add a getter to a readonly struct so lgtm

Issue #68, opened 2021-03-11
Improve the log handler debouncer


bruno-garcia:
Currently we only allow N number of calls per second. This mechanism needs to be aware of the log entry that's going through so that a spammy Debug log doesn't keep out error events.

semuserable:
merged https://github.com/getsentry/sentry-unity/pull/69

Issue #66, opened 2021-03-09
Add a clear path to opt-out of the SentryWindow approach to manual initialization


bruno-garcia:
For advanced scenarios folks will want to init the SDK programatically.

To support such scenarios, we need to include a clear:

```csharp
var options = new SentryUnityOptions();
SentryUnity.Init(options);
```

The [code done initializing](https://github.com/getsentry/sentry-unity/blob/main/src/Sentry.Unity/SentryInitialization.cs#L60) today was good enough for using SentryWindow. We need to pull out into `SentryUnity` to become the API used by customers when they opt-out from the `SentryWindow` and get access to the complete .NET API/options

This means our auto init should also go through such API.

**edit by semuserable:** we need to force `link.xml` to be created if user decides to use programmatic approach. Current approach is to copy it from `Packages/<sentry-plugin-folder>/Runtime/link.xml` into `Assets/Resources/Sentry/link.xml`. Check https://github.com/getsentry/sentry-unity/issues/84.

The approach is taken by Xamarin:
https://github.com/getsentry/sentry-xamarin/blob/63e764393415324f7fa0356e62d679d64f8d34ff/Src/Sentry.Xamarin/SentryXamarin.cs#L32-L42

semuserable:
I guess that basically means to move out the current logic from `SentryInitialization.Init` into a separate class and "make it more composable"?

bruno-garcia:
Yes. The Editor window configuration can have a toggle to turn off auto init (I believe I already added one) and we can document the user how they can initialize programatically.

We need to consider: Does that mean all the configuration set via the SentryWindow doesn't apply?
Does Unity provide some sort of hook for us to invoke a user-defined method passing our `SentryUnityOptions` instance as an argument? Similar to a Spring Bean, a service locator or some IoC container we can resolve stuff from?

semuserable:
Started looking into it. I have a question. What is the proper way of disposing of `SentrySdk`? Can I rely on `SentrySdk.Init()` `IDisposable` or use `SentrySdk.Close()`?

Looking at the approach of replacing one `SentrySdk` with another. What is the recommended way?

bruno-garcia:
> Started looking into it. I have a question. What is the proper way of disposing of `SentrySdk`? Can I rely on `SentrySdk.Init()` `IDisposable` or use `SentrySdk.Close()`?
> 
> Looking at the approach of replacing one `SentrySdk` with another. What is the recommended way?

[...]

PR #69, opened 2021-03-11
New debouncer logic

semuserable:


bruno-garcia:
Please try to name the PR in a way that it fits in the UI. It also becomes the commit message once we squash merge it

semuserable:
If tests are run in bulk, 2 of them fails for now. I'll check tomorrow. It seems because of how everything is static.

I haven't updated `package` folder with latest libs. Let's discuss first.

samples/unity-of-bugs/Assets/Scripts/DebounceTester.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class DebounceTester : MonoBehaviour
```
bruno-garcia, line 5: If this is just for testing (not for demo purposes), can this live in the Tests project instead?

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -22,7 +22,8 @@ public static class SentryInitialization
         // Temp event capture infra
         public static IEventCapture EventCapture = new EventCapture();
 
-        private static readonly IUnityLogMessageFilter _typeConditionFilter = new TypeConditionFilter();
+        private static readonly ErrorTimeDebounce _errorTimeDebounce = new();
```
bruno-garcia, line 25: We'll need to refactor these `static` fields later.
Whatever we're doing not on the method annotated with `RuntimeInitializeOnLoadMethod` will need to become callable by user-code, optionally. If they decided to opt-out from this auto-init.

The "Integration" that will hook into the logger callback will make usage of these instances, so they might be passed as a dependency to such integration. Today everything is just hacked on this file.

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -140,7 +140,13 @@ private static void OnLogMessageReceived(string condition, string stackTrace, Lo
             // TODO: 'options' not used yet
             _ = options;
 
-            if (!_typeConditionFilter.Debounced(type, condition))
+            var debounced = type switch
+            {
+                LogType.Error or LogType.Exception or LogType.Assert => _errorTimeDebounce.Debounced(),
```
bruno-garcia, line 145: `or`? TIL

src/Sentry.Unity/TimeDebounceBase.cs
```diff
@@ -0,0 +1,48 @@
...
+            return true;
+        }
+    }
+
+    /// <summary>
+    /// This class is not thread-safe and is designed to be called by Unity non-threaded logger callback
+    /// </summary>
+    internal sealed class LogTimeDebounce : TimeDebounceBase
+    {
+        public LogTimeDebounce() => _debounceOffset = TimeSpan.FromSeconds(1);
```
bruno-garcia, line 38: For now this is fine but these intervals will need to be configurable.
It'll be simpler to have a single type that takes an interval via the ctor.

bruno-garcia (approved):
Some notes but we can merge this and move on

samples/unity-of-bugs/Assets/Scripts/DebounceTester.cs
semuserable: It was needed for my demos. I'll remove it, not needed.

src/Sentry.Unity/SentryInitialization.cs
semuserable: 100%! I already started looking into https://github.com/getsentry/sentry-unity/issues/66, my hope it will be covered there)

src/Sentry.Unity/TimeDebounceBase.cs
semuserable: 100%! My thoughts too. Need to decide what value do we expect (millis, seconds etc)

semuserable:
closes #68

PR #71, opened 2021-03-12
Split integration and unit tests

bruno-garcia:
Get tests to run without Unity


src/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <LangVersion>9</LangVersion>
     <Nullable>enable</Nullable>
-    <UnityVersion>2019.4.21f1</UnityVersion>
+    <UnityVersion>2019.4.19f1</UnityVersion>
```
semuserable, line 5: Hm..

semuserable:
You branch doesn't work for me. In order to not waste time (as it's still WIP on your side), I decided to give it a go on my own. I created `net48` runnable test project and put some simple test. It errors out during the run
image

It's strange because I'm reusing the same `D.B.p` as for other tests. Hm...

semuserable:
Ok, resolved that by excluding `<Private>false</Private>`, but now another error.

image

semuserable:
Ok, probably resolved that!

image

semuserable:
So, the difference that I noticed, is we need to have a different `D.B.p` for `unit test` project with "privates" removed.

semuserable:
WTF?

image

semuserable:
This is so strange.

image

I thought it errors out because some Unity api is being called, but I used `UnityEngine.LogType` in the first test and it's totally ok with that! Why is it wrong for `UnityEventProcessor`? Hm..

semuserable:
image

Ok, now it seems we're calling very specific Unity API and it doesn't like that. Let's find out!

semuserable:
This call `Application.version` fails with `System.Security.SecurityException : ECall methods must be packaged into a system module.`

Investigation continued..

semuserable:
`Time.realtimeSinceStartup` fails with `ECall`

```
[NativeProperty("Realtime")]
public static extern float realtimeSinceStartup { [MethodImpl(MethodImplOptions.InternalCall)] get; }
```

semuserable:
`Application.installMode` fails with `ECall`

```csharp
public static extern ApplicationInstallMode installMode { [FreeFunction("GetApplicationInfo().GetInstallMode"), MethodImpl(MethodImplOptions.InternalCall)] get; }
```

semuserable:
`Debug.isDebugBuild` fails with `ECall`

```csharp
[NativeProperty(TargetType = TargetType.Field)]
[StaticAccessor("GetBuildSettings()", StaticAccessorType.Dot)]
public static extern bool isDebugBuild { [MethodImpl(MethodImplOptions.InternalCall)] get; }
```

semuserable:
As you can see, native `Unity` calls are not supported. If we want to test such logic in `unit test` manner, then we should remove all native calls.

What do you think? @bruno-garcia

src/Directory.Build.props
bruno-garcia: Only 22f1 shows up for install now for me and I don't have 21f1.

I'd revert this before mering.

bruno-garcia:
> As you can see, native `Unity` calls are not supported. If we want to test such logic in `unit test` manner, then we should remove all native calls.
> 
> What do you think? @bruno-garcia

Thanks for investigating it further. Unity has a fork of Mono so it might not work at all even with Mono (which `dotnet test` on a `net48` on macOS run on).

Might be just too much work to get it to work. I want to take a stab a this before we give up though

semuserable:
Noticed another strange thing. After opening up the solution, I've started to receive those weird errors

image

bruno-garcia:
Seems like no chance to get this to work.

bruno-garcia:
I mean for types that are pure C# (no dependency to Unity types) it works, but then I'm not sure it's worth adding.
We could consider if a lot of such types come in and things become too slow as they are

PR #72, opened 2021-03-17
Release libs updated

semuserable:
**THE RELEASE**

* ~3k lines should not intimidate you :), I added `Samples~` project for 'release'
* `WarningTimeDebounce` added & tests updated
* release package updated with the latest libs

I encourage you to try it first via this branch, before approving and merging. Here is the link which you can use for installing it via UPM: `https://github.com/getsentry/sentry-unity.git?path=/package#feature/release-upd`


src/Sentry.Unity/TimeDebounceBase.cs
```diff
@@ -45,4 +45,12 @@ internal sealed class ErrorTimeDebounce : TimeDebounceBase
     {
         public ErrorTimeDebounce(TimeSpan debounceOffset) => _debounceOffset = debounceOffset;
     }
+
+    /// <summary>
+    /// This class is not thread-safe and is designed to be called by Unity non-threaded logger callback
+    /// </summary>
+    internal sealed class WarningTimeDebounce : TimeDebounceBase
```
bruno-garcia, line 52: Can we get this in through a PR first?
Lets drop the release part of this PR and do it through the process we agreed on #75

bruno-garcia (changes requested):

Issue #75, opened 2021-03-17
Release process


bruno-garcia:
`craft prepare`:

* [x] On the machine running craft we'll bump Directory.Build.props' `<Version>` ([running script bump](https://github.com/getsentry/sentry-dotnet/blob/main/scripts/bump-version.sh))
* [x] [Bump Unity package version](https://github.com/getsentry/sentry-unity/blob/be02a7c202949fb51809dac07b6cb2a3763e6bb0/package/package.json#L3) from `package.json` in that new folder
* [x] pushes branch release/version
* [x] In CI: .NET project build in release mode 
* [x] In CI: copy the `package.json` file from the release folder (`package`) into some folder that will become the artifact
* [x] In CI: Copy `samples/` into `package/Samples~/` (which files to copy to figured out)
* [x] Copy everything from `package-dev` except `package.json` and `Tests` into that new folder.
* [x] Create a GitHub artifact out of that new folder called after the commit sha just built: https://github.com/getsentry/sentry-dotnet/blob/2337261333d5743996d6d96ca3aa289d34a9d3ec/.github/workflows/build.yaml#L58

`craft publish` downloads artifacts 
* [ ] Run script to validate the package
* [x] Replace the contents of `package` directory with the contents of the artifact.

Also:

* [x] Use release registry to keep track of versions
* [x] Delete the install instructions from the readme.
* [x] Point users to the documentation for installation instruction

The version will be updated on the docs automatically.

bruno-garcia:
Raised an issue to figure out the `craft publish` step:
https://github.com/getsentry/craft/issues/194

bruno-garcia:
Until the craft target is tackled, after `craft publish` the following release steps are done:

```bash
export release=0.0.6
export artifact=where/craft/downloaded/artifact
git checkout git@github.com:getsentry/unity.git
cd unity
git rm -r .
unzip $artifact/package-release.zip -d .
git add .
git commit -m "release $release"
git tag $release
git push origin main --tags
```

The release notes from this repo should be copied over to the other, on the tag of the same name.

PR #79, opened 2021-03-18
Prepare release artifact on CI

Tyrrrz:
Related to #75

I don't understand this step:

> Bump Unity package version from package.json in that new folder

What should be the version be bumped to? Shouldn't it be done locally, not on CI, when running the bump script? It can bump both the `Directory.Build.props` version as well as `package.json` version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,13 +75,13 @@ jobs:
 
       - name: Build project
         run: |
-          unity -quit -batchmode -nographics -logFile - -projectPath samples/unity-of-bugs -buildWindows64Player artifacts/build/game.exe | Out-Default
+          unity -quit -batchmode -nographics -logFile - -projectPath samples/unity-of-bugs -buildWindows64Player ../../artifacts/build/game.exe | Out-Default
 
       - name: Upload build artifacts
         uses: actions/upload-artifact@v1
         with:
           name: Build output
-          path: samples/unity-of-bugs/artifacts/build
+          path: artifacts/build
 
       # The working directory must be set because the project uses
       # relative paths for certain things.
@@ -97,20 +97,20 @@ jobs:
       # Note: this command doesn't work with the `-quit` option, unlike all the other ones.
       - name: Run Unity tests (playmode)
         run: |
-          unity -batchmode -nographics -logFile - -runTests -testPlatform PlayMode -projectPath samples/unity-of-bugs -testResults artifacts/test/playmode/results.xml | Out-Default
+          unity -batchmode -nographics -logFile - -runTests -testPlatform PlayMode -projectPath samples/unity-of-bugs -testResults ../../artifacts/test/playmode/results.xml | Out-Default
           $LASTEXITCODE = 0
 
       - name: Upload test artifacts (playmode)
         uses: actions/upload-artifact@v1
         with:
           name: Test results (playmode)
-          path: samples/unity-of-bugs/artifacts/test/playmode
+          path: artifacts/test/playmode
 
       # Running tests does not print anything useful to the console
       # so we need to parse the results to figure out how things went.
       - name: Parse test results (playmode)
         run: |
-          $testRun = Select-Xml -Path samples/unity-of-bugs/artifacts/test/playmode/results.xml -XPath "/test-run" | Select-Object -Last 1
+          $testRun = Select-Xml -Path artifacts/test/playmode/results.xml -XPath "/test-run" | Select-Object -Last 1
 
           Write-Output "Playmode test run completed in $($testRun.Node.duration) seconds."
           Write-Output "-- Result: $($testRun.Node.result)"
@@ -135,18 +135,18 @@ jobs:
 
       - name: Run Unity tests (editmode)
         run: |
-          unity -batchmode -nographics -logFile - -runTests -testPlatform EditMode -projectPath samples/unity-of-bugs -testResults artifacts/test/editmode/results.xml | Out-Default
+          unity -batchmode -nographics -logFile - -runTests -testPlatform EditMode -projectPath samples/unity-of-bugs -testResults ../../artifacts/test/editmode/results.xml | Out-Default
           $LASTEXITCODE = 0
 
       - name: Upload test artifacts (editmode)
         uses: actions/upload-artifact@v1
         with:
           name: Test results (editmode)
-          path: samples/unity-of-bugs/artifacts/test/editmode
+          path: artifacts/test/editmode
 
       - name: Parse test results (editmode)
         run: |
-          $testRun = Select-Xml -Path samples/unity-of-bugs/artifacts/test/editmode/results.xml -XPath "/test-run" | Select-Object -Last 1
+          $testRun = Select-Xml -Path artifacts/test/editmode/results.xml -XPath "/test-run" | Select-Object -Last 1
 
           Write-Output "Editmode test run completed in $($testRun.Node.duration) seconds."
           Write-Output "-- Result: $($testRun.Node.result)"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -76,13 +75,13 @@ jobs:
 
       - name: Build project
         run: |
-          unity -quit -batchmode -nographics -logFile - -projectPath ${{ env.PROJECT_PATH }} -buildWindows64Player artifacts/build/game.exe | Out-Default
+          unity -quit -batchmode -nographics -logFile - -projectPath samples/unity-of-bugs -buildWindows64Player artifacts/build/game.exe | Out-Default
```
bruno-garcia, line 78: isn't it better to have this path as a variable? Why walk away from it?

bruno-garcia (commented):
It's nice this is landing in CI but I wonder if these can become scripts in `scripts/*.ps1`?

I'm afraid this will become hard to test/fix going forward if we have to rely on CI. I rather we run these locally to develop and just use CI to invoke them later

bruno-garcia (approved):
Changing to scripts can be a follow up PR

bruno-garcia:
Though the artifact has 600MB so need to check that first

.github/workflows/ci.yaml
Tyrrrz: it's actually simpler for me when i see the path directly.
it would make sense if the project path changed, but otherwise it's just clearer this way imo.

Tyrrrz:
> It's nice this is landing in CI but I wonder if these can become scripts in `scripts/*.ps1`?
> 
> I'm afraid this will become hard to test/fix going forward if we have to rely on CI. I rather we run these locally to develop and just use CI to invoke them later

Yeah it's possible if there's a need to run them locally. Moving the steps into separate files will make it harder to reason about though.

Tyrrrz:
> Though the artifact has 600MB so need to check that first

Yeah it's because it copies the artifacts generated by the build too :/

I moved the prepare release steps on a separate job so that it uses an isolated file system. It doesn't need Unity there anyway.

Tyrrrz:
ok now it's 148kb

bruno-garcia:
> ok now it's 148kb

Awesome! Thanks!

bruno-garcia:
> > It's nice this is landing in CI but I wonder if these can become scripts in `scripts/*.ps1`?
> > I'm afraid this will become hard to test/fix going forward if we have to rely on CI. I rather we run these locally to develop and just use CI to invoke them later
> 
> Yeah it's possible if there's a need to run them locally. Moving the steps into separate files will make it harder to reason about though.

It does have the downside of now reading the steps the script take by reading the yml file, but with the advantage of being easier to test/debug/develop locally the script.

I for one want to run this locally before I run a release via GHA

.github/workflows/ci.yaml
```diff
@@ -175,3 +174,26 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.3
+        with:
+          submodules: recursive
+
+      - name: Prepare Sentry package for release
+        run: |
+          New-Item "package-release" -ItemType Directory
+          Copy-Item "package-dev/*" -Destination "package-release/" -Exclude "Tests", "Tests.meta", "package.json" -Recurse
```
bruno-garcia, line 191: Is this grabbing the files that were written here in the workflow above? Doesn't seem like these are related builds

Tyrrrz: yeah i didn't consider that it puts something there 😅

Tyrrrz:
@bruno-garcia i dropped `Samples~/` because that folder is 300mb even without the final build artifacts.
The package is now 2mb (800kb compressed).

Good?

Tyrrrz:
Also cache restore took 3 minutes last time 💨  That's pretty fast. Too bad the build itself takes like 15 minutes.

semuserable:
@Tyrrrz how it `package/Samples~` folder is 300mb? Are you copying it directly from `samples/unity-of-bugs`? That's not right. It seems you're taking everything that's inside `samples/unity-of-bugs` including `Libraries`, `caches` etc.

`package/Samples~` should contain just `Assets/Scenes` and `Assets/Scripts` with corresponding meta files, nothing else. `Library`, `Packages`, `ProjectSettings` should **not** be included.

I guess there is a little misunderstanding of `package/Sample~` folder. Use-case is the following: user adds Sentry via UPM, in package UI user clicks `import sample project`, the files will be extracted into **already opened** unity project. Now you can open scenes, scripts etc.

.github/workflows/ci.yaml
```diff
@@ -170,6 +169,18 @@ jobs:
             exit 1
           }
 
+      - name: Prepare Sentry package for release
+        run: |
+          New-Item "package-release" -ItemType Directory
+          Copy-Item "package-dev/*" -Destination "package-release/" -Exclude "Tests", "Tests.meta", "package.json" -Recurse
```
semuserable, line 175: `*.asmdf`, `*.asmdf.meta`  files from `Runtime` and `Editor` should not be copied, as they contain `dev` endings in file names and config itself.

```json
{
	"name": "io.sentry.unity.dev.editor"
}
```

Tyrrrz: Other `.meta` files are fine?

Tyrrrz:
> package/Samples~ should contain just Assets/Scenes and Assets/Scripts with corresponding meta files, nothing else. Library, Packages, ProjectSettings should not be included.

So it should like this right?

```
/Samples~/unity-of-bugs/
│   README.md
│
└───Assets
    ├───Scenes
    │       BugFarmScene.unity
    │       BugFarmScene.unity.meta
    │       EmptyScene.unity
    │       EmptyScene.unity.meta
    │
    └───Scripts
            BugFarm.cs
            BugFarm.cs.meta
```

Btw, should it go directly to `Samples~` or to `Samples~/unity-of-bugs`?

semuserable:
@Tyrrrz pretty much, yeah. It should go into `Samples~/unity-of-bugs` because `UPM` can support several test projects. Please check, the following https://github.com/getsentry/sentry-unity/pull/72/files#diff-5205bfc4f58dc7ca1e3a58905757b4332c9e784666f28cadea2b4a1dd4b1e41d

As you can see `package/package.json` contains description of `Samples~/*` projects.

Tyrrrz:
> @Tyrrrz pretty much, yeah. It should go into `Samples~/unity-of-bugs` because `UPM` can support several example projects. Please check https://github.com/getsentry/sentry-unity/pull/72/files#diff-5205bfc4f58dc7ca1e3a58905757b4332c9e784666f28cadea2b4a1dd4b1e41d (my release PR)
> 
> As you can see `package/package.json` contains description of `Samples~/*` projects.
> 
> ```json
> "samples": [
>   {
>     "displayName": "unity-of-bugs",
>     "description": "Sentry SDK example project",
>     "path": "Samples~/unity-of-bugs"
>   }
> ]
> ```

Thanks, I will change it.

Tyrrrz:
@semuserable it's 2mb now. Can you download and check if it's alright?

semuserable:
@Tyrrrz PR looks ok for me now. But what should I download exactly?

semuserable (approved):

Tyrrrz:
> @Tyrrrz PR looks ok for me now. But what should I download exactly?

Steps:

1. 
image

2.
image

Tyrrrz:
Updated one more time after discussion with @semuserable

semuserable:
@Tyrrrz Nice one! Thanks)

Looks good)

bruno-garcia (approved):
Nice!!

> Bump Unity package version from package.json in that new folder

We need to figure out here but sounds like we should consider this elsewhere anyway than the release step

PR #80, opened 2021-03-19
Json for Sentry config

semuserable:
No `ScriptableObject` for Sentry config. Quick & dirty. `UnitySentryOptions` reused.

* tested on
  * editor
  * Win mono build 
* issues
  *  json doesn't work on `IL2CPP` Win build although I added it to the `lnks.xml`, have I missed something?
  * build for android is around 16m, but until `IL2CPP` works, haven't tried it yet

### DIFF
--- a/src/Sentry.Unity.Editor/SentryWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryWindow.cs
@@ -65,7 +65,7 @@ namespace Sentry.Unity.Editor
                 var emptyOptions = JsonSerializer.Serialize(UnitySentryOptionsDefault, _jsonOptions);
                 File.WriteAllText(SentryOptionsJsonPathFull, emptyOptions);
 
-                // Must be called, otherwise Unity won't be able to load it with the nex call. *.meta should be created for new file
+                // Must be called, otherwise Unity won't be able to load it with the next call. *.meta should be created for new file
                 AssetDatabase.Refresh();
             }
 
@@ -126,6 +126,8 @@ namespace Sentry.Unity.Editor
 
             var text = JsonSerializer.Serialize(Options, _jsonOptions);
             File.WriteAllText(SentryOptionsJsonPathFull, text);
+            // Write is not enough for Unity, must update its asset database.
+            AssetDatabase.Refresh();
 
             /*AssetDatabase.SaveAssets();
             // TODO: This should be gone
--- a/src/Sentry.Unity.Editor/SentryWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryWindow.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Text.Json;
 using UnityEditor;
 using UnityEngine;
 
@@ -15,16 +16,27 @@ namespace Sentry.Unity.Editor
 
         protected string SentryOptionsAssetPath => $"Assets/Resources/Sentry/{SentryOptionsAssetName}.asset";
 
+        protected string SentryOptionsJsonPath => $"Sentry/SentryOptionsJson";
+
+        // Will be used only from Unity Editor
+        protected string SentryOptionsJsonPathFull => $"{Application.dataPath}/Resources/{SentryOptionsJsonPath}.json";
+
         public UnitySentryOptions Options { get; set; } = null!; // Set by OnEnable()
 
         public event Action<ValidationError> OnValidationError = _ => { };
+
+        private readonly JsonSerializerOptions _jsonOptions = new()
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase
+        };
 
         private void OnEnable()
         {
             SetTitle();
 
-            Options = AssetDatabase.LoadAssetAtPath<UnitySentryOptions>(SentryOptionsAssetPath);
-            if (Options is null)
+            Options = LoadSentryJsonConfig();
+
+            /*if (Options is null)
             {
                 Options = CreateInstance<UnitySentryOptions>();
                 if (!AssetDatabase.IsValidFolder("Assets/Resources"))
@@ -38,7 +50,28 @@ namespace Sentry.Unity.Editor
                 AssetDatabase.CreateAsset(Options, SentryOptionsAssetPath);
             }
 
-            EditorUtility.SetDirty(Options);
+            EditorUtility.SetDirty(Options);*/
+        }
+
+        private UnitySentryOptions LoadSentryJsonConfig()
+        {
+            if (!File.Exists(SentryOptionsJsonPathFull))
+            {
+                var UnitySentryOptionsDefault = new UnitySentryOptions
+                {
+                    Dsn = "https://94677106febe46b88b9b9ae5efd18a00@o447951.ingest.sentry.io/5439417",
+                    Enabled = true
+                };
+                var emptyOptions = JsonSerializer.Serialize(UnitySentryOptionsDefault, _jsonOptions);
+                File.WriteAllText(SentryOptionsJsonPathFull, emptyOptions);
+
+                // Must be called, otherwise Unity won't be able to load it with the nex call. *.meta should be created for new file
+                AssetDatabase.Refresh();
+            }
+
+            // We should use `TextAsset` for read-only access in runtime. It's platform agnostic.
+            var sentryOptionsTextAsset = Resources.Load<TextAsset>(SentryOptionsJsonPath);
+            return JsonSerializer.Deserialize<UnitySentryOptions>(sentryOptionsTextAsset.text, _jsonOptions)!;
         }
 
         private void SetTitle()
@@ -90,9 +123,13 @@ namespace Sentry.Unity.Editor
         public void OnLostFocus()
         {
             Validate();
-            AssetDatabase.SaveAssets();
+
+            var text = JsonSerializer.Serialize(Options, _jsonOptions);
+            File.WriteAllText(SentryOptionsJsonPathFull, text);
+
+            /*AssetDatabase.SaveAssets();
             // TODO: This should be gone
-            AssetDatabase.Refresh();
+            AssetDatabase.Refresh();*/
         }
 
         // ReSharper disable once UnusedMember.Local
--- a/src/Sentry.Unity/SentryInitialization.cs
+++ b/src/Sentry.Unity/SentryInitialization.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Text.Json;
 using UnityEngine;
 using UnityEngine.SceneManagement;
 using CompressionLevel = System.IO.Compression.CompressionLevel;
@@ -26,15 +28,24 @@ namespace Sentry.Unity
         internal static LogTimeDebounce LogTimeDebounce = new(TimeSpan.FromSeconds(1));
 
         internal static bool IsInit { get; private set; }
+        private static readonly JsonSerializerOptions _jsonOptions = new()
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase
+        };
+
+        private static string SentryOptionsJsonPath => $"Sentry/SentryOptionsJson";
 
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void Init()
         {
-            if (!(Resources.Load("Sentry/SentryOptions") is UnitySentryOptions options))
+            var sentryOptionsTextAsset = Resources.Load<TextAsset>(SentryOptionsJsonPath);
+            var options = JsonSerializer.Deserialize<UnitySentryOptions>(sentryOptionsTextAsset.text, _jsonOptions)!;
+
+            /*if (!(Resources.Load("Sentry/SentryOptions") is UnitySentryOptions options))
             {
                 Debug.LogWarning("Sentry Options asset not found. Did you configure it on Component/Sentry?");
                 return;
-            }
+            }*/
 
             if (!options.Enabled)
             {
--- a/src/Sentry.Unity/link.xml
+++ b/src/Sentry.Unity/link.xml
@@ -28,6 +28,8 @@
         <type fullname="System.ComponentModel.UInt32Converter" preserve="all"/>
         <type fullname="System.ComponentModel.UInt64Converter" preserve="all"/>
     </assembly>
+    
+    <assembly fullname="System.Text.Json" preserve="all"/>
 
     <!--https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.lambdaexpression.name?view=netframework-4.7.2&viewFallbackFrom=netframework-2.0-->
     <assembly fullname="System.Core">
--- a/src/test/Sentry.Unity.Editor.Tests/EditorModeTests.cs
+++ b/src/test/Sentry.Unity.Editor.Tests/EditorModeTests.cs
@@ -9,7 +9,8 @@ namespace Sentry.Unity.Editor.Tests
 {
     public sealed class EditorModeTests
     {
-        [UnitySetUp]
+        // Disabled (json approach)
+        // [UnitySetUp]
         public IEnumerator InitializeOptions()
         {
             // Due to an issue, Sentry doesn't always load UnitySentryOptions, which
--- a/src/test/Sentry.Unity.Tests/PlayModeTests.cs
+++ b/src/test/Sentry.Unity.Tests/PlayModeTests.cs
@@ -11,7 +11,8 @@ namespace Sentry.Unity.Tests
 {
     public sealed class PlayModeTests
     {
-        [UnitySetUp]
+        // Disabled (json approach)
+        // [UnitySetUp]
         public IEnumerator InitializeOptions()
         {
             // Due to an issue, Sentry doesn't always load UnitySentryOptions, which


semuserable:
Please notice, currently the json config generated dynamically and **not** included in the project.

samples/unity-of-bugs/Assets/Resources/Sentry/link.xml
```diff
@@ -0,0 +1,65 @@
+﻿<linker>
+    <assembly fullname="Sentry" preserve="all"/>
+    <assembly fullname="Sentry.Unity" preserve="all"/>
+    <assembly fullname="System.Text.Json" preserve="all"/>
```
bruno-garcia, line 4: We will for sure need to fine tune this but for now it's OK
https://github.com/getsentry/sentry-unity/issues/81

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -38,7 +50,28 @@ private void OnEnable()
...
+            EditorUtility.SetDirty(Options);*/
+        }
+
+        private UnitySentryOptionsJson LoadSentryJsonConfig()
+        {
+            if (!File.Exists(SentryOptionsJsonPathFull))
+            {
+                var unitySentryOptionsJson = new UnitySentryOptionsJson
+                {
+                    Dsn = "https://94677106febe46b88b9b9ae5efd18a00@o447951.ingest.sentry.io/5439417",
```
bruno-garcia, line 62: We need to `#ifdef` or something so the test DSN only gets included in dev mode for the SDK. Meaning it shouldn't go into the unity package

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -38,7 +50,28 @@ private void OnEnable()
...
+                    Enabled = true
+                };
+                var emptyOptions = JsonSerializer.Serialize(unitySentryOptionsJson, _jsonOptions);
+                File.WriteAllText(SentryOptionsJsonPathFull, emptyOptions);
```
bruno-garcia, line 66: Isn't there a `Stream` API we can use instead of getting a `string` allocated?

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -38,7 +50,28 @@ private void OnEnable()
...
+
+                // Must be called, otherwise Unity won't be able to load it with the next call. *.meta should be created for new file
+                AssetDatabase.Refresh();
+            }
+
+            // We should use `TextAsset` for read-only access in runtime. It's platform agnostic.
+            var sentryOptionsTextAsset = Resources.Load<TextAsset>(SentryOptionsJsonPath);
+            return JsonSerializer.Deserialize<UnitySentryOptionsJson>(sentryOptionsTextAsset.text, _jsonOptions)!;
```
bruno-garcia, line 74: We could avoid  reading/deserializing here, no?

```suggestion
            return unitySentryOptionsJson;
            } else {

            // We should use `TextAsset` for read-only access in runtime. It's platform agnostic.
            var sentryOptionsTextAsset = Resources.Load<TextAsset>(SentryOptionsJsonPath);
            return JsonSerializer.Deserialize<UnitySentryOptionsJson>(sentryOptionsTextAsset.text, _jsonOptions)!;
            }

```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -90,9 +123,16 @@ private void ValidateDsn()
         public void OnLostFocus()
         {
             Validate();
-            AssetDatabase.SaveAssets();
-            // TODO: This should be gone
+
+            // Saving `UnitySentryOptions` and not `UnitySentryOptionsJson` for OnGUI backward-compat
```
bruno-garcia, line 127: We don't need to be backwards compatible here since there hasn't been a single release.

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -26,15 +28,24 @@ public static class SentryInitialization
         internal static LogTimeDebounce LogTimeDebounce = new(TimeSpan.FromSeconds(1));
 
         internal static bool IsInit { get; private set; }
+        private static readonly JsonSerializerOptions _jsonOptions = new()
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase
+        };
+
+        private static string SentryOptionsJsonPath => $"Sentry/SentryOptionsJson";
```
bruno-garcia, line 36: Why `Json` in the name of the file? We just assume `SentryOptions` is JSON and if it fails to deserialize we log an error out to the console so the user can manually inspect `SentryOptions` and delete it. Or raise an issue here explaining how that happened.

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -26,15 +28,24 @@ public static class SentryInitialization
...
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase
+        };
+
+        private static string SentryOptionsJsonPath => $"Sentry/SentryOptionsJson";
 
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void Init()
         {
-            if (!(Resources.Load("Sentry/SentryOptions") is UnitySentryOptions options))
+            var sentryOptionsTextAsset = Resources.Load<TextAsset>(SentryOptionsJsonPath);
```
bruno-garcia, line 41: I wonder if it would make sense to load this at runtime into the `DLL`, as a static string allocation to avoid doing the file I/O in the UI thread slowing down the startup.

Something to consider for the future? https://github.com/getsentry/sentry-unity/issues/82

src/Sentry.Unity/UnitySentryOptions.cs
```diff
@@ -12,6 +12,18 @@ public enum SentryUnityCompression
         Fastest = 2,
         NoCompression = 3
     }
+
+    // We need to think carefully about what types should be placed here, because of ILL2CPP stripping
+    public sealed class UnitySentryOptionsJson
```
bruno-garcia, line 17: Can we drop this type and have only `SentryUnityOptions` though?

src/Sentry.Unity/link.xml
```diff
@@ -1,5 +1,7 @@
 ﻿<linker>
     <assembly fullname="Sentry" preserve="all"/>
+    <assembly fullname="Sentry.Unity" preserve="all"/>
```
bruno-garcia, line 3: We have this file and one on the sample project. 
Shouldn't this file be the source of truth? And somehow get added to the project at build/time?

bruno-garcia (commented):
Looking great already!
I left some notes to discuss and suggestions for the future

src/Sentry.Unity/link.xml
semuserable: We should add this file via `Sentry.Unity`. For now, I included it in both places just for the **demo**, we need to think about how and when to copy it.

src/Sentry.Unity/UnitySentryOptions.cs
semuserable: This is what I tried first, but because of some types in `SentryUnityOptions` itself, `IL2CPP` builds were failing, that's why I included a comment. I need to check which field(-s) exactly messes up `IL2CPP` and what additional lib should be included into `link.xml`. I decided to no waste too much time here, as `IL2CPP` build are **slow**.

src/Sentry.Unity/SentryInitialization.cs
semuserable: But we need to change it via Unity window. I have no clue for now.

src/Sentry.Unity/SentryInitialization.cs
semuserable: Because it's `quick-dirty-example-of-json-as-config` :)

src/Sentry.Unity.Editor/SentryWindow.cs
semuserable: `quick-dirty-example-of-json-as-config`

samples/unity-of-bugs/Assets/Resources/Sentry/link.xml
semuserable: Sure.

src/Sentry.Unity.Editor/SentryWindow.cs
semuserable: Streams are only in `*Async` version of `JsonSerializer`.

src/Sentry.Unity.Editor/SentryWindow.cs
semuserable: You're right, missed that. Will be changed.

src/Sentry.Unity.Editor/SentryWindow.cs
semuserable: I don't think we discussed that yet. Possible point for discussion #83

src/Sentry.Unity.Editor/SentryWindow.cs
semuserable: Do you want it like this?

```csharp
using var fileStream = new FileStream(SentryOptionsJsonPathFull, FileMode.Create);
using var writer = new Utf8JsonWriter(fileStream);
writer.WriteStartObject();
writer.WriteString(nameof(unitySentryOptionsJson.Dsn), unitySentryOptionsJson.Dsn);
writer.WriteBoolean(nameof(unitySentryOptionsJson.Enabled), unitySentryOptionsJson.Enabled);
writer.WriteEndObject();
writer.Flush();
```

src/Sentry.Unity.Editor/SentryWindow.cs
bruno-garcia: We can't have that hardcoded here, I'm aware we can't use unify's compiler flag since we ship the compiled DLLs. We could add a single script though that's on source control for us to use but it's not moved into the packages folders

src/Sentry.Unity.Editor/SentryWindow.cs
bruno-garcia: yeah that's the approach on the .NET SDK, without reflection. A bit more work though

src/Sentry.Unity/SentryInitialization.cs
bruno-garcia: I see. But since this isn't a draft PR I assume it was the form you were proposing we merge in

src/Sentry.Unity/SentryInitialization.cs
bruno-garcia: The Unity Window will write the asset. But at build time we read the asset and write code :hack-for-perf:.

Future ideas, not to worry until we need to find optimizations for faster startup

src/Sentry.Unity/UnitySentryOptions.cs
bruno-garcia: I see. Yeah IL2CPP is extremely slow. I believe these builds should be a lot faster on the latest Unity version. So perhaps for dev it's best to use something latest. CI we can continue on an older one to make sure we're compatible.

src/Sentry.Unity.Editor/AndroidManifestConfiguration.cs
```diff
@@ -1,3 +1,4 @@
+/*
```
bruno-garcia, line 1: accident?

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -155,6 +160,80 @@ private void OnGUI()
...
+        /// Find and copy 'link.xml' into current Unity project for IL2CPP builds
+        /// </summary>
+        private static void TryCopyLinkXml()
+        {
+            const string linkXmlFileName = "link.xml";
+
+            var linkXmlPath = $"{Application.dataPath}/Resources/{UnitySentryOptions.ConfigRootFolder}/{linkXmlFileName}";
+            if (File.Exists(linkXmlPath))
+            {
+                // TODO: should we log here?
```
bruno-garcia, line 190: at the "debug" level it would be nice.
`else` could log in debug level: no link.xml exists yet, creating one.

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -155,6 +160,80 @@ private void OnGUI()
...
+                return;
+            }
+
+            var linkPath = GetLinkXmlPath(linkXmlFileName);
+            if (linkPath == null)
+            {
+                // TODO: should we log here?
```
bruno-garcia, line 197: likely a log.fatal yes. We're going to have a broken SDK here, right?

src/Sentry.Unity/UnitySentryOptions.cs
```diff
@@ -12,32 +13,147 @@ public enum SentryUnityCompression
         Fastest = 2,
         NoCompression = 3
     }
+
     [Serializable]
```
bruno-garcia, line 17: I assume we don't need this either now?

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -155,6 +160,80 @@ private void OnGUI()
...
+                if (assetName.Contains("Runtime"))
+                {
+                    var linkFolderPath = ExtractLinkFolderPathFromEnd(assetName);
+                    return $"{linkFolderPath}/{linkXmlFileName}";
+                }
+            }
+
+            return null;
+
+            static string ExtractLinkFolderPathFromEnd(string value)
```
bruno-garcia, line 223: `Path.GetFileName` ?

src/Sentry.Unity/UnitySentryOptions.cs
```diff
@@ -12,32 +13,147 @@ public enum SentryUnityCompression
...
+
+            if (!string.IsNullOrWhiteSpace(Dsn))
+            {
+                writer.WriteString("dsn", Dsn);
+            }
+
+            writer.WriteBoolean("debug", Debug);
+            writer.WriteBoolean("debugOnlyInEditor", DebugOnlyInEditor);
+            writer.WriteString("diagnosticsLevel", DiagnosticsLevel.ToString().ToLowerInvariant());
+            writer.WriteString("requestBodyCompressionLevel", RequestBodyCompressionLevel.ToString());
```
bruno-garcia, line 68: I think best is to serialize the numeric representation of all enums. It'll version better (if we add new enum values we can always deserialize it even if that value is't defined on the new enum.

```suggestion
            writer.WriteString("diagnosticsLevel", DiagnosticsLevel.ToString("d"));
            writer.WriteString("requestBodyCompressionLevel", RequestBodyCompressionLevel.ToString("d"));
```

bruno-garcia (commented):
The approach is great! Will also perform better and since we're anyway dependant on S.T.J might as well use it.

Good thing now we can take advantage of that nice testing infrastrucutre you've setup to get these code tested. Will be particularly nice to make sure the serialization round-trip works. Would also be good to have some snapshot tests to validate we're not introducing breakage on serialzied files, as we ship new versions

src/Sentry.Unity/UnitySentryOptions.cs
semuserable: Sure!

src/Sentry.Unity.Editor/AndroidManifestConfiguration.cs
semuserable: Nope, I just commented the whole file until we need it. It uses some old parts, I just see no reason to change it now.

src/Sentry.Unity.Editor/SentryWindow.cs
semuserable: I'd say it depends :) `link.xml` is not needed for `Mono` builds, but for `IL2CPP` only. I'd not treat it as `fatal`, but merely `warning`.

src/Sentry.Unity.Editor/SentryWindow.cs
semuserable: Ah.. Yeah, I need `Path.GetDirectoryName()` in this case. Will change)

src/Sentry.Unity/UnitySentryOptions.cs
semuserable: Yes, I was gonna ask you the same question, because this was my initial idea too.

src/Sentry.Unity/UnitySentryOptions.cs
```diff
@@ -12,32 +13,146 @@ public enum SentryUnityCompression
...
-        [field: SerializeField] public bool CaptureInEditor { get; set; } = true; // Lower entry barrier, likely set to false after initial setup.
-        [field: SerializeField] public string? Dsn { get; set; }
-        [field: SerializeField] public bool Debug { get; set; } = true; // By default on only
-        [field: SerializeField] public bool DebugOnlyInEditor { get; set; } = true;
-        [field: SerializeField] public SentryLevel DiagnosticsLevel { get; set; } = SentryLevel.Error; // By default logs out Error or higher.
+        /// <summary>
+        /// Relative to Assets/Resources
+        /// </summary>
+        public const string ConfigRootFolder = "Sentry";
+        public const string ConfigName = "SentryOptions";
```
bruno-garcia, line 23:
```suggestion
        /// <summary> file named and version controled options set by the sueasdad
        internal const string ConfigName = "SentryOptions";
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -155,6 +160,67 @@ private void OnGUI()
...
+            var linkXmlAsset = AssetDatabase.LoadAssetAtPath<TextAsset>(linkPath);
+            File.WriteAllBytes(linkXmlPath, linkXmlAsset.bytes);
+        }
+
+        /// <summary>
+        /// Get Unity path to 'link.xml' file from `Packages` folder
+        /// </summary>
+        private static string? GetLinkXmlPath(string linkXmlFileName)
+        {
+            var assetIds = AssetDatabase.FindAssets("io.sentry.unity", new [] { "Packages" });
```
bruno-garcia, line 211: Best to lift this `io.sentry.unity` elsewhere so we can follow the reference and document what this means and effects of changing it.
```suggestion
            var assetIds = AssetDatabase.FindAssets("io.sentry.unity", new [] { "Packages" });

```

src/test/Sentry.Unity.Editor.Tests/EditorModeTests.cs
```diff
@@ -55,5 +58,18 @@ public void OptionsDsnField_WrongFormat_CreatesError()
...
+        public void SentryTestWindow_OpenAndLinkXmlCopied_Successful()
+        {
+            LogAssert.ignoreFailingMessages = true; // mandatory
+
+            // Open & Close window to trigger 'link.xml' logic
+            SentryTestWindow.Open().Dispose();
+
+            var linkXmlPath = $"{Application.dataPath}/Resources/{UnitySentryOptions.ConfigRootFolder}/link.xml";
+            Assert.IsTrue(File.Exists(linkXmlPath));
+        }
```
bruno-garcia, line 73: What about we make sure it wasn't there to begin with. Or should we delete it before the test runs?

```suggestion
        public void SentryTestWindow_OpenAndLinkXmlCopied_Successful()
        {
            var linkXmlPath = $"{Application.dataPath}/Resources/{UnitySentryOptions.ConfigRootFolder}/link.xml";
            Assert.IsFalse(File.Exists(linkXmlPath)); // sanity check
            LogAssert.ignoreFailingMessages = true; // mandatory

            // Open & Close window to trigger 'link.xml' logic
            SentryTestWindow.Open().Dispose();

            Assert.IsTrue(File.Exists(linkXmlPath));
        }

```

src/test/Sentry.Unity.Tests/UnitySentryOptionsTest.cs
```diff
@@ -0,0 +1,80 @@
...
+        [Test]
+        public void Options_ReadFromFile_Success()
+        {
+            var optionsFilePath = GetTestOptionsFilePath();
+            Assert.IsTrue(File.Exists(optionsFilePath));
+
+            var jsonRaw = File.ReadAllText(optionsFilePath);
+            using var jsonDocument = JsonDocument.Parse(jsonRaw);
+
+            UnitySentryOptions.FromJson(jsonDocument.RootElement);
```
bruno-garcia, line 23: This is missing the asserts I assume?

bruno-garcia (approved):
It's good to be merged though, the other notes could be a follow up PR.

Thanks!

PR #87, opened 2021-03-25
craft

bruno-garcia:


bruno-garcia:
@Tyrrrz I guess we'll need a bump-version that bumps also `package/package.json` together with the Directory.Build.props.

Since everything is done in powershell, for as long as it runs on a mac it would work. Could u give a hand with the script here plz?

Tyrrrz:
```powershell
param([string] $newVersion)

function Replace-TextInFile {
    param([string] $filePath, [string] $pattern, [string] $replacement)

    $content = [IO.File]::ReadAllText($filePath)
    $content = [Text.RegularExpressions.Regex]::Replace($content, $pattern, $replacement)
    [IO.File]::WriteAllText($filePath, $content, [Text.Encoding]::UTF8)
}

Replace-TextInFile "$PSScriptRoot/src/Directory.Build.props" '(?<=<Version>)(.*?)(?=</Version>)' $newVersion
Replace-TextInFile "$PSScriptRoot/package/package.json" '(?<="version": ")(.*?)(?=")' $newVersion
```

This uses regex so not super flexible, but it keeps things simple and realistically will never break. Usage: `./<script>.ps1 1.2.3` where `1.2.3` is the new version.

Note that there currently isn't a `<Version>` property in `Directory.Build.props`.

PR #113, opened 2021-04-19
Release package samples

### LINKED ISSUE #112: No samples available in UPM package (version 0.0.6)
When installed, no samples available, although we have [Samples~ inside](https://github.com/getsentry/unity/tree/main/Samples~).
image

In order to fix that we must add `samples` property into `release` [package.json](https://github.com/getsentry/sentry-unity/blob/main/package/package.json)

```json
"samples": [
   {
     "displayName": "Unity of bugs",
     "description": "Bug scene to test Sentry on",
     "path": "Samples~/unity-of-bugs"
   }
]
```

I know that I added this before. I guess, because of release process changing rapidly, it's got lost.

The question is, how we should include it into `package.json`? It should **not** be taken from `package-dev` (there is no `Samples~` there). Can we do it directly or some release scripts will overwrite it?

cc @bruno-garcia

semuserable:
closes #112

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Strip zeroes for ill2cpp builds (#108)
 - Proper sdk name reporting for sentry event (#111)
+- Release package samples (#113)
 
 ## 0.0.6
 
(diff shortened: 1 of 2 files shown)


semuserable:
license...

PR #118, opened 2021-04-19
use readme from package dir

bruno-garcia:
The idea here is to use the `README.md` from the 'package template directory'

Possibly better idea is to just copy all files from `package` since it's supposed to be the template directory now (i.e: All files in it will go into the package as-is)

#skip-changelog


semuserable (approved):

semuserable:
Ok, after another thought, here is Unity Package Layout
image

Do you mean `README.md` or `Documentation~` folder for the docs? In `UPM` there is a distinction.

bruno-garcia:
Both. Lets add the README in this PR

Issue #119, opened 2021-04-20
NullReference on Play before opening Tools/Sentry


bitsandfoxes:
Unity 2021.1.1f1
Sentry included via PackageManager -> Git

Sentry UnitySDK 0.0.7

### Steps to Reproduce

1. Import Sentry
2. Press Play
3. Get NullReferenced

### Expected Result

Not get NullReferenced

### Actual Result

NullReferenceException: Object reference not set to an instance of an object
Sentry.Unity.UnitySentryOptions.LoadFromUnity () (at D:/a/sentry-unity/sentry-unity/src/Sentry.Unity/UnitySentryOptions.cs:114)
Sentry.Unity.SentryInitialization.Init () (at D:/a/sentry-unity/sentry-unity/src/Sentry.Unity/SentryInitialization.cs:34)

semuserable:
We should first check if config exists before trying to read it. Check `SentryWindow.LoadUnitySentryOptions`.

bitsandfoxes:
closed with https://github.com/getsentry/sentry-unity/pull/125

PR #125, opened 2021-04-22
#119 check for options file and disable logger if not found

bitsandfoxes:
We check for the config file before we try to load them.
If not found -> Print out "What to do" for the user and skip Sentry initilization.

#skip-changelog

### DIFF
--- a/src/Sentry.Unity.Editor/SentryTestWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryTestWindow.cs
@@ -14,7 +14,7 @@ namespace Sentry.Unity.Editor
         public void Dispose()
         {
             Close(); // calls 'OnLostFocus' implicitly
-            File.Delete(SentryOptionsAssetPath);
+            File.Delete(UnitySentryOptions.GetConfigPath());
             AssetDatabase.Refresh();
         }
     }
--- a/src/Sentry.Unity.Editor/SentryTestWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryTestWindow.cs
@@ -14,7 +14,7 @@ namespace Sentry.Unity.Editor
         public void Dispose()
         {
             Close(); // calls 'OnLostFocus' implicitly
-            File.Delete(UnitySentryOptions.GetConfigPath());
+            File.Delete(UnitySentryOptions.GetConfigPath(SentryOptionsAssetName));
             AssetDatabase.Refresh();
         }
     }
--- a/src/Sentry.Unity/SentryInitialization.cs
+++ b/src/Sentry.Unity/SentryInitialization.cs
@@ -32,9 +32,9 @@ namespace Sentry.Unity
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void Init()
         {
-            if (!File.Exists($"{Application.dataPath}/Resources/{UnitySentryOptions.ConfigRootFolder}/{UnitySentryOptions.ConfigName}.json"))
+            if (!File.Exists(UnitySentryOptions.GetConfigPath()))
             {
-                Debug.LogWarning("Sentry Logging has been disabled.\nSentry has not been configured yet. Please go to Tools/Sentry...");
+                Debug.LogWarning("Couldn't find the configuration file SentryOptions.json. Did you already configure Sentry?\nYou can do that through the editor: Tools -> Sentry");
                 return;
             }
 
--- a/src/Sentry.Unity/SentryInitialization.cs
+++ b/src/Sentry.Unity/SentryInitialization.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.SceneManagement;
@@ -31,6 +32,12 @@ namespace Sentry.Unity
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void Init()
         {
+            if (!File.Exists($"{Application.dataPath}/Resources/{UnitySentryOptions.ConfigRootFolder}/{UnitySentryOptions.ConfigName}.json"))
+            {
+                Debug.LogWarning("Sentry Logging has been disabled.\nSentry has not been configured yet. Please go to Tools/Sentry...");
+                return;
+            }
+
             var options = UnitySentryOptions.LoadFromUnity();
 
             if (!options.Enabled)
--- a/src/Sentry.Unity/UnitySentryOptions.cs
+++ b/src/Sentry.Unity/UnitySentryOptions.cs
@@ -26,6 +26,9 @@ namespace Sentry.Unity
         /// </summary>
         public const string ConfigName = "SentryOptions";
 
+        public static string GetConfigPath(string? notDefaultConfigName = null)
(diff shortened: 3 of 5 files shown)


src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -31,6 +32,12 @@ public static class SentryInitialization
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void Init()
         {
+            if (!File.Exists($"{Application.dataPath}/Resources/{UnitySentryOptions.ConfigRootFolder}/{UnitySentryOptions.ConfigName}.json"))
```
bruno-garcia, line 35: Is this path being built elsewhere? In a way that we can call a function to get it instead of building it in different locations (potentially going out of sync)?

semuserable:
You should add changelog entry as the `Danger` suggested :)

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -31,6 +32,12 @@ public static class SentryInitialization
...
+            {
+                Debug.LogWarning("Sentry Logging has been disabled.\nSentry has not been configured yet. Please go to Tools/Sentry...");
```
bruno-garcia, line 37: If I understood correctly this will happen if the `SentryWindow` was never open, and therefore there was never a json file created.

Maybe we can mitigate this by creating the json file somehow.
Nonetheless we could reword the messageL
```suggestion
                Debug.LogDebug("Couldn't find the configuration file SentryOptions.json. Did you already configure Sentry?.\nYou can do that through the editor: Tools -> Sentry");
```

bruno-garcia (commented):
Thanks for fixing this Stefan!
Is it possible to have some tests to validate this?

src/Sentry.Unity/SentryInitialization.cs
semuserable, line 35: I'd move the common part into something like `UnitySentryOptions.GetConfigPath()` and reused it inside `SentryWindow.cs`

SentryWindow.cs
```csharp
protected string SentryOptionsAssetPath => UnitySentryOptions.GetConfigPath(SentryOptionsAssetName)
```

SentryInitialization.cs
```csharp
if(!File.Exists(UnitySentryOptions.GetConfigPath()))
```

and `UnitySentryOptions.GetConfigPath` can be written like

UnitySentryOptions.cs
```csharp
public static GetConfigPath(string? notDefaultConfigName)
  => $"{Application.dataPath}/Resources/{ConfigRootFolder}/{notDefaultConfigName ?? ConfigName}.json"
```

What do you think?

src/Sentry.Unity/SentryInitialization.cs
semuserable: Btw, forcing json creation sounds reasonable. We will be sure it's always created from our side. The logic about creation is inside `SentryWindow.OnEnable()`.

bruno-garcia: We should look into creating it at package install time but we still need to do this validation to avoid a NullRefException if somehow the file isn't there. But in that case we could use `Debug.LogError` even to alert somethign went wrong.

bitsandfoxes: When you say "forcing" do you mean to force the user to open the window to create the json or that we somehow do that without user intervention?

src/Sentry.Unity/SentryInitialization.cs
bitsandfoxes: I like it. The only problem I see is that UnitySentryOptions is centred around Resource right now and has no access to Application.dataPath and would return a relative path. But we should probably look for a different location to store options anyway.

bitsandfoxes: I like it! I'll. get right on it.

src/Sentry.Unity/SentryInitialization.cs
bitsandfoxes: > We should look into creating it at package install time but we still need to do this validation to avoid a NullRefException if somehow the file isn't there. But in that case we could use `Debug.LogError` even to alert somethign went wrong.

I think we can hook into the AssetImportPipeline to get notified of the package gets installed/updated.

src/Sentry.Unity.Editor/SentryTestWindow.cs
```diff
@@ -14,7 +14,7 @@ public static SentryTestWindow Open()
         public void Dispose()
         {
             Close(); // calls 'OnLostFocus' implicitly
-            File.Delete(SentryOptionsAssetPath);
+            File.Delete(UnitySentryOptions.GetConfigPath());
```
semuserable, line 17: You kinda broke tests :) You must use `SentryTestWindow.SentryOptionsAssetName` path.

src/Sentry.Unity/UnitySentryOptions.cs
```diff
@@ -26,6 +26,9 @@ public sealed class UnitySentryOptions
         /// </summary>
         public const string ConfigName = "SentryOptions";
 
+        public static string GetConfigPath(string? notDefaultConfigName = null)
```
bruno-garcia, line 29: If this can be internal, it should so we can move it around without breaking consumers
```suggestion
        internal static string GetConfigPath(string? notDefaultConfigName = null)
```

bitsandfoxes: If they are internal the UnityWindow won't be able to reach it, right?

bitsandfoxes:
#skip-changelog

src/Sentry.Unity/SentryInitialization.cs
bitsandfoxes: We'll revisit the automated config-file generation in a separate issue. https://github.com/getsentry/sentry-unity/issues/136#issue-871121703

bruno-garcia:
I think `#skip-changelog` must be put on the PR description, just adding to a comment won't do.
I'll try

src/Sentry.Unity/UnitySentryOptions.cs
bruno-garcia: We could make `InternalsVisibleTo` for `Sentry.Unity.Editor` so we can share API between the two without committing to public API to users of the Sentry Unity SDK

src/Sentry.Unity/UnitySentryOptions.cs
```diff
@@ -1,9 +1,12 @@
 using System;
 using System.IO;
+using System.Runtime.CompilerServices;
 using System.Text.Json;
 using Sentry.Extensibility;
 using UnityEngine;
 
+[assembly: InternalsVisibleTo("Sentry.Unity.Editor")]
```
bruno-garcia, line 8: We already have this here:

https://github.com/getsentry/sentry-unity/blob/fd4a26e239d18865f1c947e3551178a1cde00120/src/Sentry.Unity/Properties/AssemblyInfo.cs#L3

semuserable:
image
This issues is still not resolved.

src/Sentry.Unity.Editor/SentryTestWindow.cs
semuserable, line 17: Still not resolved.

semuserable (changes requested):

PR #126, opened 2021-04-23
bump dotnet 3.3.2

bruno-garcia:
Technically the same changes we've shipped on 0.0.8 but this time checking out the tag as we agreed to keep track off 

No changelog since no relevant change (the .NET bump is already in the changelog)
#skip-changelog


semuserable (approved):

bruno-garcia:
@semuserable the markdown file copy is failing here too. Seems flaky for some reason

semuserable:
I don't understand. `Documentation~` is only in `main`. Why it tries to find it in other branches? Do you have any idea where to look?

PR #130, opened 2021-04-23
Unity Sentry SDK programmatic setup

semuserable:
Initial attempt of programmatic setup, there are a lot to cover.

**Current:**

* Only one `SentryUnity` can exist, all subsequent inits will dispose old one and replace with a new one.

Example from `SentryInitialization.Init`:
```csharp
// options from Sentry Window, DSN = https://94677106febe46b88b9b9ae5efd18a00@o447951.ingest.sentry.io/5439417
SentryUnity.Init(options);

Debug.LogError("Error!"); // sent to https://94677106febe46b88b9b9ae5efd18a00@o447951.ingest.sentry.io/5439417

// previous disposed, new one created
SentryUnity.Init(new UnitySentryOptions
{
    Enabled = true,
    Dsn = "YOUR_DSN"
});

Debug.LogError("Error next!"); // sent to 'YOUR_DSN'
```

All of the above are checked on `il2cpp` builds. 

* Tests rewritten. Now uses `SentryUnity`

**Things to consider:**

* `link.xml` is not implemented for this approach yet
* Unity options should be renamed to be consistent with Sentry dotnet SDKs
* Should notify a user (more gracefully then now) that a previous `SentryUnity` is disposed (removed/destroyed) and a new one is created. My thoughts: retrieve call site (stack? frame? CallMember?) of the `Init` methods and store it somewhere, then check and compare when new `Init` happens -> notify user.
* More use cases?

Let's discuss further steps.


src/Sentry.Unity/SentryUnity.cs
```diff
@@ -0,0 +1,140 @@
+﻿using System;
+using System.Collections.Generic;
+using Sentry.Extensibility;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace Sentry.Unity
+{
+    public sealed class SentryUnity : IDisposable
```
bruno-garcia, line 9: Instead of having this class be disposable, can we make the return of `Init` be?
It's the way `Sentry` static works so it's easier to reason about it.

https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/SentrySdk.cs#L81

Also, calling Init multiple times isn't just we support. We it's nice that we dispose one and create a new one (so events are flushed on the old one first) but this is not something users should rely on.

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -0,0 +1,140 @@
...
+        internal readonly WarningTimeDebounce WarningTimeDebounce = new(TimeSpan.FromSeconds(1));
+
+        internal UnitySentryOptions? Options { get; private set; }
+
+        internal IEventCapture EventCapture { get; set; } = new EventCapture();
+
+        public static SentryUnity Init(UnitySentryOptions unitySentryOptions, Action<SentryUnity>? unityConfigure = null)
+        {
+            // Force dispose if new Init
+            if (Instance != null)
```
bruno-garcia, line 24: This is racy. If `Init` is called from different threads this wouldn't work well.

Check Sentry .NET:

https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/SentrySdk.cs#L102-L104

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            var unitySentryOptions = new UnitySentryOptions();
+            unitySentryOptionsConfigure.Invoke(unitySentryOptions);
+            return Init(unitySentryOptions, unityConfigure);
+        }
+
+        private SentryUnity WithApplicationEvents()
+        {
+            Application.logMessageReceived += OnLogMessageReceived;
+            Application.quitting += OnQuitting;
```
bruno-garcia, line 54: Not sure these private methods are really worth it. I wonder if they get inlined with IL2CPP.

I think it's best we move this stuff to a static internal class. This class has enough to do.

```suggestion
            Application.logMessageReceived += OnLogMessageReceived;
            Application.quitting += OnQuitting;
```

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        private SentryUnity FinalizeSetup(IDiagnosticLogger? diagnosticLogger)
+        {
+            var data = SceneManager.GetActiveScene().name is { } name
+                ? new Dictionary<string, string> {{"scene", name}}
+                : null;
+
+            SentrySdk.AddBreadcrumb("BeforeSceneLoad", data: data);
+
+            diagnosticLogger?.Log(SentryLevel.Debug, "Complete Sentry SDK initialization.");
```
bruno-garcia, line 67: Is this because we don't have the extensions in the .NET SDK available here?

Purists will cry but lets `InternalsVisibleTo` to this assembly and use those. And that Json extensions copied over.
Assembly will get smaller too :)

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            return this;
+        }
+
+        private void OnLogMessageReceived(string condition, string stackTrace, LogType type)
```
bruno-garcia, line 72: This one I meant can live on its own class like: `UnityApplicationLoggingIntegration` and could implement `ISdkIntegration`.

We add it like:

```
Sentry.Init(o => 
  o.Dsn = "";
  o.AddIntegration(new UnityApplicationLoggingIntegration(o);
);
```

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -0,0 +1,140 @@
...
+                };
+        }
+
+        private void OnQuitting()
+        {
+            // Note: iOS applications are usually suspended and do not quit. You should tick "Exit on Suspend" in Player settings for iOS builds to cause the game to quit and not suspend, otherwise you may not see this call.
+            //   If "Exit on Suspend" is not ticked then you will see calls to OnApplicationPause instead.
+            // Note: On Windows Store Apps and Windows Phone 8.1 there is no application quit event. Consider using OnApplicationFocus event when focusStatus equals false.
+            // Note: On WebGL it is not possible to implement OnApplicationQuit due to nature of the browser tabs closing.
+            Application.logMessageReceived -= OnLogMessageReceived;
```
bruno-garcia, line 129: This line would be on the `Unregister` method of `UnityApplicationLoggingIntegration`.

src/Sentry.Unity/UnitySentryOptions.cs
```diff
@@ -91,6 +92,27 @@ public void WriteTo(Utf8JsonWriter writer)
             writer.Flush();
         }
 
+        public SentryOptions ToSentryOptions()
```
bruno-garcia, line 95: I'm not sure this is the best way for us to go.

Maybe it's time we inherit from `SentryOptions`? We don't need to serialize the whole thing. We serialize what we use from `SentryWindow` only

src/test/Sentry.Unity.Tests/PlayModeTests.cs
```diff
@@ -157,23 +135,15 @@ private static IEnumerator SetupSceneCoroutine(string sceneName)
...
-        }
+        private static void InitSentrySdk(Action<SentryUnity>? unityConfigure = null)
+            => SentryUnity.Init(
+                opt =>
+                {
+                    opt.Enabled = true;
+                    opt.Dsn = "https://94677106febe46b88b9b9ae5efd18a00@o447951.ingest.sentry.io/5439417";
+                    opt.Logger = new UnityLogger(SentryLevel.Warning);
+                },
+                unityConfigure);
```
bruno-garcia, line 146: Snippets are usually formatted like:
```suggestion
            => SentryUnity.Init(o =>
                {
                    o.Enabled = true; // Do we need this? Should be true by default.
                    o.Dsn = "https://94677106febe46b88b9b9ae5efd18a00@o447951.ingest.sentry.io/5439417";
                    o.Logger = new UnityLogger(SentryLevel.Warning);
                },
                unityConfigure);

```

Alternatively use the full `option` as opposed to `o` though it's far from uncommon the single letter variable name for the lambda argument and keeps the snippet short.

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -51,147 +28,14 @@ public static void Init()
...
-                LogType.Error or LogType.Exception or LogType.Assert => ErrorTimeDebounce.Debounced(),
-                LogType.Log => LogTimeDebounce.Debounced(),
-                LogType.Warning => WarningTimeDebounce.Debounced(),
-                _ => true
-            };
-            if (!debounced)
-            {
-                return;
-            }
+            SentryUnity.Init(options);
```
bruno-garcia, line 31: We'll need an opt-out I guess. So probably need to read the SentryOptions json to check for a flag to decide whether to init here or if the user is going to do it programatically.

src/Sentry.Unity/Extensions/SentryOptionsExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+                ? release
+                : Application.version;
+
+        public static void ConfigureEnvironment(this SentryOptions sentryOptions, UnitySentryOptions unitySentryOptions)
+            => sentryOptions.Environment = unitySentryOptions.Environment is { } environment
+                ? environment
+                : Application.isEditor
+                    ? "editor"
+                    // TODO: should we remove #if?
+#if DEVELOPMENT_BUILD
```
bruno-garcia, line 32: We'll need this to get through via a script that compiles with Unity to use their compiler flags.

src/Sentry.Unity/UnitySentryOptions.cs
semuserable: We probably should after looking at other SDKs. This will unblock `AddIntegration` facility.

src/Sentry.Unity/SentryUnity.cs
semuserable: I'm not sure I understood you. This part was before, I just moved it here.

src/Sentry.Unity/SentryInitialization.cs
semuserable: Let's discuss.

src/Sentry.Unity/SentryUnity.cs
semuserable: Let's discuss.

src/Sentry.Unity/SentryUnity.cs
bruno-garcia: Example:

https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Integrations/AppDomainUnhandledExceptionIntegration.cs

https://github.com/getsentry/sentry-dotnet/blob/main/test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Internal/AppDomainAdapter.cs

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                return;
+            }
+
+            // TODO: to check against 'MinBreadcrumbLevel'
+            if (type != LogType.Error && type != LogType.Exception && type != LogType.Assert)
+            {
+                // TODO: MinBreadcrumbLevel
+                // options.MinBreadcrumbLevel
+                SentrySdk.AddBreadcrumb(condition, level: ToBreadcrumbLevel(type));
```
bruno-garcia, line 45: You take a `Hub` also through `Register` so there's no need to call the static `SentrySdk` here.
This was done to ease testing (check Sentry .NET tests for integrations)

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -0,0 +1,87 @@
...
+                return;
+            }
+
+            var sentryEvent = new SentryEvent(new UnityLogException(condition, stackTrace));
+            sentryEvent.SetTag("log.type", ToEventTagType(type));
+            _ = _eventCapture?.Capture(sentryEvent);
```
bruno-garcia, line 51: You can validate this through mocking `IHub` instead:

https://github.com/getsentry/sentry-dotnet/blob/d25c97f0e5c77b520e30877444e93a9147afb9cd/test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs#L36

src/Sentry.Unity/Integrations/UnityBeforeSceneLoadIntegration.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public void Register(IHub hub, SentryOptions options)
+        {
+            var data = SceneManager.GetActiveScene().name is { } name
+                ? new Dictionary<string, string> {{"scene", name}}
+                : null;
+
+            SentrySdk.AddBreadcrumb("BeforeSceneLoad", data: data);
+
+            options.DiagnosticLogger?.Log(SentryLevel.Debug, "Complete Sentry SDK initialization.");
```
bruno-garcia, line 17: This log message is confusing, what about:

```suggestion
            options.DiagnosticLogger?.Log(SentryLevel.Debug, "Registered BeforeSceneLoad integration.");
```

Arent' we supposed to hook into an event to get notified when a scene loads?
This will register the scene name when the SDK initializes and never again, right?

semuserable:
What do you think about this updated window?
image

Main points

* Disable Programmatic Initialization
* Compression level UI based on `SentryOptions` - `System.IO.Compression.CompressionLevel`

semuserable:
Checked `InternalsVisibleTo` on `Sentry.Unity`
image

As `Sentry` is signed, we must also sign `Sentry.Unity` if we want to reuse internals.

cc @bruno-garcia

semuserable:
Green, yay!

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using Sentry.Integrations;
+using Sentry.Unity.Extensions;
+using Sentry.Unity.Integrations;
+
+namespace Sentry.Unity
+{
+    public sealed class SentryUnity
```
bruno-garcia, line 8:
```suggestion
    public static class SentryUnity

```

src/Sentry.Unity/Integrations/IAppDomain.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace Sentry.Unity.Integrations
+{
+    internal interface IAppDomain
```
bruno-garcia, line 7: Should this be `IApplication` instead?

The .NET SDK is creating an abstraction over a class called `AppDomain` that's why `IAppDomain` there.

src/Sentry.Unity/Integrations/IAppDomain.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Sentry.Unity.Integrations
+{
+    internal interface IAppDomain
+    {
+        event Application.LogCallback LogMessageReceived;
+        event Action Quitting;
+        string ActiveSceneName { get; }
+    }
+
+    internal sealed class UnityAppDomain : IAppDomain
```
bruno-garcia, line 14:
```suggestion
    internal sealed class ApplicationAdapter : IApplication

```

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -0,0 +1,98 @@
...
+        internal void OnLogMessageReceived(string condition, string stackTrace, LogType type)
+        {
+            var debounced = type switch
+            {
+                LogType.Error or LogType.Exception or LogType.Assert => ErrorTimeDebounce.Debounced(),
+                LogType.Log => LogTimeDebounce.Debounced(),
+                LogType.Warning => WarningTimeDebounce.Debounced(),
+                _ => true
+            };
+            if (!debounced)
```
bruno-garcia, line 43: Lets check `_hub` for null once here to avoid the `_hub?.` below this point:

```suggestion
            if (!debounced || _hub is null)

```

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (type != LogType.Error && type != LogType.Exception && type != LogType.Assert)
+            {
+                // TODO: MinBreadcrumbLevel
+                // options.MinBreadcrumbLevel
+                _hub?.AddBreadcrumb(condition, level: ToBreadcrumbLevel(type));
+                return;
+            }
+
+            var sentryEvent = new SentryEvent(new UnityLogException(condition, stackTrace));
+            sentryEvent.SetTag("log.type", ToEventTagType(type));
```
bruno-garcia, line 58: Is this supposed to be `event.level` instead?

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -0,0 +1,98 @@
...
+            _hub?.AddBreadcrumb(condition, level: ToBreadcrumbLevel(type));
+        }
+
+        private void OnQuitting()
+        {
+            // Note: iOS applications are usually suspended and do not quit. You should tick "Exit on Suspend" in Player settings for iOS builds to cause the game to quit and not suspend, otherwise you may not see this call.
+            //   If "Exit on Suspend" is not ticked then you will see calls to OnApplicationPause instead.
+            // Note: On Windows Store Apps and Windows Phone 8.1 there is no application quit event. Consider using OnApplicationFocus event when focusStatus equals false.
+            // Note: On WebGL it is not possible to implement OnApplicationQuit due to nature of the browser tabs closing.
+            Application.logMessageReceived -= OnLogMessageReceived;
```
bruno-garcia, line 72:
```suggestion
            _application.logMessageReceived -= OnLogMessageReceived;

```

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -0,0 +1,98 @@
...
+            SentrySdk.Close();
```
bruno-garcia, line 73:
```suggestion
            _hub?.Dispose();
```

One downside of us closing the SDK here (which happened in aspnetcore) is that the order of stuff unregistering is unclear so we can turn off the SDK and more code running ends up on an error that we don't capture.

Alternatively we could simply Flush here (though we only have async flush so would require `GetAwaiter().GetResult()` here)

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -1,197 +1,45 @@
...
-
-            if (!options.CaptureInEditor && Application.isEditor)
+            if (!options.DisableProgrammaticInitialization)
             {
-                options.Logger?.Log(SentryLevel.Info, "Disabled while in the Editor.");
+                /*
+                 * Call Unity's `Debug.Log()` directly via Sentry `UnityLogger`
+                 * We want to display the message in spite of how SentryOptions.json is configured
+                 */
+                new UnityLogger(SentryLevel.Info).Log(SentryLevel.Info, "Programmatic access enabled. Configure Sentry manually.");
```
bruno-garcia, line 19: Why not call `options.DiagnosticLogger` here?

And if really not, why not just `Debug.Log` then?

src/test/Sentry.Unity.Tests/Stubs/TestHub.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using UnityEngine;
+
+namespace Sentry.Unity.Tests.Stubs
+{
+    internal sealed class TestHub : IHub
+    {
+        private readonly List<SentryEvent> _capturedEvents = new ();
```
bruno-garcia, line 10:
```suggestion
        private readonly List<SentryEvent> _capturedEvents = new();

```

src/test/Sentry.Unity.Tests/UnityApplicationLoggingIntegrationTests.cs
```diff
@@ -0,0 +1,47 @@
...
+using Sentry.Unity.Tests.Stubs;
+using UnityEngine;
+
+namespace Sentry.Unity.Tests
+{
+    public sealed class UnityApplicationLoggingIntegrationTests
+    {
+        private class Fixture
+        {
+            public TestHub Hub => new ();
```
bruno-garcia, line 12:
```suggestion
            public TestHub Hub => new();
```

src/test/Sentry.Unity.Tests/UnityApplicationLoggingIntegrationTests.cs
```diff
@@ -0,0 +1,47 @@
...
+            public IAppDomain AppDomain => new TestAppDomain();
+
+            public UnityApplicationLoggingIntegration GetSut() => new (AppDomain);
```
bruno-garcia, line 15:
```suggestion
            public UnityApplicationLoggingIntegration GetSut() => new(AppDomain);

```

src/test/Sentry.Unity.Tests/UnityEventExceptionProcessorTests.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System.Linq;
+using NUnit.Framework;
+
+namespace Sentry.Unity.Tests
+{
+    public class UnityEventExceptionProcessorTests
+    {
+        [Test]
+        public void Process_IL2CPPStackTraceFilenameWithZeroes_ShouldReturnEmptyString()
```
bruno-garcia, line 9: Turns out this should be taken care by the .NET SDK:

https://github.com/getsentry/sentry-dotnet/blob/a9304a0a4b4702d0e62e2703d55c66483d27a0e5/test/Sentry.Tests/Internals/MonoSentryStackTraceFactoryTests.cs#L24-L49

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
bruno-garcia: https://github.com/getsentry/sentry-unity/issues/141

src/Sentry.Unity/Extensions/SentryOptionsExtensions.cs
semuserable: Still not sure what you want to do with this. Your first comment suggests to leave it "as it", the second one is "to remove".

Overall, what value do you expect for `sentryOptions.Environment` field in **the current** PR?

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
   - https://github.com/getsentry/sentry-dotnet/blob/main/CHANGELOG.md#334
   - Bug fixes for performance monitoring
   - Ability to keep failed envelopes for troubleshooting when they are too large
+- Unity Sentry SDK programmatic setup (#130)
```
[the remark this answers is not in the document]
semuserable: ?

Looks fine as is.
image

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -105,10 +107,16 @@ public void OnLostFocus()
         // ReSharper disable once UnusedMember.Local
         private void OnGUI()
         {
+            GUILayout.Label(new GUIContent(GUIContent.none), EditorStyles.boldLabel);
+            Options.DisableProgrammaticInitialization = EditorGUILayout.BeginToggleGroup(
+                new GUIContent("Disable Programmatic Initialization", "Disable manual Sentry setup. Rely on SentryOptions config."),
```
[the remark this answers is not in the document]
semuserable: Yeah, I tried to make it like you described, that's why it feels **backwards** for now. I encountered some problems with the way Unity provides API for this enable/disable functionality. Probably need to take another shot at this.

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -138,6 +142,15 @@ private void OnGUI()
                                                        "If not set, auto detects such as 'development', 'production' or 'editor'."),
                 Options.Environment);
 
+            GUILayout.Label(new GUIContent(GUIContent.none), EditorStyles.boldLabel);
+            Options.DisableAutoCompression = EditorGUILayout.BeginToggleGroup(
+                new GUIContent("Disable Auto Compress Payload", "Disable auto Sentry setup. Rely on SentryOptions config."),
```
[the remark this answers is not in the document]
semuserable: My idea was to use Sentry `CompressionLevel` as is for this window. 

I can create another enum with the same structure as `CompressionLevel`, call it `UnityCompressionLevel` and do the proper mapping to `CompressionLevel`.

Do you prefer the second approach (with `UnityCompressionLevel` enum)?

src/Sentry.Unity/Extensions/JsonExtensions.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Text.Json;
+
+namespace Sentry.Unity.Extensions
+{
+    internal static class JsonExtensions
```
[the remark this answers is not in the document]
semuserable: Currently, not exactly. `GetEnumOrNull` which I created for Unity is not in Sentry SDK. `GetPropertyOrNull` can be reused from Sentry SDK.

Regarding `InternalsVisibleTo`. We must sign `Sentry.Unity` in order to do so. Create an issue for this?

src/Sentry.Unity/Extensions/SentryOptionsExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        public static void ConfigureRequestBodyCompressionLevel(this UnitySentryOptions unitySentryOptions)
+        {
+            // The target platform is known when building the player, so 'auto' should resolve there.
+            // Since some platforms don't support GZipping fallback no no compression.
+            if (unitySentryOptions.DisableAutoCompression)
+            {
+                return;
+            }
+
+            unitySentryOptions.RequestBodyCompressionLevel = CompressionLevel.NoCompression;
```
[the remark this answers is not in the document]
semuserable: No, because it will already contain the selected `CompressionLevel` value from window.

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
semuserable: I don't know what you mean. Can you elaborate?

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
semuserable: I can't call `Dispose` on `IHub`.

src/test/Sentry.Unity.Tests/UnityEventExceptionProcessorTests.cs
semuserable: Hm.. Interesting thing to consider. As a Unity dev I expect Unity related tests to be in Unity and not in Sentry SDK. Should we move it here instead?

src/Sentry.Unity/SentryInitialization.cs
semuserable: There are a couple of reasons behind this approach. My reasoning is the following

# First question

In order to use `SentryOptions.DiagnosticLogger` it should be specifically configured in `SentryOptions`. For example, it should have `SentryOptions.Debug` set to `true` **and** it should have some object assigned to `SentryOptions.DiagnosticLogger` already (ie. `UnityLogger`) **and** some other Unity specific conditions like `!DebugOnlyInEditor || Application.isEditor`. 
 
If user disables Sentry Window we should display the message **in spite of** what was set in `SentryOptions`. I think we should not expect a user to properly configure `SentryWindow` to enable `UnityLogger`, then disable the whole thing so he can see a message.

In order to do it in `SentryOptions` context and not rely on `SentryWindow` settings:
```csharp
options.Debug = true;
options.DiagnosticLogger = new UnityLogger(SentryLevel.Info);
options.DiagnosticLogger.Log(SentryLevel.Info, "Programmatic access enabled. Configure Sentry manually.");
```
but this is not needed as we can just call `UnityLogger` directly.

# Second question

`UnityLogger` has specific format that it outputs into Unity. Yes, we can call it directly via Unity, but we probably want to have Sentry specific formatting applied so the user can differentiate that this message came from Sentry specifically.

src/Sentry.Unity/Extensions/SentryOptionsExtensions.cs
```diff
@@ -6,7 +6,7 @@ namespace Sentry.Unity.Extensions
     internal static class SentryOptionsExtensions
     {
         public static void ConfigureRelease(this SentryOptions sentryOptions)
-            // Uses the game `version` as Release
+            // Uses the game `version` as Release unless the user defined one via the Options
             => sentryOptions.Release = sentryOptions.Release is { } release
                 ? release
                 : Application.version;
```
bruno-garcia: Can't this be:

```suggestion
            => sentryOptions.Release ??= Application.version;
```

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
bruno-garcia: I mean, instead of adding a tag, we can just use `sentryEvent.Level = `

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
bruno-garcia:
```suggestion
            (_hub as IDisposable)?.Dispose()
```

bruno-garcia (commented):
Great improvements. Thanks

src/test/Sentry.Unity.Tests/UnityEventExceptionProcessorTests.cs
bruno-garcia: Now that we have a nice home for it, we could move it here (IL2CPP stuff). Mono things need to be there since Xamarin uses Mono

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
semuserable: What do you expect for `LogType.Assert`?

image

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
semuserable: So, `Sentry.Internal.Hub` inherits from `IDisposable` but not `IHub` itself. Is this some kind of internal contract? Why `IHub` itself is not `IDisposable`?

For example, if I decide to create `AnotherHub` (not test one) which inherits from `IHub`, do I need to also inherit from `IDisposable`? Is this implicit contract each hub should abide to?

src/Sentry.Unity/SentryInitialization.cs
semuserable: @bruno-garcia is the current explanation fine for you? Can I resolve it?

src/test/Sentry.Unity.Tests/UnityEventExceptionProcessorTests.cs
semuserable: Nice! Issue created.

src/Sentry.Unity/Extensions/SentryOptionsExtensions.cs
bruno-garcia: With this, if it wasn't clear:
> We'll need this to get through via a script that compiles with Unity to use their compiler flags.

I meant that `DEVELOPMENT_BUILD` doesn't work here. Unless I'm missing something.
Is this defined by Unity?

src/Sentry.Unity.Editor/SentryWindow.cs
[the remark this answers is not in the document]
bruno-garcia: I think it's simpler to have `auto` in the UI and not bother creating a new enum

src/Sentry.Unity/Extensions/JsonExtensions.cs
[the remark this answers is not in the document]
bruno-garcia: Yeah true, that can be a new PR anyway.

src/Sentry.Unity/SentryInitialization.cs
bruno-garcia: Yeah that makes.

bitsandfoxes:
> What do you think about this updated window?
> image
> 
> Main points
> 
> * Disable Programmatic Initialization
> * Compression level UI based on `SentryOptions` - `System.IO.Compression.CompressionLevel`
> 
> cc @bruno-garcia @bitsandfoxes

I'm not a huge fan of the "Disable Programmatic Initialization" checkbox tbh.
Written code should always overwrite window-settings since it coveys much more intent?
The code initialization could also write to the same config file and overwrite it so the window 'syncs' up with the code.
We could throw warnings if we spot a difference between config file and code in the editor?

CHANGELOG.md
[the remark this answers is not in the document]
bruno-garcia: my bad

bruno-garcia:
We should discuss this on a call and wrap it up together

semuserable:
Update with the latest.

cc @bruno-garcia let's review it together on a call and close it.

PR #139, opened 2021-05-02
build game before dotnet project

bruno-garcia:
Build the Unity sample (installs packages required to build .NET) before building the .NET project

Prob not the final format since we dont' want this to run each time. 
Perhaps best to run after a Clean, or if we detect the unity packages haven't been restored yet.

Also useful for CI (won't need the manual step

- Somehow fails the first build. Unity builds seems to return before the test DLLs are available. The .NET build that kicks off right after the Unity's fail. But If I trigger it again right after, it works.
Seems like it worked on Windows in CI, but that's what I noticed on a mac (CI cache maybe?)

#skip-changelog

### DIFF
--- a/package-dev/Tests/Runtime/TestSentryOptions.json
+++ b/package-dev/Tests/Runtime/TestSentryOptions.json
@@ -1,0 +1,1 @@
+{"enabled":true,"captureInEditor":true,"dsn":"http://test.com","debug":true,"debugOnlyInEditor":true,"diagnosticsLevel":3,"requestBodyCompressionLevel":1,"attachStacktrace":true,"sampleRate":1,"release":"release","environment":"test"}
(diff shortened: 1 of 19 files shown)


package-dev/Tests/Runtime/TestSentryOptions.json
```diff
@@ -0,0 +1 @@
+{"enabled":true,"captureInEditor":true,"dsn":"http://test.com","debug":true,"debugOnlyInEditor":true,"diagnosticsLevel":3,"requestBodyCompressionLevel":1,"attachStacktrace":true,"sampleRate":1,"release":"release","environment":"test"}
```
bruno-garcia, line 1: should be this be in gitignore?

bruno-garcia:
@semuserable @bitsandfoxes thoughts?

package-dev/Tests/Runtime/TestSentryOptions.json
bitsandfoxes: If we ship the package pre-configured wouldn't that be a good thing?
Install Package -> Hit Play -> Events get already logged on our Unity-Project?

bruno-garcia: This one seems to be only used on tests though (based on the name). I believe it's created during Play tests.

semuserable: This is used in tests only, yes. Seems ok to be `.gitignore`d.

semuserable (approved):

PR #142, opened 2021-05-05
POC: FPS metric

bitsandfoxes:
Test if we can validate fps:

Metric used here:

- Good frames ( < 20ms )
- Bad frames ( > 20ms but not yet frozen)
- Frozen frames

Frozen frame issue: Unity has a limit on how big delta time can get to preserve continuity for the player experience. We could work around that by calculating the delta time ourselves -> MeasureDeltaTime()
Things to consider for the metrics:

- How often did the game get frozen
- How long was is frozen for

Screenshot 2021-05-05 at 10 37 48

### DIFF
--- a/src/Sentry.Unity/SentryBehavior.cs
+++ b/src/Sentry.Unity/SentryBehavior.cs
@@ -1,18 +1,68 @@
-﻿using Sentry;
+﻿using System;
+using System.Collections;
+using System.Threading;
 using UnityEngine;
-using Sentry.Unity;
-using System;
-using Sentry.Protocol;
 
-public class SentryBehavior : MonoBehaviour
+namespace Sentry.Unity
 {
-    public void Disable()
+    public class SentryBehavior : MonoBehaviour
     {
-  
-    }
+        private int goodFrames;
+        private int slowFrames;
+        private int frozFrames;
 
-    // TODO: Flush events. See note on OnApplicationQuit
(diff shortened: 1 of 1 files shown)


src/Sentry.Unity/SentryBehavior.cs
```diff
@@ -1,18 +1,68 @@
...
+            {
+                frozFrames++;
+            }
+            else if (Time.deltaTime > slowThreshold)
+            {
+                slowFrames++;
+            }
+            else
+            {
+                goodFrames++;
```
bruno-garcia, line 56: Some thoughts:
Should we use long here? And if so, do we need `Interlocked.Increment`? And if we do :) does that work everywhere on Unity, like WebGL?

bruno-garcia:
Once we have some transactions being generated (see: https://github.com/getsentry/sentry-unity/issues/235) we can get back to this PR and add some measurements

bruno-garcia:
@bitsandfoxes this one soon will make sense? with the new metrics stuff

bruno-garcia:
@bitsandfoxes with 4.0.0 you could use this now

bitsandfoxes:
> @bitsandfoxes with 4.0.0 you could use this now

I think the name is misleading. Initially, we wanted to have those as mobile vitals.

bitsandfoxes:
Closing this as `stale`.

PR #149, opened 2021-05-10
Android support

bruno-garcia:
Brings back the `crash` stuff I did on: https://github.com/bruno-garcia/unity-of-bugs

Starting to look on the best approach for the Android support. I believe the submodule is the way to go. The PoC used a gradle dependency on the template (done directly on the app. Not sure it's possible through a library) but I anyway believe it's best to build from source, using git submodule as the ideal dev loop/quick patch.

We can have CI build in separate steps that run in parallel and have a final assemble part with integration tests for the bits that put together the Sentry Unity/.NET stuff with native plugins. What y'all think?

To figure out:

 - [x] How do bring dependencies in? Currently only Gson is missing. Ideally we'd automate copying all dependencies of the `gradle` build into the Unity page. Similar to how we did with .NET. @marandaneto might have ideas here
 - [x] Build project as part of msbuild build. Result of build should go into package-dev. Publish script to copy artifacts to final `package`. `Clean` task from msbuild to remove artifacts and call `gradlew clean`
 - [ ] Add Gradle plugin to upload symbols (or use MSBuild wrapping sentry-cli, [part of this PR](https://github.com/getsentry/sentry-cli/pull/961))
 - [ ] Add [back manifest manipulation](https://github.com/getsentry/sentry-unity/blob/fd4a26e239d18865f1c947e3551178a1cde00120/src/Sentry.Unity.Editor/AndroidManifestConfiguration.cs) to pass DSN (or opt-out of auto init to use programatic approach)
 - [ ] If we're letting it init via content provider, how do we sync scope/session/etc?
 - [ ] Define story to opt-out from native dependencies (i.e: SentryWindow option to remove the .aar/jar ?)


semuserable:
A little demo would be helpful :)

bruno-garcia:
Will do once I get time to mature this a bit

PR #150, opened 2021-05-10
POC: Installation config window

bitsandfoxes:
We hook into the AssetPostprocessor to check if the Sentry package has been modified.

Screenshot 2021-05-10 at 13 35 54

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -9,8 +9,8 @@ artifactProvider:
   name: github
 targets:
   - name: upm
-    releaseOwner: "getsentry"
-    releaseRepo: "unity"
+    releaseRepoOwner: "getsentry"
+    releaseRepoName: "unity"
   - name: github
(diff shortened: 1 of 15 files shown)


.craft.yml
```diff
@@ -9,8 +9,8 @@ artifactProvider:
   name: github
 targets:
   - name: upm
-    releaseOwner: "getsentry"
-    releaseRepo: "unity"
+    releaseRepoOwner: "getsentry"
```
bruno-garcia, line 12: Should this land on another PR? Or just push to main if we need this to release once craft is out.

bruno-garcia:
@bitsandfoxes if you don't plan to use this branch anymore, best to nuke it:

image

We configure the GitHub repos to remove branches automatically when we merge then, but if we close the PR, they linger

PR #158, opened 2021-05-11
commit options

bruno-garcia:
Keep the options with the test dsn checked-in


bitsandfoxes (approved):
Makes sense to me.

semuserable:
Why keeping `link.xml`? Do we want to manage it in both places from now on? `Package/Runtime/link.xml` and `Assets/Resources/Sentry/link.xml`?

Do we need to remove the logic of copying `link.xml` based on this?

PR #167, opened 2021-05-13
test UnityLogExceptionTests

bruno-garcia:
I noticed some stack traces rendered odd:

image

To be able to test the parsing logic, we needed some tests to validate the current behavior.
This PR adds some of those tests. Currently only for Mono stack traces. 

Pending IL2CPP examples.

The goal is to be able to change the logic with confidence.

Earlier today I patched the .NET SDK so it can handle `-1` line numbers:
https://github.com/getsentry/sentry-dotnet/pull/994

Turns out this is added by this current parsing logic so we can also change here to avoid adding a line altogether if it can't parse the proper line number instead of adding `-1`. 

Changes will come in follow up PRs.

#skip-changelog


bruno-garcia:
Surely it breaks when parsing on Windows:

```
Expected string length 199 but was 364. Strings differ at index 199.
  Expected: "...s+EventFunction`1[T1] functor)"
  But was:  "...s+EventFunction`1[T1] functor) (at /Applications/Unity/Hub...
```

The verbatim string on Windows becomes `\r\n` so the parsing code must be choking on that.

bruno-garcia:
Good test case:

```
at System.Linq.Enumerable.Single[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x0002d] in <351e49e2a5bf4fd6beabb458ce2255f3>:0 
  at Sentry.Unity.Tests.SceneManagerIntegrationTests.SceneLoaded_EnabledHub_CrumbAdded () [0x00065] in /Users/bruno/git/sentry-unity/test/Sentry.Unity.Tests/SceneManagerIntegrationTests.cs:37 
  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00032] in <eae584ce26bc40229c1b1aa476bfa589>:0
```

semuserable:
I called `ExceptionToString` from editor on the latest main (Windows) and from 0.0.13 `il2cpp` build (Windows):

Editor:
image

il2cpp
image

Seems fine for me. Am I missing something?

bruno-garcia:
Maybe the version. Maybe macOS x Windows difference, not sure.

bruno-garcia:
What I meant that breaks on Windows are my tests (I used verbatin string so the line break is platform sensitive and my test isn't taking that into account. Needs fixing.

semuserable:
`Environment.NewLine` probably?

semuserable:
Just pulled and ran tests

image

What am I doing wrong? :)

bruno-garcia:
Thanks for checking!
The line break I was referring to isn't the whole file but the string I used verbatin in (the hard coded stack traces.
But if tests passed on your box then it must something else.

semuserable:
* After investigating for a bit, I noticed a strange exception

image

Not sure it's related to the filed tests, but still.

* test errors

From CI
image

Test Code
image

It's just a space.

Still, running fine on my machine, but fails on CI.

bruno-garcia:
> From CI
> image
> 
> Test Code
> image
> 
> It's just a space.
> 
> Still, running fine on my machine, but fails on CI.

Also worked on my Windows VM and continues to fail in CI. In CI it seems to not parse the line (the space is followed by the file path in there)

CI should be running exactly what we are so not sure why it fails, still

bruno-garcia:
Very strange. The code looks for `IndexOf(')')` and in the middle of the stacktrace, for one of the frames, it doesn't find the first occurance, but only the last one;


```
UnityEngine.EventSystems.ExecuteEvents.Execute[T] (UnityEngine.GameObject target, UnityEngine.EventSystems.BaseEventData eventData, UnityEngine.EventSystems.ExecuteEvents+EventFunction`1[T1] functor) (at /Applications/Unity/Hub/Editor/2019.4.21f1/Unity.app/Contents/Resources/PackageManager/BuiltInPackages/com.unity.ugui/Runtime/EventSystem/ExecuteEvents.cs:261)
```

bruno-garcia:
Guess what fixed it: 

image

PR #170, opened 2021-05-13
feat: scene manager integration

bruno-garcia:
Currently breadcrumbs.

I haven't tested this on a game with scenes so I'm not sure if it makes sense to add crumbs for the 3 events.
Does the Active Scene one always trigger one either Load or Unload kicks off?

We can also use this, hopefully to get timing info (spans) and perhaps track metrics too.


bruno-garcia:
Merged by accident.

It does fire for both btw:

image

PR #171, opened 2021-05-13
dont install dotnet in ci

bruno-garcia:
.NET 5 is already available on GHA

#skip-changelog


bruno-garcia:
Down to 12 min:

image

PR #174, opened 2021-05-13
fix: git submodule restore

bruno-garcia:
Removing from CI parts that dotnet build is supposed to figure out now

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
       # relative paths for certain things.
       # Also, we need to provide the path to Unity because
       # the test project references some of its assemblies.
-      - name: Build Sentry package
+      - name: Build Sentry.Unity Solution
         run: |
           dotnet nuget locals all --clear
           dotnet build -c Release
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,24 +36,24 @@ jobs:
       # Caching still takes some time, but brings it down to about 5 minutes.
       # Note: if you change the installation step, remember to update the value
       # at the end of the key.
-      - name: Restore cached Unity installation
-        id: cache-unity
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -31,8 +31,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2.3.3
-        with:
-          submodules: recursive
```
bruno-garcia, line 35: Removed because restoring submodules is part of `dotnet build`;

https://github.com/getsentry/sentry-unity/pull/174/files#diff-d77672cdc87b873cf5dd4359e72764340e39d697cb8100c88e1363ae5ff2c67a

.github/workflows/ci.yml
```diff
@@ -58,13 +56,13 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('samples/unity-of-bugs/Packages/packages-lock.json') }}
 
       - name: Install Unity
+        if: ${{ !steps.cache-unity.outputs.cache-hit }}
```
bruno-garcia, line 59: Easier to read when the condition precedes the step instructions

.github/workflows/ci.yml
```diff
@@ -95,18 +93,14 @@ jobs:
...
-
       # The working directory must be set because the project uses
       # relative paths for certain things.
       # Also, we need to provide the path to Unity because
       # the test project references some of its assemblies.
-      - name: Build Sentry package
+      - name: Build Sentry.Unity Solution
         run: |
           dotnet nuget locals all --clear
-          dotnet build -c Release "-p:UnityPath=C:/Program Files/Unity/Editor/Data/Managed"
```
bruno-garcia, line 109: msbuild needs to find Unity (we want that logic to be tested here in CI) so not providing the path to it

PR #177, opened 2021-05-14
Fix: Programmatic config - logger attachment

bitsandfoxes:
When the sdk gets configured programmatically we attach the logger just before we initialize the SDK itself.
#skip-changelog


bruno-garcia (approved):
Nice!

PR #178, opened 2021-05-14
ref: package cleanup

bitsandfoxes:
removed unused packages from the manifest
#skip-changelog


bruno-garcia:
lets call `ref` for refactor.

PR #181, opened 2021-05-14
try to silence the warning

semuserable:
Try to silence "Assembly for Assembly Definition File 'Packages/io.sentry.unity.dev/Runtime/io.sentry.unity.dev.runtime.asmdef' will not be compiled, because it has no scripts associated with it."

#skip-changelog


bruno-garcia (approved):
That was easy

PR #184, opened 2021-05-17
Flag simulator based on Application.isEditor

semuserable:


bitsandfoxes:
Since IApplication now has a concept of IsSimulator:
https://github.com/getsentry/sentry-unity/blob/9768a0d5b47c45466c8100af0a9bae9aca29c909/src/Sentry.Unity/SentryUnityOptions.cs#L93
https://github.com/getsentry/sentry-unity/blob/9768a0d5b47c45466c8100af0a9bae9aca29c909/src/Sentry.Unity/SentryUnityOptions.cs#L111
I think those would be fixed too?

semuserable:
Tbh, not sure about `SentryUnityOptions`, because we have `IApplication` dependency only on integrations and not on `SentryUnityOptions` construction itself.

src/Sentry.Unity/Integrations/IApplication.cs
```diff
@@ -27,6 +28,8 @@ private ApplicationAdapter()
 
         public string ActiveSceneName => SceneManager.GetActiveScene().name;
 
+        public bool IsSimulator => Application.isEditor;
```
bruno-garcia, line 31: Lets map the IApplication 1 to 1 with the Application class from Unity

```suggestion
        public bool IsEditor => Application.isEditor;

```

src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -20,6 +21,13 @@ internal static class UnitySdkInfo
 
     internal class UnityEventProcessor : ISentryEventProcessor
     {
+        private readonly IApplication _appDomain;
+
+        public UnityEventProcessor(IApplication? appDomain = null)
```
bruno-garcia, line 26: there's no AppDomain here:

```suggestion
        public UnityEventProcessor(IApplication? application = null)

```

bruno-garcia:
> Tbh, not sure about `SentryUnityOptions`, because we have `IApplication` dependency only on integrations and not on `SentryUnityOptions` construction itself.

Ideally we'd decouple those but on a follow up PR woudd be fine. That's not high priority given we're aiming at shipping 0.1.0 asap

PR #185, opened 2021-05-18
Transition scene to test scene loading events (breadcrumbs)

semuserable:
* New `TransitionScene` with two buttons: load `BugFarmScene` and `ThrowNull`. 
* New button (Load TransitionScene) in `BugFarmScene`

Now can test load scene events.


samples/unity-of-bugs/Assets/Scripts/SceneLoaderBehaviour.cs
```diff
@@ -0,0 +1,12 @@
+﻿using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace Assets.Scripts
```
bitsandfoxes, line 4: We have two scripts in samples: One lives inside a namespace, the other doesn't.

samples/unity-of-bugs/Assets/Scripts/SceneLoaderBehaviour.cs
```diff
@@ -0,0 +1,12 @@
+﻿using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace Assets.Scripts
+{
+    public sealed class SceneLoaderBehaviour : MonoBehaviour
```
bitsandfoxes, line 6: Same with sealed. 
I would like for us to keep the naming consistent. We have BugFarm so this could be BugFarmSceneLoader or even just SceneLoader.

bitsandfoxes (commented):
Now that we have multiple scenes we could rename them to BugFarmLevel_1 / Level_2 respectively. Other than us noone knows what a "TransitionScene" is and what it's for. With the numeric postfix there is an implied hierarchy that helps with self documenting.

bitsandfoxes:
This is nitpicking but please reset the transform for TransitionScene/Scripts GameObject.

samples/unity-of-bugs/Assets/Scripts/SceneLoaderBehaviour.cs
semuserable: +

samples/unity-of-bugs/Assets/Scripts/SceneLoaderBehaviour.cs
semuserable: This class contains actions like `LoadTransitionScene` or `LoadBugFarmScene`. Why to rename it to `BugFarmSceneLoader`? It's not `BugFarm` related. `TransitionScene` calls `LoadBugFarmScene` for eaxmple.

What about `sealed`?

semuserable:
> Now that we have multiple scenes we could rename them to BugFarmLevel_1 / Level_2 respectively. Other than us noone knows what a "TransitionScene" is and what it's for. With the numeric postfix there is an implied hierarchy that helps with self documenting.

I have no strict preference to naming the scenes, but I prefer to give descriptive names. "Us" is developers who develop the package, so "we" understand what "TransitionScene" means.

bitsandfoxes:
> > Now that we have multiple scenes we could rename them to BugFarmLevel_1 / Level_2 respectively. Other than us noone knows what a "TransitionScene" is and what it's for. With the numeric postfix there is an implied hierarchy that helps with self documenting.
> 
> I have no strict preference to naming the scenes, but I prefer to give descriptive names. "Us" is developers who develop the package, so "we" understand what "TransitionScene" means. We can also document it.

But it's not just "us developers" who get to see and use it, right? We ship it as sample with our package to users. And they should not need to read the documentation to know which scene is the first they should want to open.

samples/unity-of-bugs/Assets/Scripts/SceneLoaderBehaviour.cs
semuserable: I'll rename `BugFarm` to `BugFarmBehaviour` then.

I don't agree with your point on `sealed` classes. `sealed` classes usually can be further optimized by a compiler, so I'm not buying into "visual noise".

semuserable:
> > > Now that we have multiple scenes we could rename them to BugFarmLevel_1 / Level_2 respectively. Other than us noone knows what a "TransitionScene" is and what it's for. With the numeric postfix there is an implied hierarchy that helps with self documenting.
> > 
> > 
> > I have no strict preference to naming the scenes, but I prefer to give descriptive names. "Us" is developers who develop the package, so "we" understand what "TransitionScene" means. We can also document it.
> 
> But it's not just "us developers" who get to see and use it, right? We ship it as sample with our package to users. And they should not need to read the documentation to know which scene is the first they should want to open.

As I was saying, I have no strong preference about scene namings. If in Unity community it's better to name scenes like Level_n, then sure, let's do that.

samples/unity-of-bugs/Assets/Scripts/BugFarm.cs
```diff
@@ -6,121 +6,124 @@
 using UnityEngine;
 using UnityEngine.Assertions;
 
-public class BugFarm : MonoBehaviour
+namespace Assets.Scripts
```
bruno-garcia, line 9: Do we really need a namespace?

Because this is a sample for customers to learn about Sentry, I wanted to keep this as simple as possible and originally didn't add a namespace.

semuserable: I thought you just missed the namespace in the first place :)

If you don't want it, we can remove it. But personally, I'd prefer to leave it as is and not in `global::` scope. 

Remove?

bruno-garcia: Sorry I got back from PTO today. I think it's worth keeping these samples (that go into the user games through the UPM sample feature) simple and remove the namespace altogether

semuserable: ok, will remove namespaces then

semuserable:
updated

bitsandfoxes (approved):
Looks good to me! Thanks!

PR #194, opened 2021-05-21
Feat: Embedded link.xml in assembly

bitsandfoxes:
Instead of having to copy the link.xml into Assets/ we can embed the file into the assembly.

https://sentry.io/organizations/sentry-sdks/issues/2410989926/?project=5439417&query=is%3Aunresolved


test/Sentry.Unity.Tests/EmbeddedResourcesTests.cs
```diff
@@ -0,0 +1,18 @@
...
+using NUnit.Framework;
+
+namespace Sentry.Unity.Tests
+{
+    public sealed class EmbeddedResourcesTests
+    {
+        [Test]
+        public void LinkXml_Embedded()
+        {
+            var resourceNames = Assembly.GetAssembly(typeof(SentryUnity)).GetManifestResourceNames();
```
bitsandfoxes, line 12: I'm not sure if this is the proper way to get to the Assembly so any feedback would be greatly appreciated.

bruno-garcia:
Does Unity support this? `link.xml` is used by Unity at compile time. It doesn't need to ship with the game, but IL2CPP needs this file when it runs to make sure it doesn't drop the types defined in it.

bitsandfoxes:
Yes they do! https://forum.unity.com/threads/the-current-state-of-link-xml-in-packages.995848/ <- last reply point a)
I tested it on an IL2CPP build on the phone and added the link to the captured event to the description.

bruno-garcia:
@bitsandfoxes if you remove link.xml completely, does it fail to capture?
I'm unsure at this point which rules we really need.

test/Sentry.Unity.Tests/EmbeddedResourcesTests.cs
bruno-garcia, line 12: Easier like:
```suggestion
            var resourceNames = typeof(SentryUnity).Assembly.GetManifestResourceNames();
```

test/Sentry.Unity.Tests/EmbeddedResourcesTests.cs
```diff
@@ -0,0 +1,18 @@
...
+
+            Assert.NotNull(resourceNames);
+            Assert.IsTrue(resourceNames.Contains("Sentry.Unity.xml"));
```
bruno-garcia, line 15: is there a:
```suggestion
            Assert.Contains(resourceNames, "Sentry.Unity.xml");
```
instead?

Or the order is inverted.
This might be xUnit only

bruno-garcia (approved):
This is a great find btw! Much nicer way to manage the link.xml

semuserable (approved):

bitsandfoxes:
> @bitsandfoxes if you remove link.xml completely, does it fail to capture?
> I'm unsure at this point which rules we really need.

It does. If the file is missing but the itemgroup still includes it the test fails regardless.

test/Sentry.Unity.Tests/EmbeddedResourcesTests.cs
bitsandfoxes: There was! Totally missed it. Thanks!

PR #202, opened 2021-05-27
Fix: Add product name to release as default

bitsandfoxes:
Release should be` productName@version` if it's not overwritten by the user.

### DIFF
--- a/src/Sentry.Unity/SentryUnity.cs
+++ b/src/Sentry.Unity/SentryUnity.cs
@@ -32,7 +32,7 @@ namespace Sentry.Unity
             // Uses the game `version` as Release unless the user defined one via the Options
             if (unitySentryOptions.Release == null)
             {
-                unitySentryOptions.Release = Application.version;
+                unitySentryOptions.Release = Application.productName + "@" + Application.version;
                 unitySentryOptions.DiagnosticLogger?.Log(SentryLevel.Debug,
                     "Setting Sentry Release to Unity App.Version: {0}",
                     null, unitySentryOptions.Release);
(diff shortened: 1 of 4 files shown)


src/Sentry.Unity/SentryUnity.cs
```diff
@@ -32,7 +32,7 @@ public static void Init(SentryUnityOptions unitySentryOptions)
             // Uses the game `version` as Release unless the user defined one via the Options
             if (unitySentryOptions.Release == null)
             {
-                unitySentryOptions.Release = Application.version;
+                unitySentryOptions.Release = Application.productName + "@" + Application.version;
```
bruno-garcia, line 35: If tests don't break, it'll be a bad sign :D

bruno-garcia, line 35: Lets go with interpolation:

```suggestion
                unitySentryOptions.Release = $"{Application.productName}@{Application.version}";

```

bruno-garcia (approved):
nice!

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -32,7 +32,8 @@ public static void Init(SentryUnityOptions unitySentryOptions)
             // Uses the game `version` as Release unless the user defined one via the Options
             if (unitySentryOptions.Release == null)
             {
-                unitySentryOptions.Release = Application.productName + "@" + Application.version;
+                unitySentryOptions.Release = $"{Application.productName}@{Application.version}";
```
bruno-garcia, line 35: Just realized something: Can `productName` be null or an empty string? We might want to check for that before appending `@`

bitsandfoxes: It can not be null or empty. But it could be just a whitespace. To what should we default in that case? Just the version number without the @?

PR #203, opened 2021-05-27
Breadcrumb categories added

semuserable:


bruno-garcia (commented):
LGTM but some tests to validate this and avoid regression would be good

semuserable:
I'll check how it can be done.

semuserable:
New PR will be created, this one is disbanded. Don't want to spend time on conflicts based on line endings.

PR #205, opened 2021-05-28
Fix: Fallback for missing/empty product name

bitsandfoxes:
If for some reason the product name is missing / just whitespace we just use the version as release.

### DIFF
--- a/src/Sentry.Unity/SentryUnity.cs
+++ b/src/Sentry.Unity/SentryUnity.cs
@@ -32,7 +32,7 @@ namespace Sentry.Unity
             // Uses the game `version` as Release unless the user defined one via the Options
             if (unitySentryOptions.Release == null)
             {
-                unitySentryOptions.Release = String.IsNullOrWhiteSpace(Application.productName)
+                unitySentryOptions.Release = string.IsNullOrWhiteSpace(Application.productName)
                     ? $"{Application.version}"
                     : $"{Application.productName}@{Application.version}";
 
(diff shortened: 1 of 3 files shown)


src/Sentry.Unity/SentryUnity.cs
```diff
@@ -32,7 +32,9 @@ public static void Init(SentryUnityOptions unitySentryOptions)
             // Uses the game `version` as Release unless the user defined one via the Options
             if (unitySentryOptions.Release == null)
             {
-                unitySentryOptions.Release = $"{Application.productName}@{Application.version}";
+                unitySentryOptions.Release = String.IsNullOrWhiteSpace(Application.productName)
```
semuserable, line 35: `String` can bind to another class, while `string` is an alias to `System.String`. Use `string` instead of `String`.

semuserable (approved):

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -32,7 +32,9 @@ public static void Init(SentryUnityOptions unitySentryOptions)
             // Uses the game `version` as Release unless the user defined one via the Options
             if (unitySentryOptions.Release == null)
             {
-                unitySentryOptions.Release = $"{Application.productName}@{Application.version}";
+                unitySentryOptions.Release = string.IsNullOrWhiteSpace(Application.productName)
+                    ? $"{Application.version}"
+                    : $"{Application.productName}@{Application.version}";
```
bruno-garcia, line 37: `productName` is a native method so it requires P/Invoke to call into:

```
    public static extern string productName { [FreeFunction("GetPlayerSettings().GetProductName"), MethodImpl(MethodImplOptions.InternalCall)] get; }
```

It's better to call it once:

```suggestion
                unitySentryOptions.Release = Application.productName is string productName 
                                         && !string.IsNullOrWhiteSpace(productName)
                ? $"{productName}@{Application.version}"
                : Application.version;
```

PR #207, opened 2021-05-31
test/through msbuild

bruno-garcia:
The goal is to run Unity tests via `dotnet`. This way we can more easily run things via the CLI and troubleshoot CI issues localy.

At first I made the tests run through `dotnet test` but that kicks off two msbuilds (one for each test project) and both trying to run Unity fails. Unity is only able to have a single instance at a time running on a project.
So instead of depending on `VSTest` target, I made it depend on no target, and we can run it with `/t:`. See `test.sh`.

I've also implemented the test result parsing so we can see the summary of the test run on the console.
I noticed the current version:

https://github.com/getsentry/sentry-unity/blob/53bff3ea62a60f8ad29249a71e3209034f60c7f5/.github/workflows/ci.yml#L133-L143

Displays the test counts but the code supposed to show the messages doesn't work. The new version does print out all the error messages.

#skip-changelog


Directory.Build.targets
```diff
@@ -0,0 +1,194 @@
+<Project>
+
+  <PropertyGroup>
+    <UnityVersion>2019.4.21f1</UnityVersion>
+  </PropertyGroup>
+
```
Tyrrrz, line 6: Just some comments to make it easier to parse.

```suggestion

    <!-- Unity paths on Windows -->
```

Directory.Build.targets
```diff
@@ -0,0 +1,194 @@
...
+  <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
+    <UnityRoot Condition="Exists('C:\Program Files\Unity\Hub\Editor\$(UnityVersion)\Editor\Data\Managed\UnityEngine.dll')">C:\Program Files\Unity\Hub\Editor\$(UnityVersion)\Editor</UnityRoot>
+    <!--If not using Unity Hub, tries to pick whatever Unity version is installed on the machine-->
+    <UnityRoot Condition="$(UnityRoot) == '' AND Exists('C:\Program Files\Unity\Editor\Data\Managed\UnityEngine.dll')">C:\Program Files\Unity\Editor</UnityRoot>
+    <UnityManagedPath>$(UnityRoot)\Data\Managed</UnityManagedPath>
+    <UnityExec>$(UnityRoot)\Unity.exe</UnityExec>
+    <StandalonePlayerName>buildWindows64Player</StandalonePlayerName>
+    <ArtifactName>game.exe</ArtifactName>
+  </PropertyGroup>
+
```
Tyrrrz, line 16:
```suggestion

    <!-- Unity paths on macOS -->
```

Directory.Build.targets
```diff
@@ -0,0 +1,194 @@
...
+    <Message Importance="High" Text="Running Unity EditMode tests."></Message>
+
+    <Exec Command="&quot;$(UnityExec)&quot; -batchmode -nographics -logFile - -runTests -testPlatform EditMode -projectPath $(UnitySampleProjectPath) -testResults $(UnityTestEditModeResultFilePath) "
+          IgnoreStandardErrorWarningFormat="true"
+          IgnoreExitCode="true"></Exec>
+
+    <UnityTestResults Path="$(UnityTestEditModeResultFilePath)" />
+  </Target>
+
```
Tyrrrz, line 84:
```suggestion

    <!-- Parse test results -->
```

Directory.Build.targets
```diff
@@ -0,0 +1,194 @@
...
+      [usbmuxd] Stop listen thread
+      [usbmuxd] Error:
+      [usbmuxd] Listen thread exitingit -batchmode -nographics -logFile - -projectPath $(UnitySampleProjectPath) -$(StandalonePlayerName) $(ArtifactName)
+    Related: https://forum.unity.com/threads/6572-debugger-agent-unable-to-listen-on-27.500387/  -->
+    <Exec Command="&quot;$(UnityExec)&quot; -quit -batchmode -nographics -logFile - -projectPath $(UnitySampleProjectPath) -$(StandalonePlayerName) $(ArtifactName) "
+      IgnoreStandardErrorWarningFormat="true"
+      IgnoreExitCode="true"></Exec>
+  </Target>
+
+  <!-- Run PlayMode tests with dotnet test -->
```
Tyrrrz, line 56: Does it actually run with `dotnet test`? Looks like it runs with `dotnet msbuild /t:UnityPlayModeTest`. It's possible to make it run with `dotnet test` though if you make a custom test adapter, which is not difficult but most likely not worth it though.

bruno-garcia, line 56:
```suggestion
  <!-- Run PlayMode tests with dotnet msbuild /t:UnityPlayModeTest -->
```

Directory.Build.targets
```diff
@@ -0,0 +1,194 @@
...
+
+    <Message Importance="High" Text="Running Unity PlayMode tests."></Message>
+    <Exec Command="&quot;$(UnityExec)&quot; -batchmode -nographics -logFile - -runTests -testPlatform PlayMode -projectPath $(UnitySampleProjectPath) -testResults $(UnityTestPlayModeResultFilePath) "
+          IgnoreStandardErrorWarningFormat="true"
+          IgnoreExitCode="true"></Exec>
+
+    <UnityTestResults Path="$(UnityTestPlayModeResultFilePath)" />
+  </Target>
+
+  <!-- Run EditMode tests with dotnet test. -->
```
bruno-garcia, line 70:
```suggestion
  <!-- Run EditMode tests with dotnet msbuild /t:UnityPlayModeTest. -->
```

Directory.Build.targets
bruno-garcia: fixed

PR #208, opened 2021-05-31
Feat: Offline caching

bitsandfoxes:
Setting default path on option creation.


src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -84,6 +84,7 @@ public SentryUnityOptions()
         {
             // IL2CPP doesn't support Process.GetCurrentProcess().StartupTime
             DetectStartupTime = StartupTimeDetectionMode.Fast;
+            CacheDirectoryPath = Application.persistentDataPath;
```
bruno-garcia, line 87: I wonder if we might have multiple instances of the same game open and that would conflict here.
It isnt' a problem on mobile but it could happen on desktop. @Tyrrrz would the .NET SDK handle that today?

Tyrrrz: No, it won't. We decided not to add it for now.

bruno-garcia: Yeah. That makes me wonder if we need to solve this now, or just document this.

> Note: For Desktop games: If more than one instance of your game can run at the same time, you'll need to pass a separate path explicitly.

Or something along those li es

Tyrrrz: It's not a terribly hard problem to solve i think

bruno-garcia: We talked briefly and it was a bit annoying. Since crashes are sent after restart, we need a segment of the path that survives restart. So `pid` won't do for example. Not a hard problem but not trivial as 'add a hash of X in there' kind of trivial

src/Sentry.Unity/SentryDefaultOptionSetter.cs
```diff
@@ -0,0 +1,42 @@
...
+using Sentry.Unity.Integrations;
+using UnityEngine;
+
+namespace Sentry.Unity
+{
+    internal static class SentryDefaultOptionSetter
+    {
+        public static void SetRelease(SentryUnityOptions options, IApplication? appDomain = null)
+        {
+            var application = appDomain ?? ApplicationAdapter.Instance;
```
bruno-garcia, line 11:
```suggestion
        public static void SetRelease(SentryUnityOptions options, IApplication? application = null)
        {
           application ??= ApplicationAdapter.Instance;
```

src/Sentry.Unity/SentryDefaultOptionSetter.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            options.CacheDirectoryPath ??= application.PersistentDataPath;
+
+            Log(options.DiagnosticLogger, "Cache Directory", options.CacheDirectoryPath);
+        }
+
+        private static void Log(IDiagnosticLogger? logger, string sentryValue, object value)
+        {
+            logger?.Log(SentryLevel.Debug, "Setting Sentry " + sentryValue + " to: {0}", null, value);
```
bruno-garcia, line 39: The goal of the template `{x}` is to avoid allocations (string concatenation) if logging is disabled:

```suggestion
            logger?.Log(SentryLevel.Debug, "Setting Sentry {0} to: {1}", null, sentryValue, value);
```

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -29,24 +29,9 @@ public static void Init(SentryUnityOptions unitySentryOptions)
...
-                    "Setting Sentry Release to Unity App.Version: {0}",
-                    null, unitySentryOptions.Release);
-            }
-
-            unitySentryOptions.Environment = unitySentryOptions.Environment is { } environment
-                ? environment
-                : Application.isEditor // TODO: Should we move it out and use via IApplication something?
-                    ? "editor"
-                    : "production";
+            SentryDefaultOptionSetter.SetRelease(unitySentryOptions);
```
bruno-garcia, line 32: What about a single method that sets all defaults?

`SentryDefaultOptionSetter.SetDefaults(options)`

src/Sentry.Unity/Integrations/UnityBeforeSceneLoadIntegration.cs
```diff
@@ -7,8 +7,8 @@ internal sealed class UnityBeforeSceneLoadIntegration : ISdkIntegration
     {
         private readonly IApplication _application;
 
-        public UnityBeforeSceneLoadIntegration(IApplication? appDomain = null)
-            => _application = appDomain ?? ApplicationAdapter.Instance;
+        public UnityBeforeSceneLoadIntegration(IApplication? application = null)
+            => _application = application ?? ApplicationAdapter.Instance;
```
bruno-garcia, line 11: [C# 8 can help here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator)

```suggestion
            => _application ??= ApplicationAdapter.Instance;
```

When using Rider, it should show you a hint to refactor that (with CMD+Enter or whatever shortcut you opt-ed for)

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -0,0 +1,44 @@
...
+        private static void SetCacheDirectoryPath(SentryUnityOptions options, IApplication application)
+        {
+            options.CacheDirectoryPath ??= application.PersistentDataPath;
+            Log(options.DiagnosticLogger, "Cache Directory", options.CacheDirectoryPath);
+        }
+
+        private static void Log(IDiagnosticLogger? logger, string option, object value)
+        {
+            logger?.Log(SentryLevel.Debug, "Setting Sentry {0} to: {1}", null, option, value);
+        }
```
bruno-garcia, line 42: Nit:
```suggestion
        => logger?.Log(SentryLevel.Debug, "Setting Sentry {0} to: {1}", null, option, value);
```

bruno-garcia (approved):
I have a couple of C# nits but it already LGTM!

semuserable (approved):

src/Sentry.Unity/Integrations/UnityBeforeSceneLoadIntegration.cs
bitsandfoxes: I think in this particular case, it doesn't work because `application` gets passed as an argument and we want to set `_application`?

bruno-garcia: Sorry, I missed that.

PR #211, opened 2021-06-04
Fix: Set default before invoking options

bitsandfoxes:
Aims to fix two issues: 

1. Setting defaults *before* invoking the options so the user can read/overwrite them properly
2. Log options *after* they have been overwritten so the log output makes sense

### DIFF
--- a/src/Sentry.Unity/SentryOptionsUtility.cs
+++ b/src/Sentry.Unity/SentryOptionsUtility.cs
@@ -1,5 +1,6 @@
 using Sentry.Extensibility;
 using Sentry.Unity.Integrations;
+using UnityEngine;
 
 namespace Sentry.Unity
 {
@@ -24,9 +25,17 @@ namespace Sentry.Unity
                 return;
             }
 
-            Log(logger, "Release", options.Release);
-            Log(logger, "Environment", options.Environment);
-            Log(logger, "Cache Directory", options.CacheDirectoryPath);
+            LogOption(logger, "Release", options.Release);
+            LogOption(logger, "Environment", options.Environment);
+
+            if (options.CacheDirectoryPath == null)
+            {
+                logger.Log(SentryLevel.Debug, "Offline Caching disabled.");
+            }
+            else
+            {
+                LogOption(logger, "Cache Directory", options.CacheDirectoryPath);
+            }
         }
 
         private static void SetRelease(SentryUnityOptions options, IApplication application)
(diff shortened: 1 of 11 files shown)


src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -1,42 +1,62 @@
...
 
             SetRelease(options, application);
             SetEnvironment(options, application);
             SetCacheDirectoryPath(options, application);
         }
 
+        public static void LogOptions(SentryUnityOptions options)
+        {
+            var logger = options.DiagnosticLogger;
+            if (logger == null)
```
bruno-garcia, line 23:
```suggestion
            if (logger is null)
```

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -1,42 +1,62 @@
...
+            {
+                return;
+            }
+
+            LogOption(logger, "Release", options.Release);
+            LogOption(logger, "Environment", options.Environment);
```
bruno-garcia, line 29: Could this be misleading?

We're only setting if the value isn't set yet, right? The log message is 'setting X to Y'

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -1,42 +1,62 @@
...
 
         private static void SetCacheDirectoryPath(SentryUnityOptions options, IApplication application)
         {
             options.CacheDirectoryPath ??= application.PersistentDataPath;
-            Log(options.DiagnosticLogger, "Cache Directory", options.CacheDirectoryPath);
         }
 
-        private static void Log(IDiagnosticLogger? logger, string option, object value)
-            => logger?.Log(SentryLevel.Debug, "Setting Sentry {0} to: {1}", null, option, value);
+        private static void LogOption(IDiagnosticLogger logger, string option, object value)
```
bruno-garcia, line 59: `object value` will box (allocate) any value type. Ideally we'd take `<T>` and first check if Logger.IsEnabled for that level before we allocate memory (otherwise it's just a waste).

The .NET SDK has proper extension methods which was made public in the .NET SDK 3.4.0 so if we bump the sub module here we can use it.

No more `logger.Log(level, ..`. We just use `LogDebug(..` at that point

https://github.com/getsentry/sentry-dotnet/blob/1095abca85eeefaac7170e7dcf482ffc052d53c9/src/Sentry/Extensibility/DiagnosticLoggerExtensions.cs#L11

bruno-garcia:
I opened a PR that changes the logging bits:

https://github.com/getsentry/sentry-unity/pull/214

src/Sentry.Unity/SentryOptionsUtility.cs
bitsandfoxes: My thinking is: We're setting lots of things by default because we expect there to be a value. So if the user overrides the value with empty it should look odd. 
But we could log it explicitly like with "offline caching disabled"?

semuserable (approved):

PR #212, opened 2021-06-04
Platform specific inits

bitsandfoxes:
Would that work for us?

#skip-changelog


bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-unity/pull/272

PR #213, opened 2021-06-04
Getting started window on installation

bitsandfoxes:
Popup after package installation:
We request the user to enter their DSN (maybe even with a link where they'd get it from).
Button to press to activate/enable the SDK & close the window? If no DSN is provided and the window is closed the SDK could just stay silent and disabled?

Screenshot 2021-06-04 at 13 06 03

#skip-changelog

### DIFF
--- a/src/Sentry.Unity.Editor/SentryAssetPostprocessor.cs
+++ b/src/Sentry.Unity.Editor/SentryAssetPostprocessor.cs
@@ -1,0 +1,52 @@
+using System.Linq;
+using System.Threading;
+using UnityEditor;
+using UnityEditor.PackageManager;
+using UnityEngine;
+
+namespace Sentry.Unity.Editor
+{
+    public class SentryAssetPostprocessor : AssetPostprocessor
+    {
+        private static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths)
+        {
+            var imported = importedAssets.Any(path => path.StartsWith("Packages/"));
+            var deleted = deletedAssets.Any(path => path.StartsWith("Packages/"));
+
+            if (imported || deleted)
+            {
+                InitializeOnLoad();
+            }
+        }
+
+        /**
+         * Instead of hooking into AssetPostprocessing we could InitializeOnLoad
+         * But this gets called on domain reload (i.e. enter playmode) and could be disabled by the user.
+         */
+        // [InitializeOnLoadMethod]
+        private static void InitializeOnLoad()
+        {
+            // ASYNC! Fetching all packages the current project depends on
+            var listRequest = Client.List(true);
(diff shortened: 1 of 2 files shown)


src/Sentry.Unity.Editor/SentryAssetPostprocessor.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Sentry.Unity.Editor
+{
+    public class SentryAssetPostprocessor : AssetPostprocessor
+    {
+        private static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths)
+        {
+            var imported = importedAssets.Any(path => path.StartsWith("Packages/"));
+            var deleted = deletedAssets.Any(path => path.StartsWith("Packages/"));
+
+            if (imported || deleted)
```
bruno-garcia, line 16: Do we want to prompt also when asset deleted? I wonder the chance of false positive here

src/Sentry.Unity.Editor/SentryAssetPostprocessor.cs
```diff
@@ -0,0 +1,52 @@
...
+         * But this gets called on domain reload (i.e. enter playmode) and could be disabled by the user.
+         */
+        // [InitializeOnLoadMethod]
+        private static void InitializeOnLoad()
+        {
+            // ASYNC! Fetching all packages the current project depends on
+            var listRequest = Client.List(true);
+            while (!listRequest.IsCompleted)
+            {
+                Thread.Sleep(100);
```
bruno-garcia, line 33: This might freeze up the editor though, right?
Can we schedule the continuation to run async too? If this was TPL it would be `listRequest.ContinuesWith(p => ..` but I'm not sure the type here (need to load the editor to do a proper review).

bruno-garcia (commented):
This is awesome!!
I know it's a draft to serve as a PoC but some notes to discuss the strategy to trigger it

bruno-garcia (changes requested):
Could we use `Events.registeringPackages += RegisteringPackagesEventHandler;` instead? 

Seems like this API would be perfect:

https://docs.unity3d.com/Manual/upm-api.html

bitsandfoxes:
> Could we use `Events.registeringPackages += RegisteringPackagesEventHandler;` instead?
> 
> Seems like this API would be perfect:
> 
> https://docs.unity3d.com/Manual/upm-api.html

Yes, definitely! This is way better.

src/Sentry.Unity.Editor/SentryAssetPostprocessor.cs
bitsandfoxes: Maybe not a prompt but maybe we should delete the assets we create too? Like link.xml and the options file?

PR #216, opened 2021-06-08
Build mac/windows IL2CPP player

bruno-garcia:
Build in CI through msbuild.

This is the first step of our verification. It only verifies that it successful builds which already spot an issue with the missing `.meta` file that was breaking IL2CPP builds.

TODO:
- [ ] Validate the IL2CPP player was built successfully (ideally run it with an argument and check its exit code

#skip-changelog


bruno-garcia:
@bitsandfoxes @semuserable Here's the fix for the IL2CPP builds (at least using `package-dev`).
Also started with the il2cpp build but still need to validate the output of the buid in CI. So far tested locally and worked on my machine ™️

semuserable (approved):

bruno-garcia:
Lets merge as-is and I'll add the validation step on a new PR

Issue #217, opened 2021-06-08
Compress payload 'Optimal', 'Fastest' don't work with il2cpp


semuserable:
Enable Sentry debug mode

![image](https://user-images.githubusercontent.com/2175097/121143979-41ac2a00-c846-11eb-8e4b-870ab5ba3d0c.png)

The following exception is received when an event is sent. This is an **excerpt** and not the full one.

```
Sentry: Error
Error while processing event 3afb1376ffb64aa5b16bc4f2ab5bf817: System.IO.IOException: Internal error (no progress possible) Flush
  at System.IO.Compression.DeflateStreamNative.CheckResult (System.Int32 result, System.String where) [0x00000] in <00000000000000000000000000000000>:0 
  at System.IO.Compression.DeflateStream.Flush () [0x00000] in <00000000000000000000000000000000>:0 
  at System.Action`1[T].Invoke (T obj) [0x00000] in <00000000000000000000000000000000>:0 
  at System.Threading.Tasks.Task.Execute () [0x00000] in <00000000000000000000000000000000>:0 
  at System.Threading.ContextCallback.Invoke (System.Object state) [0x00000] in <00000000000000000000000000000000>:0 
  at System.Threading.ExecutionContext.RunInternal (System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, System.Object state, System.Boolean preserveSyncCtx) [0x00000] in <00000000000000000000000000000000>:0 
  at System.Threading.Tasks.Task.ExecuteWithThreadLocal (System.Threading.Tasks.Task& currentTaskSlot) [0x00000] in <00000000000000000000000000000000>:0 
  at System.Threading.Tasks.Task.ExecuteEntry (System.Boolean bPreventDoubleExecution) [0x00000] in <00000000000000000000000000000000>:0 
  at System.Threading.ThreadPoolWorkQueue.Dispatch () [0x00000] in <00000000000000000000000000000000>:0 
--- End of stack trace from previous location where exception was thrown ---
```

PR #219, opened 2021-06-09
Force 'Auto' from sentry window

semuserable:
closes https://github.com/getsentry/sentry-unity/issues/217

I decided to force `Auto` if you configure from sentry window. We can differentiate between scripting backends with the following code

```csharp
PlayerSettings.GetScriptingBackend(EditorUserBuildSettings.selectedBuildTargetGroup) == ScriptingImplementation.IL2CPP
```

but it's available only in `Unity.Editor.dll`. We could do that logic not on options level, but on window level itself.

Which path should we take?

**PS**. Please notice, I didn't include tests as I'm trying to understand which approach we take.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Force 'Auto' from sentry window ([#219](https://github.com/getsentry/sentry-unity/pull/219))
+
 ### Features
 
 - Offline caching ([#208](https://github.com/getsentry/sentry-unity/pull/208))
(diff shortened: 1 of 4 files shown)


semuserable:
updated

bruno-garcia (commented):
We can merge this but I used the programatic approach and turns out it doesn't default to `Auto` there so it ends up breaking IL2CPP.

So basically what we need is to always set `base.CompressionLevel = NoCompression` through the `SetDefaults` method so that's done regardless of how `SentryUnityOptions` is created

semuserable:
Interesting. Ok, will move it into defaults then.

semuserable:
Changed to `NoCompression` and moved into defaults.

cc @bruno-garcia

bitsandfoxes (approved):

PR #220, opened 2021-06-09
Config: default & serialization

bitsandfoxes:
Two main goals:

1. Set default values at the appropriate time
2. Set up a user-facing options object to deal with overrides and other limitations (i.e. serialization)

I created a new class for the user-facing options object and decided to go with unity's scriptable object. That way we completely avoid any serialization/deserialization on our part + it's editable in the inspector (if we choose to support that).

### DIFF
--- a/src/Sentry.Unity.Editor/SentryTestWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryTestWindow.cs
@@ -14,8 +14,7 @@ namespace Sentry.Unity.Editor
         public void Dispose()
         {
             Close(); // calls 'OnLostFocus' implicitly
-            // File.Delete(SentryUnityOptions.GetConfigPath(SentryOptionsAssetName));
-            AssetDatabase.Refresh();
+            AssetDatabase.DeleteAsset(ScriptableSentryUnityOptions.GetConfigPath(SentryOptionsAssetName));
         }
     }
 }
--- a/src/Sentry.Unity.Editor/SentryTestWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryTestWindow.cs
@@ -14,7 +14,7 @@ namespace Sentry.Unity.Editor
         public void Dispose()
         {
             Close(); // calls 'OnLostFocus' implicitly
-            File.Delete(SentryUnityOptions.GetConfigPath(SentryOptionsAssetName));
+            // File.Delete(SentryUnityOptions.GetConfigPath(SentryOptionsAssetName));
             AssetDatabase.Refresh();
         }
     }
--- a/src/Sentry.Unity.Editor/SentryWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryWindow.cs
@@ -68,10 +68,6 @@ namespace Sentry.Unity.Editor
                 new GUIContent("Event Sample Rate", "What random sample rate to apply. 1.0 captures everything, 0.7 captures 70%."),
                 Options.SampleRate, 0.01f, 1);
 
-            Options.RequestBodyCompressionLevel = (CompressionLevelWithAuto)EditorGUILayout.EnumPopup(
-                new GUIContent("Compress Payload", "The level of which to compress the Sentry event before sending to Sentry."),
-                Options.RequestBodyCompressionLevel);
-
             Options.AttachStacktrace = EditorGUILayout.Toggle(
                 new GUIContent("Stacktrace For Logs", "Whether to include a stack trace for non error events like logs. " +
                                                       "Even when Unity didn't include and no Exception was thrown.."),
--- a/src/Sentry.Unity.Editor/SentryWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryWindow.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using Sentry.Extensibility;
+using Sentry.Unity.Json;
 using UnityEditor;
 using UnityEngine;
 
@@ -33,7 +34,7 @@ namespace Sentry.Unity.Editor
             var options = AssetDatabase.LoadAssetAtPath(
                 ScriptableSentryUnityOptions.GetConfigPath(SentryOptionsAssetName), typeof(ScriptableSentryUnityOptions)) as ScriptableSentryUnityOptions;
 
-            if (options == null)
+            if (options is null)
             {
                 CreateConfigDirectory();
                 options = CreateScriptableSentryUnityOptions();
@@ -52,7 +53,7 @@ namespace Sentry.Unity.Editor
             }
 
             var scriptableOptions = CreateScriptableSentryUnityOptions();
-            JsonSentryUnityOptions.ConvertToScriptable(sentryOptionsTextAsset, scriptableOptions);
+            JsonSentryUnityOptions.ToScriptableOptions(sentryOptionsTextAsset, scriptableOptions);
 
             EditorUtility.SetDirty(scriptableOptions);
             AssetDatabase.SaveAssets();
@@ -171,8 +172,7 @@ namespace Sentry.Unity.Editor
             var validationError = new ValidationError(fullFieldName, "Invalid DSN format. Expected a URL.");
             OnValidationError(validationError);
 
-            new UnityLogger(new SentryOptions{DiagnosticLevel = SentryLevel.Warning})
-                .LogWarning(validationError.ToString());
+            new UnityLogger(new SentryOptions()).LogWarning(validationError.ToString());
         }
 
         /// <summary>
--- a/src/Sentry.Unity.Editor/SentryWindow.cs
+++ b/src/Sentry.Unity.Editor/SentryWindow.cs
@@ -1,6 +1,5 @@
 using System;
 using System.IO;
-using Sentry.Extensibility;
 using UnityEditor;
 using UnityEngine;
 
@@ -14,89 +13,37 @@ namespace Sentry.Unity.Editor
         public static SentryWindow OpenSentryWindow()
             => (SentryWindow)GetWindow(typeof(SentryWindow));
 
-        protected virtual string SentryOptionsAssetName { get; } = SentryUnityOptions.ConfigName;
+        protected virtual string SentryOptionsAssetName { get; } = ScriptableSentryUnityOptions.ConfigName;
 
-        public SentryUnityOptions Options { get; set; } = null!; // Set by OnEnable()
+        public ScriptableSentryUnityOptions Options { get; private set; } = null!; // Set by OnEnable()
 
         public event Action<ValidationError> OnValidationError = _ => { };
 
         private void OnEnable()
         {
             SetTitle();
-
-            TryCreateSentryFolder();
-
-            Options = LoadUnitySentryOptions();
+            Options = LoadOptions();
         }
 
-        private static SentryUnityOptions LoadUnitySentryOptions()
+        private ScriptableSentryUnityOptions LoadOptions()
         {
-            var options = SentryUnityOptions.LoadFromUnity();
-            if (options != null)
+            var options = AssetDatabase.LoadAssetAtPath(
+                ScriptableSentryUnityOptions.GetConfigPath(), typeof(ScriptableSentryUnityOptions)) as ScriptableSentryUnityOptions;
+
+            if (options == null)
             {
-                return options;
+                CreateConfigDirectory();
+                options = CreateInstance<ScriptableSentryUnityOptions>();
+                SentryOptionsUtility.SetDefaults(options);
+
+                AssetDatabase.CreateAsset(options, ScriptableSentryUnityOptions.GetConfigPath());
+
+                EditorUtility.SetDirty(options);
+                AssetDatabase.SaveAssets();
+                AssetDatabase.Refresh();
             }
 
-            var unitySentryOptions = new SentryUnityOptions { Enabled = true };
-            unitySentryOptions
-                .TryAttachLogger()
-                .SaveToUnity(SentryUnityOptions.GetConfigPath());
-
-            return unitySentryOptions;
-        }
-
-        private void SetTitle()
-        {
-            var isDarkMode = EditorGUIUtility.isProSkin;
-            var texture = new Texture2D(16, 16);
-            using var memStream = new MemoryStream();
-            using var stream = GetType().Assembly
-                .GetManifestResourceStream($"Sentry.Unity.Editor.SentryLogo{(isDarkMode ? "Light" : "Dark")}.png");
-            stream.CopyTo(memStream);
-            stream.Flush();
-            memStream.Position = 0;
-            texture.LoadImage(memStream.ToArray());
-
-            titleContent = new GUIContent("Sentry", texture, "Sentry SDK Options");
-        }
-
-        private void Validate()
-        {
-            if (!Options.Enabled)
-            {
-                return;
-            }
-
-            ValidateDsn();
-        }
-
-        private void ValidateDsn()
-        {
-            if (Options.Dsn == null)
-            {
-                return;
-            }
-
-            if (Uri.IsWellFormedUriString(Options.Dsn, UriKind.Absolute))
-            {
-                return;
-            }
-
-            var fullFieldName = $"{nameof(Options)}.{nameof(Options.Dsn)}";
-            var validationError = new ValidationError(fullFieldName, "Invalid DSN format. Expected a URL.");
-            OnValidationError(validationError);
-            Debug.LogError(validationError.ToString());
-        }
-
-        /// <summary>
-        /// Called if window focus is lost or 'Close' is called
-        /// </summary>
-        public void OnLostFocus()
-        {
-            Validate();
-
-            Options.SaveToUnity(SentryUnityOptions.GetConfigPath());
-            AssetDatabase.Refresh();
+            return options;
         }
 
         // ReSharper disable once UnusedMember.Local
@@ -116,9 +63,10 @@ namespace Sentry.Unity.Editor
             Options.Dsn = EditorGUILayout.TextField(
                 new GUIContent("DSN", "The URL to your project inside Sentry. Get yours in Sentry, Project Settings."),
                 Options.Dsn);
+
             Options.SampleRate = EditorGUILayout.Slider(
                 new GUIContent("Event Sample Rate", "What random sample rate to apply. 1.0 captures everything, 0.7 captures 70%."),
-                Options.SampleRate ?? 1.0f, 0.01f, 1);
+                Options.SampleRate, 0.01f, 1);
 
             Options.RequestBodyCompressionLevel = (CompressionLevelWithAuto)EditorGUILayout.EnumPopup(
                 new GUIContent("Compress Payload", "The level of which to compress the Sentry event before sending to Sentry."),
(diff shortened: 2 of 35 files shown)


test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -98,6 +97,8 @@ public IEnumerator BugFarmScene_EventCaptured_IncludesApplicationVersionAsReleas
 
             // assert
             Assert.AreEqual(Application.version, testEventCapture.Events.First().Release);
+
+            // PlayerSettings.productName = originalProductName;
```
bruno-garcia, line 101: See: https://github.com/getsentry/sentry-unity/pull/230#pullrequestreview-686412931

```suggestion
            // PlayerSettings.productName = originalProductName;
        }

        [UnityTest]
        public IEnumerator BugFarmScene_EventCaptured_UserNameIsMachineNameWithDefaultPii()
        {
            yield return SetupSceneCoroutine("1_BugFarmScene");
        
            var testEventCapture = new TestEventCapture();
            using var _ = InitSentrySdk(o =>
            {
                o.AddIntegration(new UnityApplicationLoggingIntegration(eventCapture: testEventCapture));
                o.SendDefaultPii = true;
                o.IsEnvironmentUser = true;
            });
            var testBehaviour = new GameObject("TestHolder").AddComponent<TestMonoBehaviour>();
        
            testBehaviour.gameObject.SendMessage(nameof(testBehaviour.TestException));
        
            // assert
            Assert.Equals(Environment.UserName, testEventCapture.Events.First().User.Username);
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -160,20 +138,72 @@ private void OnGUI()
...
+
+            if (Uri.IsWellFormedUriString(Options.Dsn, UriKind.Absolute))
+            {
+                return;
+            }
+
+            var fullFieldName = $"{nameof(Options)}.{nameof(Options.Dsn)}";
+            var validationError = new ValidationError(fullFieldName, "Invalid DSN format. Expected a URL.");
+            OnValidationError(validationError);
+            Debug.LogError(validationError.ToString());
```
semuserable, line 174: Should we log here via Sentry logger?

src/Sentry.Unity/JsonSentryUnityOptions.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                options.Environment = environment.GetString();
+            }
+
+            return options;
+        }
+
+        public static void ConvertToScriptable(TextAsset sentryOptionsTextAsset, ScriptableSentryUnityOptions options)
+        {
+            Debug.Log("converting ");
```
semuserable, line 72: redundant?

src/Sentry.Unity/JsonSentryUnityOptions.cs
```diff
@@ -0,0 +1,95 @@
...
+            var jsonOptions = LoadFromJson(sentryOptionsTextAsset);
+
+            options.Enabled = jsonOptions.Enabled;
+
+            if (jsonOptions.Dsn is { } dsn)
+            {
+                Debug.Log($"{dsn}");
```
semuserable, line 79: redundant?

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Path for the config for Unity
+        /// </summary>
+        internal static string GetConfigPath(string? notDefaultConfigName = null)
+            => $"Assets/Resources/{ConfigRootFolder}/{notDefaultConfigName ?? ConfigName}.asset";
+
+        [field: SerializeField] public bool Enabled { get; set; }
+
+        [field: SerializeField] public bool CaptureInEditor { get; set; }
+
+        [field: SerializeField] public string Dsn { get; set; } = "";
```
semuserable, line 29: I prefer to use `string.Empty` instead of `""`

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -0,0 +1,94 @@
...
+            var options = new SentryUnityOptions();
+            SentryOptionsUtility.SetDefaults(options);
+
+            options.Enabled = scriptableOptions.Enabled;
+            options.CaptureInEditor = scriptableOptions.CaptureInEditor;
+            options.Dsn = scriptableOptions.Dsn;
+            options.SampleRate = scriptableOptions.SampleRate;
+            options.AttachStacktrace = scriptableOptions.AttachStacktrace;
+
+            if (!string.IsNullOrEmpty(scriptableOptions.ReleaseOverride))
```
semuserable, line 72: We should probably replace all `string.IsNullOrEmpty` with `string.IsNullOrWhiteSpace` calls.

src/Sentry.Unity.Editor/SentryTestWindow.cs
```diff
@@ -14,7 +14,7 @@ public static SentryTestWindow Open()
         public void Dispose()
         {
             Close(); // calls 'OnLostFocus' implicitly
-            File.Delete(SentryUnityOptions.GetConfigPath(SentryOptionsAssetName));
+            // File.Delete(SentryUnityOptions.GetConfigPath(SentryOptionsAssetName));
```
bruno-garcia, line 17: dead code. SHould we delete this?

src/Sentry.Unity/JsonSentryUnityOptions.cs
```diff
@@ -0,0 +1,93 @@
+using System.Text.Json;
+using Sentry.Unity.Extensions;
+using UnityEngine;
+
+namespace Sentry.Unity
+{
+    public static class JsonSentryUnityOptions
```
bruno-garcia, line 7:
```suggestion
    // This code exists for backward compatibility and will be removed in the future
    public static class JsonSentryUnityOptions
```

bruno-garcia (commented):
first pass

src/Sentry.Unity.Editor/SentryTestWindow.cs
semuserable: If this line is not used, than the class can be removed. The further call to `AssetDatabase.Refresh()` doesn't make sense too.

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -9,35 +9,105 @@ public static void SetDefaults(SentryUnityOptions options, IApplication? applica
...
+
+            options.EnableOfflineCaching = true;
+
+            options.Debug = Debug;
+            options.DebugOnlyInEditor = DebugOnlyInEditor;
+            options.DiagnosticLevel = DiagnosticLevel;
+        }
+
+        private static bool Enabled = true;
+        private static bool CaptureInEditor = true;
```
semuserable, line 54: Not sure I understand the need for separate variables for default values. Why not setting them directly?

test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -100,12 +99,33 @@ public IEnumerator BugFarmScene_EventCaptured_IncludesApplicationVersionAsReleas
             Assert.AreEqual(Application.version, testEventCapture.Events.First().Release);
         }
 
+        [UnityTest]
+        public IEnumerator BugFarmScene_EventCaptured_UserNameIsEnvironmentUserNameWithDefaultPii()
```
semuserable, line 103: Should this be a unit test instead?

test/Sentry.Unity.Tests/UnitySentryOptionsTests.cs
```diff
@@ -11,21 +13,22 @@ public sealed class UnitySentryOptionsTests
         private const string TestSentryOptionsFileName = "TestSentryOptions.json";
 
         [Test]
-        public void Options_ReadFromFile_Success()
+        public void Options_ReadFromJson_Success()
         {
             var optionsFilePath = GetTestOptionsFilePath();
+            Debug.Log(optionsFilePath);
```
semuserable, line 19: seems redundant

test/Sentry.Unity.Tests/UnitySentryOptionsTests.cs
```diff
@@ -11,21 +13,22 @@ public sealed class UnitySentryOptionsTests
...
             Assert.IsTrue(File.Exists(optionsFilePath));
 
-            var jsonRaw = File.ReadAllText(optionsFilePath);
-            using var jsonDocument = JsonDocument.Parse(jsonRaw);
+            Debug.Log("file existed");
```
semuserable, line 22: seems redundant

test/Sentry.Unity.Tests/UnitySentryOptionsTests.cs
```diff
@@ -11,21 +13,22 @@ public sealed class UnitySentryOptionsTests
...
+            var jsonTextAsset = new TextAsset(File.ReadAllText(GetTestOptionsFilePath()));
 
-            SentryUnityOptions.FromJson(jsonDocument.RootElement);
+            Debug.Log("parsed");
```
semuserable, line 25: seems redundant

test/Sentry.Unity.Tests/UnitySentryOptionsTests.cs
```diff
@@ -34,24 +37,39 @@ public void Options_WriteRead_Equals()
...
+
+        [Test]
+        public void Options_JsonAndScriptable_Equal()
+        {
+            var jsonTextAsset = new TextAsset(File.ReadAllText(GetTestOptionsFilePath()));
+            var scriptableOptions = ScriptableObject.CreateInstance<ScriptableSentryUnityOptions>();
+
+            var expectedOptions = JsonSentryUnityOptions.LoadFromJson(jsonTextAsset);
+
+            // Act
```
semuserable, line 68: probably should be removed or arrange, assert be added

src/Sentry.Unity/SentryOptionsUtility.cs
bruno-garcia: I think it'll be easier to read to just set all defaults straight in the method instead of declaring these static fields.

A single method where we can read through and easily spot all the default values

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -9,35 +9,105 @@ public static void SetDefaults(SentryUnityOptions options, IApplication? applica
...
         {
-            options.CacheDirectoryPath ??= application.PersistentDataPath;
-            options.DiagnosticLogger?.LogDebug("option.CacheDirectoryPath: {0}", options.CacheDirectoryPath);
+            var logger = options.DiagnosticLogger;
+            if (logger == null)
+            {
+                return;
+            }
+
+            // TODO: Add missing logs for options of interest
```
bruno-garcia, line 99: Alternatively we could implement `ToString` on the options object and just log out a single message after we process the instance. Instead of logging each value of each property.

```suggestion
            logger.LogDebug("Options: {0}", options);
```

test/Sentry.Unity.Tests/IntegrationTests.cs
bruno-garcia: I think it makes sense to be an integration test since it's toggling an option and verifying what the event processing pipeline did at the end (what user the event ended up with)

test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -100,12 +99,33 @@ public IEnumerator BugFarmScene_EventCaptured_IncludesApplicationVersionAsReleas
...
+            Assert.AreEqual(Environment.UserName, testEventCapture.Events.First().User.Username);
+        }
+
         [UnityTest]
         public IEnumerator BugFarmScene_EventCaptured_IncludesApplicationVersionAsRelease_WhenProductNameEmpty()
         {
             yield return SetupSceneCoroutine("1_BugFarmScene");
 
             // arrange
+            // var originalProductName = PlayerSettings.productName;
```
bruno-garcia, line 128: there's deadcode in this, forgot to delete?

test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -188,6 +210,20 @@ public IEnumerator BugFarmScene_MultipleSentryInit_SendEventForTheLatest()
...
 
+        [UnityTest]
+        public IEnumerator BugFarmScene_Init_OptionsAreDefaulted()
+        {
+            yield return SetupSceneCoroutine("1_BugFarmScene");
+
+            var expectedOptions = new SentryUnityOptions();
+            SentryOptionsUtility.SetDefaults(expectedOptions);
+
+            var actualOptions = new SentryUnityOptions();
```
bruno-garcia, line 221:
```suggestion
            SentryUnityOptions actualOptions = null!;
```

test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -188,6 +210,20 @@ public IEnumerator BugFarmScene_MultipleSentryInit_SendEventForTheLatest()
             Assert.AreEqual(1, nextEventCapture.Events.Count, nextDsn);
         }
 
+        [UnityTest]
+        public IEnumerator BugFarmScene_Init_OptionsAreDefaulted()
+        {
+            yield return SetupSceneCoroutine("1_BugFarmScene");
```
bruno-garcia, line 216: Do we need to load a scene for this test?

test/Sentry.Unity.Tests/UnitySentryOptionsTests.cs
bruno-garcia:
```suggestion
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -160,20 +137,74 @@ private void OnGUI()
             // EditorGUILayout.EndToggleGroup();
         }
 
+        private void OnLostFocus()
+        {
+            Validate();
+
+            EditorUtility.SetDirty(Options);
+            AssetDatabase.SaveAssets();
```
bruno-garcia, line 145: Should this be instead:

```suggestion
            if (Validate())
            {
                EditorUtility.SetDirty(Options);
                AssetDatabase.SaveAssets();
            }
```

src/Sentry.Unity/JsonSentryUnityOptions.cs
```diff
@@ -0,0 +1,94 @@
+using System.Text.Json;
+using Sentry.Unity.Extensions;
+using UnityEngine;
+
+namespace Sentry.Unity
```
bruno-garcia, line 5: We could move this to another namespace like `Sentry.Unity.Json` for example?

The motivation is to avoid exposing types that are not really "user facing" in the SDK root namespace where, for example, we have `SentryUnity` class.

An alternative is to make this stuff internal and use `InternalsVisibleTo` to Sentry.Unity.Editor.
But right now we're exposing a bunch of SDK internal bits that we don't expect to bump the major version if we change/break since they are not designed to be used by the users of the SDK

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        [field: SerializeField] public bool Debug { get; set; }
+        [field: SerializeField] public bool DebugOnlyInEditor { get; set; }
+        [field: SerializeField] public SentryLevel DiagnosticLevel { get; set; }
+
+        public static SentryUnityOptions? LoadSentryUnityOptions()
+        {
+            // TODO: Deprecated and to be removed once we update far enough.
+            var sentryOptionsTextAsset = Resources.Load<TextAsset>($"{ConfigRootFolder}/{ConfigName}");
+            if (sentryOptionsTextAsset != null)
```
bruno-garcia, line 46:
```suggestion
            if (sentryOptionsTextAsset is not null)
```

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        internal static SentryUnityOptions LoadFromScriptableObject(ScriptableSentryUnityOptions scriptableOptions)
+        {
+            var options = new SentryUnityOptions();
+            SentryOptionsUtility.SetDefaults(options);
+
+            options.Enabled = scriptableOptions.Enabled;
+            options.CaptureInEditor = scriptableOptions.CaptureInEditor;
+            options.Dsn = scriptableOptions.Dsn;
```
bruno-garcia, line 68: `scriptableOptions` has the default value of `string.Empty`. Isn't it going to override anything we make default (which is `null` for this field)

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Path for the config for Unity
+        /// </summary>
+        internal static string GetConfigPath(string? notDefaultConfigName = null)
+            => $"Assets/Resources/{ConfigRootFolder}/{notDefaultConfigName ?? ConfigName}.asset";
+
+        [field: SerializeField] public bool Enabled { get; set; }
+
+        [field: SerializeField] public bool CaptureInEditor { get; set; }
+
+        [field: SerializeField] public string Dsn { get; set; } = string.Empty;
```
bruno-garcia, line 29: Any reason why we default to `""` instead of leaving it as `null`?

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -12,10 +9,11 @@ public static class SentryInitialization
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void Init()
         {
-            var options = SentryUnityOptions.LoadFromUnity();
+            var options = ScriptableSentryUnityOptions.LoadSentryUnityOptions();
             if (options == null)
             {
-                new UnityLogger(SentryLevel.Warning).Log(SentryLevel.Warning, "Sentry has not been configured. You can do that through the editor: Tools -> Sentry");
+                new UnityLogger(new SentryOptions{DiagnosticLevel = SentryLevel.Warning}).LogWarning(
```
bruno-garcia, line 15:
```suggestion
                new UnityLogger(new SentryOptions()).LogWarning(
```

src/Sentry.Unity/SentryInitialization.cs
```diff
@@ -12,10 +9,11 @@ public static class SentryInitialization
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void Init()
         {
-            var options = SentryUnityOptions.LoadFromUnity();
+            var options = ScriptableSentryUnityOptions.LoadSentryUnityOptions();
             if (options == null)
```
bruno-garcia, line 13: `is null`

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -9,35 +10,66 @@ public static void SetDefaults(SentryUnityOptions options, IApplication? applica
         {
             application ??= ApplicationAdapter.Instance;
 
-            // 'Optimal' and 'Fastest' don't work on IL2CPP. Forcing 'NoCompression'.
+            options.Enabled = true;
+            options.Dsn = string.Empty;
```
bruno-garcia, line 14: Why don't we just leave this null?
```suggestion
```

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -9,35 +10,66 @@ public static void SetDefaults(SentryUnityOptions options, IApplication? applica
...
+            options.CaptureInEditor = true;
             options.RequestBodyCompressionLevel = CompressionLevelWithAuto.NoCompression;
+            options.AttachStacktrace = false;
+            options.StackTraceMode = default;
```
bruno-garcia, line 18: Can we please be explicit here? Even a comment on top of it (i.e: Enhanced isn't supported on IL2CPP)
```suggestion
            options.StackTraceMode = Original;
```

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -9,35 +10,66 @@ public static void SetDefaults(SentryUnityOptions options, IApplication? applica
...
+            options.SampleRate = 1.0f;
```
bruno-garcia, line 19: Do we need a default here? Ins't `null` the right default?
```suggestion
```

src/Sentry.Unity.Editor/SentryWindow.cs
bitsandfoxes: If the user closes the window we would throw away all their other changes just because the DSN is invalid.

PR #223, opened 2021-06-10
release health

bruno-garcia:
To figure out:

* When do we close the session gracefully? The current approach is limited on iOS
* Add opt-out for auto instrumentation
* Crashed sessions (since unity handles all exceptions)


bruno-garcia:
Superseded by #225

PR #225, opened 2021-06-10
Release health integration & Event-listener

bitsandfoxes:
So we can hook into Monobehaviour-Events like OnApplicationPause e.t.c.
Adds an invisible GameObject with the GameEventListener component to the active scene.


bruno-garcia (commented):
This seem pretty elegant to me. I'll give this a try with the release health stuff on Monday!

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -32,6 +31,7 @@ public static void Init(SentryUnityOptions unitySentryOptions)
             SentryOptionsUtility.SetDefaults(unitySentryOptions);
 
             SentrySdk.Init(unitySentryOptions);
+            SentrySdk.StartSession();
```
bruno-garcia, line 34: This approach will be taken by the .NET SDK already out of the box.
It'll introduce a `EnableAutoSessionTracking` that for the time being will be set to `false` by default. But once we release a new major version of the SDK for .NET we'll flip it to `true`.

So here we can set it to `false` before passing the options instance to `SentrySdk.Init` to make sure the .NET SDK is not automatically creating a session when `Init` is called and we're free to use the `MonoBehavior` approach without interference. 

See: https://github.com/getsentry/sentry-dotnet/issues/1054

src/Sentry.Unity/GameEventListener.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using UnityEngine;
+
+namespace Sentry.Unity
+{
+    public class GameEventListener : MonoBehaviour
```
bruno-garcia, line 6:
```suggestion
    /// <summary>
    ///  A MonoBehavior used to forward game events to subscribers.
    /// </summary>
    public class GameEventListener : MonoBehaviour
```

src/Sentry.Unity/GameEventListener.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        public event Action? ApplicationEnter;
```
bruno-garcia, line 8: Useful comment could be:

<summary>
Fires when a MonoBehavior OnApplicationPause is invoked 
</summary>
<remarks>
This hook doesn't fire on Android. Because ..
</remarks>

src/Sentry.Unity/GameEventListener.cs
```diff
@@ -0,0 +1,46 @@
...
+        private void OnApplicationPause(bool pauseStatus)
+        {
+            if (Application.platform == RuntimePlatform.Android)
+            {
+                return;
+            }
+
+            if (pauseStatus)
+            {
+                ApplicationExit?.Invoke();
```
bruno-garcia, line 20:
```suggestion
                ApplicationBackground?.Invoke();
```

src/Sentry.Unity/GameEventListener.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+            else
+            {
+                ApplicationEnter?.Invoke();
```
bruno-garcia, line 24:
```suggestion
                ApplicationForeground?.Invoke();
```

src/Sentry.Unity/Integrations/ITime.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using UnityEngine;
+
+namespace Sentry.Unity
+{
+    internal interface ITime
+    {
+        DateTime Now { get; }
```
bruno-garcia, line 8:
```suggestion
        DateTimeOffset Now();
```

bruno-garcia: Can be replaced by using `Sentry.Infrastructure.ISystemClock.GetUtcNow()`

And; https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Infrastructure/SystemClock.cs

src/Sentry.Unity/Integrations/ReleaseHealthIntegration.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+using Sentry.Integrations;
+using UnityEngine;
+
+namespace Sentry.Unity.Integrations
+{
+    public class ReleaseHealthIntegration : ISdkIntegration
```
bruno-garcia, line 7:
```suggestion
    public class SessionTrackingIntegration : ISdkIntegration
```

src/Sentry.Unity/Integrations/ReleaseHealthIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        private readonly ISceneManager _sceneManager;
+        private readonly ITime _time;
+
+        public ReleaseHealthIntegration() : this(SceneManagerAdapter.Instance, TimeAdapter.Instance)
```
bruno-garcia, line 12:
```suggestion
        public ReleaseHealthIntegration() : this(SceneManagerAdapter.Instance, SystemClock.Clock)
```

src/Sentry.Unity/Integrations/ReleaseHealthIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        internal ReleaseHealthIntegration(ISceneManager sceneManager, ITime time)
+        {
+            _sceneManager = sceneManager;
+            _time = time;
+        }
+
+        public void Register(IHub hub, SentryOptions options)
+        {
+            options.DiagnosticLogger?.Log(SentryLevel.Debug, "Registering SceneManager integration.");
```
bruno-garcia, line 24:
```suggestion
            options.DiagnosticLogger?.LogDebug("Registering SceneManager integration.");
```

src/Sentry.Unity/Integrations/ReleaseHealthIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        private void CreateEventListener(SceneAdapter fromScene, SceneAdapter toScene)
+        {
+            var gameListenerObject = new GameObject("SentryListener");
+            gameListenerObject.hideFlags = HideFlags.HideInHierarchy;
+
+            var gameListener = gameListenerObject.AddComponent<GameEventListener>();
+            gameListener.ApplicationEnter += OnApplicationEnter;
+            gameListener.ApplicationExit += OnApplicationExit;
+        }
```
bruno-garcia, line 37:
```suggestion
            _sceneManager.ActiveSceneChanged += CreateEventListener;

          void CreateEventListener(SceneAdapter fromScene, SceneAdapter toScene)
          {
              var gameListenerObject = new GameObject("SentryListener");
              gameListenerObject.hideFlags = HideFlags.HideInHierarchy;
  
              var gameListener = gameListenerObject.AddComponent<GameEventListener>();
              gameListener.ApplicationEnter += OnApplicationEnter;
              gameListener.ApplicationExit += OnApplicationExit;
          }
        }
```

src/Sentry.Unity/Integrations/ReleaseHealthIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        private void OnApplicationEnter()
+        {
+            // If a session has been running and not properly closed
```
bruno-garcia, line 41: An idea to have this work with the "app mode" session that the .NET SDK will introduce:

We don't restart the session on the first run (given that the SDK will already have initialized one). The goal is to have a session open even before the first event loop cycle in case there's an error, we can already mark that session as failed.

src/Sentry.Unity/Integrations/ReleaseHealthIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+                    Debug.Log("Session end");
+                    // SentrySdk.StartSession();
+                    Debug.Log("New Session start");
+                }
+            }
+            else
+            {
+                // SentrySdk.StartSession();
+                Debug.Log("Session start");
+            }
```
bruno-garcia, line 63: Prefer `DateTimeOffset.TryParse` over `Convert.ToDateTime(dateTimeString);` to handle bad serialized/missing data:

```suggestion
            const string exitTime = "ExitTime";
            if (PlayerPrefs.HasKey(exitTime))
            {
                var dateTimeString = PlayerPrefs.GetString(exitTime);
                PlayerPrefs.DeleteKey(exitTime);

                var lastExit = Convert.ToDateTime(dateTimeString);

                var inactiveDuration = _time.Now - lastExit;
                if (inactiveDuration.TotalSeconds >= 5.0f)
                {
                    // TODO: fetch session end status
                    // _hub.EndSession();
                    Debug.Log("Session end");
                    // _hub.StartSession();
                    Debug.Log("New Session start");
                }
            }
            else
            {
                // _hub.StartSession();
                Debug.Log("Session start");
            }
```

src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -11,6 +11,7 @@ public static void SetDefaults(SentryUnityOptions options, IApplication? applica
 
             // 'Optimal' and 'Fastest' don't work on IL2CPP. Forcing 'NoCompression'.
             options.RequestBodyCompressionLevel = CompressionLevelWithAuto.NoCompression;
+            options.EnableAutoSessionTracking = false;
```
bruno-garcia, line 14: In the .NET SDK this will be `false` by default until our next major version (to avoid surprising users with 'new data' being sent on a minor update. But here since we're on a experimental version (0.x) we can just default to true:

```suggestion
            options.EnableAutoSessionTracking = true;
```

src/Sentry.Unity/SentryUnity.cs
bruno-garcia: @bitsandfoxes and I discussed on a call and after this auto mode came in, it made sense to have both approaches working side-by-side. 

The .NET SDK will start a session on Init, and the Unity SDK would only be able to start after the first event loop/monobehavior event handlers are invoked so makes sense to have a session already started as early as possible (the .NET default beahvior) and the Unity integration wil make sure not to end it when the first "OnForeground" event kicks in.

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -80,6 +80,11 @@ public new CompressionLevelWithAuto RequestBodyCompressionLevel
             }
         }
 
+        /// <summary>
+        /// Whether the SDK should be tracking the session automatically.
+        /// </summary>
+        public bool EnableAutoSessionTracking { get; set; } = false;
```
bruno-garcia, line 86:
```suggestion
        /// <summary>
        /// Whether the SDK should be tracking the session automatically.
        /// </summary>
        /// <remakrs>
        /// This will come from the .NET SDK main options and can be removed in the future.
        /// </remarks>
        public bool EnableAutoSessionTracking { get; set; }
```

bruno-garcia (commented):
Great stuff! Just the ideas we discussed on a call

src/Sentry.Unity/GameEventListener.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+
+        private void OnApplicationFocus(bool hasFocus)
+        {
+
```
bruno-garcia, line 30:
```suggestion
```

src/Sentry.Unity/Integrations/ReleaseHealthIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+                PlayerPrefs.DeleteKey("ExitTime");
+
+                var lastExit = Convert.ToDateTime(dateTimeString);
+
+                var inactiveDuration = _time.Now - lastExit;
+                if (inactiveDuration.TotalSeconds >= 5.0f)
+                {
+                    // TODO: fetch session end status
+                    // SentrySdk.EndSession();
+                    Debug.Log("Session end");
```
bruno-garcia, line 54: The idea is to manage this in the .NET SDK and add the persistency in there.
So we can do: `SentrySdk.Session(Forground)` and background and use this code in any desktop/mobile app we want.

src/Sentry.Unity/ApplicationFocusListener.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+
+        /// <summary>
+        /// To receive Leaving/Resuming events on all platforms except Android.
+        /// </summary>
+        /// <param name="hasFocus"></param>
+        private void OnApplicationFocus(bool hasFocus)
+        {
+            // To avoid event duplication on Android
```
semuserable, line 60: What does it mean? Why duplication is only relevant to Android?

src/Sentry.Unity/Integrations/SessionIntegration.cs
```diff
@@ -0,0 +1,83 @@
...
+                        options.DiagnosticLogger?.LogDebug("Starting session.");
+                        return;
+                    }
+
+                    var dateTimeString = PlayerPrefs.GetString(LastLeaveKey);
+                    PlayerPrefs.DeleteKey(LastLeaveKey);
+
+                    if (!DateTimeOffset.TryParse(dateTimeString, out var lastApplicationLeave))
+                    {
+                        return;
```
semuserable, line 59: Probably log an error?

src/Sentry.Unity/ApplicationFocusListener.cs
bitsandfoxes: I added to the comment. According to testing and the Unity docs the OnAppliationFocus event is somewhat flaky on [Android](https://docs.unity3d.com/2019.4/Documentation/ScriptReference/MonoBehaviour.OnApplicationFocus.html#:~:text=On%20Android%2C%20when%20the%20on-screen%20keyboard%20is%20enabled%2C%20it%20causes%20an%20OnApplicationFocus(%20false%20)%20event.%20Additionally%2C%20if%20you%20press%20Home%20at%20the%20moment%20the%20keyboard%20is%20enabled%2C%20the%20OnApplicationFocus()%20event%20is%20not%20called%2C%20but%20OnApplicationPause()%20is%20called%20instead.), specifically in combination with the on-screen keyboard. It's possible to leave the app without receiving the focus event. That's why we trigger the SDK sessions via OnApplicationPause on Android instead.

src/Sentry.Unity/Integrations/SessionIntegration.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            options.DiagnosticLogger?.LogDebug("Registering Session integration.");
+
+            var gameListenerObject = new GameObject("SentryListener");
+            gameListenerObject.hideFlags = HideFlags.HideAndDontSave;
+
+            var gameListener = gameListenerObject.AddComponent<ApplicationPauseListener>();
+            gameListener.ApplicationResuming += () =>
+            {
+                Debug.Log("resuming");
```
semuserable, line 30: `options.DiagnosticLogger`?

src/Sentry.Unity/Integrations/SessionIntegration.cs
```diff
@@ -0,0 +1,46 @@
...
+        public void Register(IHub hub, SentryOptions options)
+        {
+            if (!options.AutoSessionTracking)
+            {
+                return;
+            }
+
+            options.DiagnosticLogger?.LogDebug("Registering Session integration.");
+
+            var gameListenerObject = new GameObject("SentryListener");
```
semuserable, line 24: Do we need to check if the object already created?

Example: user configured via Sentry window approach, but then decided to use a programmatic one without disabling window - 2 game objects in the scene. Or we don't support such cases?

semuserable (approved):

bruno-garcia (approved):
:shipit:

PR #230, opened 2021-06-14
Default IsEnvironmentUser = false;

bitsandfoxes:


src/Sentry.Unity/SentryOptionsUtility.cs
```diff
@@ -11,6 +11,7 @@ public static void SetDefaults(SentryUnityOptions options, IApplication? applica
 
             // 'Optimal' and 'Fastest' don't work on IL2CPP. Forcing 'NoCompression'.
             options.RequestBodyCompressionLevel = CompressionLevelWithAuto.NoCompression;
+            options.IsEnvironmentUser = false;
```
bruno-garcia, line 14: We could test that a captured event doesn't have `evt.User.UserName` equal to `Environment.MachineName` with the default `SentryUnityOptions`

bitsandfoxes: As discussed we're checking for Environment.UserName. See
https://github.com/getsentry/sentry-dotnet/blob/0f4358d1f37f664d7e144d33f6db5db842180726/src/Sentry/Internal/Enricher.cs#L80

test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -142,6 +142,45 @@ public IEnumerator BugFarmScene_EventCaptured_IncludesApplicationInEditorOrProdu
...
+        //         o.SendDefaultPii = true;
+        //         o.IsEnvironmentUser = true;
+        //     });
+        //     var testBehaviour = new GameObject("TestHolder").AddComponent<TestMonoBehaviour>();
+        //
+        //     testBehaviour.gameObject.SendMessage(nameof(testBehaviour.TestException));
+        //
+        //     // assert
+        //     Assert.Equals(Environment.UserName, testEventCapture.Events.First().User.Username);
+        // }
```
bruno-garcia, line 182: What about we remove it from here and add it to that other PR?
```suggestion
```

semuserable (approved):

Issue #49, opened 2021-02-06
Add Unity Context


bruno-garcia:
Values that could be part:

[`Application.stackTraceLogType`](https://docs.unity3d.com/ScriptReference/StackTraceLogType.html): Key `stackTraceLogType` and value should be its value to indicate what's being logged.

i.e: 

```
`@event.Contexts["Unity", new {
"stackTraceLogType": stackTraceLogType
};
```

There's a lot more context data that Unity surfaces via an API, we need to find out what's useful for troubleshooting and add to events:

![image](https://user-images.githubusercontent.com/1633368/107130901-174e5e00-68a0-11eb-95d7-2c9380ab5fb6.png)

Move all the `Extra` stuff into the context:
![image](https://user-images.githubusercontent.com/1633368/107230041-8b961800-69ec-11eb-882f-e7d0b8b222f5.png)

Consider that some inforamation Unity provides already fits our current protocol. For example `device.online` could be set based on Unity's `internetReachability`

Spec for protocol types: https://develop.sentry.dev/sdk/event-payloads/contexts/#device-context

bruno-garcia:
Suggestion for a breakdown of data;

```
### Device Context

* Application.genuine
* Application.internetReachability

### Application Context

### Unity Context
* Application.isEdtor

```

We can also consider low cardinality values to be used as tags. Such as `Application.internetReachability` which would help identify crashes that only happen when the device is offline, for example.

semuserable:
# [Existing context] Device

* [New] ProcessorCount - SystemInfo.processorCount
* [New] ProcessorType - SystemInfo.processorType
* [New] ProcessorFrequency - SystemInfo.processorFrequency
* [New] SupportsVibration - SystemInfo.supportsVibration
* [New] DeviceType - SystemInfo.deviceType
* [New | Unity] BatteryStatus - SystemInfo.batteryStatus
* [New] DeviceUniqueIdentifier - SystemInfo.deviceUniqueIdentifier
* [New] MaxTextureSize - SystemInfo.maxTextureSize
* [New] SupportsAccelerometer - SystemInfo.supportsAccelerometer
* [New] SupportsGyroscope - SystemInfo.supportsGyroscope
* [New] SupportsAudio - SystemInfo.supportsAudio
* [New] SupportsLocationService - SystemInfo.supportsLocationService
* [New] SupportsComputeShaders - SystemInfo.supportsComputeShaders
[...]

PR #234, opened 2021-06-22
Unity protocol

semuserable:
closes https://github.com/getsentry/sentry-unity/issues/49

~WIP~

### DIFF
--- a/src/Sentry.Unity/UnityEventProcessor.cs
+++ b/src/Sentry.Unity/UnityEventProcessor.cs
@@ -19,10 +19,12 @@ namespace Sentry.Unity
 
     internal class UnityEventProcessor : ISentryEventProcessor
     {
+        private readonly SentryOptions _sentryOptions;
         private readonly IApplication _application;
 
-        public UnityEventProcessor(IApplication? application = null)
+        public UnityEventProcessor(SentryOptions sentryOptions, IApplication? application = null)
         {
+            _sentryOptions = sentryOptions;
             _application = application ?? ApplicationAdapter.Instance;
         }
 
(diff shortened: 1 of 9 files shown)


src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -30,104 +28,119 @@ public UnityEventProcessor(IApplication? application = null)
...
+            sdk.AddPackage(UnitySdkInfo.PackageName, UnitySdkInfo.Version);
+            sdk.Name = UnitySdkInfo.Name;
+            sdk.Version = UnitySdkInfo.Version;
+        }
 
-            @event.Contexts.App.StartTime = DateTimeOffset.UtcNow.AddSeconds(-Time.realtimeSinceStartup);
+        private static void PopulateApp(App app)
+        {
+            app.StartTime = DateTimeOffset.UtcNow
+                // NOTE: Time API requires main thread
```
bruno-garcia, line 53: Can we do a runtime check? We better not run this if we're not on the main thread.
Fair point this is unrelated to the PR but rather broken right now

src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -30,104 +28,119 @@ public UnityEventProcessor(IApplication? application = null)
...
+            device.DeviceType = SystemInfo.deviceType.ToString();
+            device.CpuDescription = SystemInfo.processorType;
+            device.Timezone = TimeZoneInfo.Local;
+            device.ProcessorCount = SystemInfo.processorCount;
+            device.SupportsVibration = SystemInfo.supportsVibration;
+            device.Name = SystemInfo.deviceName;
+
+            // This is the approximate amount of system memory in megabytes.
+            // This function is not supported on Windows Store Apps and will always return 0.
+            device.MemorySize = SystemInfo.systemMemorySize;
```
bruno-garcia, line 79: Good comment. What about we check if it's `0` and avoid adding it altogether?

bruno-garcia (commented):
Looking pretty good for a draft btw

src/Sentry.Unity/UnityEventProcessor.cs
semuserable: Check if we're on the main thread before doing that?

src/Sentry.Unity/UnityEventProcessor.cs
semuserable: I noticed that it's used 2 times and we already have the logic to check against `0`, will remove that one altogether)

semuserable:
I think we should probably add `SystemInfo.graphicsShaderLevel` on `IApplication` level so we can unit test the conversion logic.

semuserable:
Ready to be reviewed.

bruno-garcia (approved):

PR #237, opened 2021-06-23
Un-Embedding the link.xml 

bitsandfoxes:
What happened so far:

1. Initially link.xml was inside Resources/Sentry and got shipped with the build
2. Wanted to move it somewhere else. Can't be inside /Packages because the managed code stripper doesn't look there.
3. Embedded it into the assembly. But that doesn't work with a wide range of Unity versions.
4. Placing it now in Assets/Plugins/Sentry


src/Sentry.Unity.Editor/SentryPreprocessBuild.cs
```diff
@@ -1,3 +1,4 @@
+using System.IO;
```
semuserable, line 1: discard?

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -175,6 +179,33 @@ private static void TryCreateSentryFolder()
...
+                AssetDatabase.CreateFolder("Assets/Plugins", "Sentry");
+            }
+
+            if (!AssetDatabase.IsValidFolder(LinkXmlPath))
+            {
+                using var fileStream = File.Create(LinkXmlPath);
+                using var resourceStream = GetType().Assembly.GetManifestResourceStream("Sentry.Unity.Editor.Resources.link.xml");
+                resourceStream.CopyTo(fileStream);
+
+                fileStream.Close();
```
semuserable, line 203: This seems redundant as you already have `using` on `File.Create`.

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -175,6 +179,33 @@ private static void TryCreateSentryFolder()
...
+                resourceStream.Flush();
```
semuserable, line 204: Same as above.

semuserable (approved):

bruno-garcia (approved):
Thanks!

PR #240, opened 2021-06-25
Lowercase true or false for tag value

semuserable:
I decided not to call `.ToLower()`, as in order to do so we need to convert `bool` to `string` first and then call `.ToLower()`.

Benefits: less allocations.

#skip-changelog


bitsandfoxes (approved):

bruno-garcia:
Good approach

PR #243, opened 2021-06-28
Config window options added

bitsandfoxes:
1. A new window with pretty much all of the Sentry Options. 
2. A custom inspector for our scriptable option object. That way we can have things like `string? dsn`.
3. Scriptable Options object fetches the default options from SentryUnityOptions and overrides only those that we want to be different in the config window (like debugging, or overrides)

ConfigWindow
ScriptableEditor


src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+                                                   "Information."),
+                Options.SendDefaultPii);
+
+            Options.IsEnvironmentUser = EditorGUILayout.Toggle(
+                new GUIContent("Auto Set UserName", "Whether to report the 'Environment.UserName' as " +
+                                                    "the User affected in the event. Should be disabled for " +
+                                                    "Android and iOS."),
+                Options.IsEnvironmentUser);
+
+            Options.ServerNameOverride = EditorGUILayout.TextField(
```
bruno-garcia, line 223: Do we even want to expose this? I this targeting running Unity on the server?

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+            EditorGUILayout.Space();
+
+            Options.MaxBreadcrumbs = EditorGUILayout.IntField(
+                new GUIContent("Max Breadcrumbs", "Maximum number of breadcrumbs that get captured." +
+                                                  "\nDefault: 100"),
+                Options.MaxBreadcrumbs);
+
+            Options.ReportAssembliesMode = (ReportAssembliesMode)EditorGUILayout.EnumPopup(
+                new GUIContent("Report Assemblies Mode", "Whether or not to include referenced assemblies " +
+                                                         "in each event sent to sentry."),
```
bruno-garcia, line 241: It's not just about include or not include but also about Version vs InformationalVersion
```suggestion
                new GUIContent("Report Assemblies Mode", "Whether or not to include referenced assemblies Version or InformationalVersion" +
                                                         "in each event sent to sentry."),
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+
+        private void DisplayTransport()
+        {
+            Options.EnableOfflineCaching = EditorGUILayout.BeginToggleGroup(
+                new GUIContent("Enable Offline Caching", ""),
+                Options.EnableOfflineCaching);
+            Options.MaxCacheItems = EditorGUILayout.IntField(
+                new GUIContent("Max Cache Items", "The maximum number of items to keep in cache. SDK " +
+                                                  "deletes the oldest and migrates the sessions to the next to " +
+                                                  "maintain the integrity of your release health stats.\nDefault: 30"),
```
bruno-garcia, line 253: Not sure it's worth talking about migrate session, it's too much internal stuff for the user at this point.
```suggestion
                new GUIContent("Max Cache Items", "The maximum number of files to keep in the disk cache. SDK " +
                                                  "deletes the oldest when limit is reached.\nDefault: 30"),
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+                Options.MaxCacheItems);
+
+            Options.InitCacheFlushTimeout = EditorGUILayout.IntField(
+                new GUIContent("Init Flush Timeout [ms]", "The timeout that limits how long the SDK " +
+                                                          "will attempt to flush existing cache during initialization."),
```
bruno-garcia, line 258: What about clarifying even further why this matters:

```suggestion
                new GUIContent("Init Flush Timeout [ms]", "The timeout that limits how long the SDK " +
                                                          "will attempt to flush existing cache during initialization.\nThis features allows capturing errors that happen during game startup and would not be captured because the process would be killed before Sentry had a chance to capture the event."),
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+                Options.InitCacheFlushTimeout);
+
+            EditorGUILayout.EndToggleGroup();
+
+            EditorGUILayout.Space();
+            EditorGUI.DrawRect(EditorGUILayout.GetControlRect(false, 1), Color.gray);
+            EditorGUILayout.Space();
+
+            // Options.RequestBodyCompressionLevel = (CompressionLevelWithAuto)EditorGUILayout.EnumPopup(
```
bruno-garcia, line 267:
```suggestion
            // Currently not supported on IL2CPP so avoiding exposing this:
            // Options.RequestBodyCompressionLevel = (CompressionLevelWithAuto)EditorGUILayout.EnumPopup(
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+
+            GUILayout.Label("Stacktrace", EditorStyles.boldLabel);
+
+            Options.AttachStacktrace = EditorGUILayout.Toggle(
+                new GUIContent("Stacktrace For Logs", "Whether to include a stack trace for non " +
+                                                      "error events like logs. Even when Unity didn't include and no " +
+                                                      "exception was thrown. Refer to AttachStacktrace on sentry docs."),
+                Options.AttachStacktrace);
+
+            // Options.StackTraceMode = (StackTraceMode) EditorGUILayout.EnumPopup(
```
bruno-garcia, line 202:
```suggestion
            // Enhanced not supported on IL2CPP so not displaying this for the time being:
            // Options.StackTraceMode = (StackTraceMode) EditorGUILayout.EnumPopup(
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+            // Options.RequestBodyCompressionLevel = (CompressionLevelWithAuto)EditorGUILayout.EnumPopup(
+            //     new GUIContent("Compress Payload", "The level of which to compress the Sentry event " +
+            //                                        "before sending to Sentry."),
+            //     Options.RequestBodyCompressionLevel);
+
+            var sampleRate = Options.SampleRate ??= 1.0f;
+            sampleRate = EditorGUILayout.Slider(
+                new GUIContent("Event Sample Rate", "Indicates the percentage of events that are " +
+                                                    "captured. Setting this to 0.1 captures 10% of events. " +
+                                                    "Setting this to 1.0 captures all events. Events are picked randomly."),
```
bruno-garcia, line 276:
```suggestion
                new GUIContent("Event Sample Rate", "Indicates the percentage of events that are " +
                                                    "captured. Setting this to 0.1 captures 10% of events. " +
                                                    "Setting this to 1.0 captures all events. This affects only errors and logs, not performance (transactions) data. See TracerSampleRate for that."),
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+                sampleRate, 0.01f, 1);
+            Options.SampleRate = (sampleRate < 1.0f) ? sampleRate : null;
+
+            Options.ShutdownTimeout = EditorGUILayout.IntField(
+                new GUIContent("Shut Down Timeout [ms]", "How many seconds to wait before shutting down to " +
+                                                         "give Sentry time to send events from the background queue."),
+                Options.ShutdownTimeout);
+
+            Options.MaxQueueItems = EditorGUILayout.IntField(
+                new GUIContent("Max Queue Items", "The maximum number of events to keep while the " +
```
bruno-garcia, line 286:
```suggestion
                new GUIContent("Max Queue Items", "The maximum number of events to keep in memory while the " +
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
+                                                  "worker attempts to send them."),
+                Options.MaxQueueItems
+            );
+        }
+
+        private void DisplaySessions()
+        {
+            Options.AutoSessionTracking = EditorGUILayout.BeginToggleGroup(
+                new GUIContent("Auto Session Tracking", "Whether the SDK should start a session " +
```
bruno-garcia, line 295: A bit misleading because we can also end if the app is on the background for 30 seconds or more. If the platform supports that.

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -63,85 +93,250 @@ private void CheckForAndConvertJsonConfig()
...
         private void OnLostFocus()
         {
+            if (Options is null)
+            {
+                return;
+            }
+
+            if (!AssetDatabase.IsValidFolder(ScriptableSentryUnityOptions.GetConfigPath(SentryOptionsAssetName)))
+            {
+                return;
```
bruno-garcia, line 337: When would this happen? Should we use the diagnostic logger here to let us know?

bitsandfoxes: I kept crashing Unity when I deleted the config while the window was still open and then closing it.

bruno-garcia: I see. Should we add that as a comment? I say that because just by eye balling the code it seemed redundant

bruno-garcia (approved):

Issue #238, opened 2021-06-23
Change approach for Unity main thread retrieval


semuserable:
We have the following line in `ApplicationAdapter`

```csharp
public bool IsMainThread => Thread.CurrentThread.ManagedThreadId == 1;
```

It seems we cant rely on `ManagedThreadId == 1`, suggested approach https://stackoverflow.com/questions/26452609/find-out-if-im-on-the-unity-thread. From Unity answers https://answers.unity.com/questions/62631/i-wonder-about-thread-id-plz-answer-me.html

Also, should we store this info on `IApplication` or put it somewhere else?

semuserable:
After investigating for a bit, I think it's quite "safe" to rely on `ManagedThreadId == 1`.

```csharp
Debug.Log($"[ThreadId.Main] {Thread.CurrentThread.ManagedThreadId}"); // always 1
Task.Run(() => Debug.Log($"[Thread.TaskRun] {Thread.CurrentThread.ManagedThreadId}")); // differs
new Thread(() => Debug.Log($"[Thread.ThreadStart] {Thread.CurrentThread.ManagedThreadId}")).Start(); // differs
```

semuserable:
After investigating for a bit, we can "save" the thread in any `MonoBehaviour.Start`. As a possible approach, we can create `GameObject` during `SentryInitialization.Init`, assign our `MonoBehaviour` and "save the thread" from there (`static` holder?).

cc @bruno-garcia

bruno-garcia:
That sounds like a good plan. We have a MonoBehavior that @bitsandfoxes created. We discussed renaming it to a more generic thing like `SentryUnityMonoBehavior` or something. And have `event` we can subscribe/unsubscribe from the SDK to do things.

I see we have `SentryBehavior` already in. Is that used?

Keeping the reference to the main thread from a MonoBehavior is a good approach. It would need to be nullable and while processing events we'd need consider the fact that the field wasn't set yet (MonoBehavior wasn't called yet or failed to hook)

semuserable:
`SentryBehaviour` is not used, no. Ok, will try to cook something up :)

Issue #233, opened 2021-06-16
Error on PS4 with il2cpp


d11ahirst:
### Environment

Plugin 0.2.0

How do you use Sentry?
Onpremise

Which SDK and version?
Net 2.0.0

### Steps to Reproduce

Initialised sentry plugin on PS4 Console on boot

### Expected Result

No Errors

What you thought would happen.

No Errors

### Actual Result

  Debug: Logging enabled with ConsoleDiagnosticLogger and min level: Debug
ArgumentOutOfRangeException: The added or subtracted value results in an un-representable DateTime.
Parameter name: value
  at Sentry.Internal.ProcessInfo..ctor (Sentry.SentryOptions options, System.Func`1[TResult] findPreciseStartupTime) [0x00000] in <00000000000000000000000000000000>:0
  at Sentry.SentrySdk.InitHub (Sentry.SentryOptions options) [0x00000] in <00000000000000000000000000000000>:0
  at Sentry.SentrySdk.Init (Sentry.SentryOptions options) [0x00000] in <00000000000000000000000000000000>:0
  at Bootstrap.Start () [0x00000] in <00000000000000000000000000000000>:0

(Filename: currently not available on il2cpp Line: -1)

What actually happened. Maybe a screenshot/recording? Maybe some logs?

bruno-garcia:
@d11ahirst what version of the SDK are you using?
I believe this was resolved on version 0.0.8 already: https://github.com/getsentry/sentry-unity/blob/main/CHANGELOG.md#fixes-8

`ProcessInfo` is not supported on IL2CPP so we turned that feature off for Unity.

d11ahirst:
Sorry for the confusion, we are using Unity 2019.3 with il2cpp on the PS4

semuserable:
Very interesting... My assumption. 

Does it mean that `DateTimeOffset.UtcNow` gets something like `01/01/0001` (ie `DateTimeOffset.MinValue`) on `PS4`? Sentry .NET "substracts" time in `ProcessInfo` during hub creation.

```csharp
var now = DateTimeOffset.UtcNow;
StartupTime = now;
try
{
    BootTime = now.AddTicks(-Stopwatch.GetTimestamp()
                            / (Stopwatch.Frequency
                               / TimeSpan.TicksPerSecond));
}
```

If we try to substract something from `DateTimeOffset.MinValue`, we recieve the mentioned exception.

bruno-garcia:
[...]

PR #245, opened 2021-06-30
Main thread data

semuserable:
~MVP~

Idea is the following: create singleton root game object from `SentryUnityOptions` (doesn't matter how many times you call `new()`), assign needed `MonoBehaviour`s on it and return them immediately. Access `MonoBehaviour` and pass needed data downstream.

Additionally, we can cache whatever data we want to acquire on the main thread and reuse it later (ex. `UnityEventProcessor`). In the initial commit, I made an example for `SystemInfo.operatingSystem` (which should be called on the main thread only).

The current PR is related to https://github.com/getsentry/sentry-unity/issues/238, https://github.com/getsentry/sentry-unity/issues/233


src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -1,14 +1,44 @@
...
+    internal partial class SentryMonoBehaviour
+    {
+        private static SentryMonoBehaviour? Instance;
+
+        // ReSharper disable once UnusedMember.Local
+        private void Awake()
+        {
+            // Unity overrides `==` operator in MonoBehaviours
+            // ReSharper disable once ConvertIfStatementToNullCoalescingAssignment
+            if (Instance is null)
```
bitsandfoxes, line 23: I would use '==' instead of 'is' especially since Unity overrides the operator. It probably does not matter in this case but I'd use '==' for all things MonoBehaviour.

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -64,19 +65,37 @@ public new CompressionLevelWithAuto RequestBodyCompressionLevel
...
+
+        private SentryMonoBehaviour CreateSentryMonoBehaviour()
+        {
+            if (_sentryMonoBehaviour is not null)
+            {
+                return _sentryMonoBehaviour;
+            }
+
+            // HideFlags.HideAndDontSave hides the GameObject in the hierarchy and prevents changing of scenes from destroying it
+            var rootGameObject = new GameObject("SentryMonoBehaviour") { hideFlags = HideFlags.DontUnloadUnusedAsset };
```
bitsandfoxes, line 97: 'HideFlags.DontUnloadUnusedAsset' so it will show up in the Hierarchy?

src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -93,6 +123,62 @@ internal void OnApplicationFocus(bool hasFocus)
         }
 
         // The GameObject has to destroy itself since it was created with HideFlags.HideAndDontSave
+        // ReSharper disable once UnusedMember.Local
```
bruno-garcia, line 126:
```suggestion
```

src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -1,14 +1,44 @@
...
+            }
+        }
+    }
+
     /// <summary>
     ///  A MonoBehavior used to forward application focus events to subscribers.
     /// </summary>
     [DefaultExecutionOrder(-900)]
-    internal class SentryMonoBehaviour : MonoBehaviour
+    [SuppressMessage("ReSharper", "ClassNeverInstantiated.Global")]
```
bruno-garcia, line 40:
```suggestion
```

src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -93,6 +123,62 @@ internal void OnApplicationFocus(bool hasFocus)
...
 
         // The GameObject has to destroy itself since it was created with HideFlags.HideAndDontSave
+        // ReSharper disable once UnusedMember.Local
         private void OnApplicationQuit() => Destroy(gameObject);
     }
+
+    /// <summary>
+    /// Main thread data collector.
+    /// </summary>
+    [SuppressMessage("ReSharper", "ClassNeverInstantiated.Global")]
```
bruno-garcia, line 133:
```suggestion
```

src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -93,6 +123,62 @@ internal void OnApplicationFocus(bool hasFocus)
...
+        {
+            get
+            {
+                _sentrySystemInfo ??= SentrySystemInfoAdapter.Instance;
+                return _sentrySystemInfo;
+            }
+            set => _sentrySystemInfo = value;
+        }
+
+        // ReSharper disable once UnusedMember.Local
```
bruno-garcia, line 149:
```suggestion
```

src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -163,20 +164,28 @@ static string ToGraphicShaderLevelDescription(int shaderLevel)
...
-                @event.SetTag("unity.device.unique_identifier", SystemInfo.deviceUniqueIdentifier);
+                @event.SetTag("unity.gpu.supports_instancing", _mainThreadData.SupportsDrawCallInstancing.Value ? "true" : "false");
+            }
+
+            if (_mainThreadData.DeviceType is not null)
+            {
+                @event.SetTag("unity.device.device_type", _mainThreadData.DeviceType);
+            }
+
+            if (_sentryOptions.SendDefaultPii && _mainThreadData.DeviceUniqueIdentifier is not null)
```
bruno-garcia, line 186: What about this to avoid the double read (which is `extern` so P/Invoke):

```suggestion
            if (_sentryOptions.SendDefaultPii && _mainThreadData.DeviceUniqueIdentifier is {} deviceUniqueId)
```

bruno-garcia (commented):
Looks great! I have just nit picks.

@bitsandfoxes thoughts?

src/Sentry.Unity/UnityEventProcessor.cs
semuserable: It's not extern anymore as it's cached in `MainThreadData` already.

PR #250, opened 2021-07-02
Invalid memory size on iOS

bitsandfoxes:
> contexts.device.memory_size: Discard invalid value expected an unsigned integer
> -334495744

On versions older than [2019.4.23 iOS seems to report memory sizes bigger than 2GB](https://issuetracker.unity3d.com/issues/ios-if-memory-of-device-is-2gb-over-systeminfo-dot-systemmemorysize-is-returned-wrong-value-on-ios) as some negative integer.

Check if memory size is > 0 before adding to the context.

Fixed with Unity 2019.4.23 onwards.

### DIFF
--- a/src/Sentry.Unity/UnityEventProcessor.cs
+++ b/src/Sentry.Unity/UnityEventProcessor.cs
@@ -108,7 +108,8 @@ namespace Sentry.Unity
 
             // This is the approximate amount of system memory in megabytes.
             // This function is not supported on Windows Store Apps and will always return 0.
-            if (SystemInfo.systemMemorySize != 0)
+            // Additionally: iOS issue with reporting negative memory size https://issuetracker.unity3d.com/issues/ios-if-memory-of-device-is-2gb-over-systeminfo-dot-systemmemorysize-is-returned-wrong-value-on-ios
+            if (SystemInfo.systemMemorySize > 0)
             {
                 device.MemorySize = SystemInfo.systemMemorySize * 1048576L; // Sentry device mem is in Bytes
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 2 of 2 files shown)


src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -108,7 +108,8 @@ private void PopulateDevice(Device device)
 
             // This is the approximate amount of system memory in megabytes.
             // This function is not supported on Windows Store Apps and will always return 0.
-            if (SystemInfo.systemMemorySize != 0)
+            // Additionally: iOS issue with reporting negative memory size https://issuetracker.unity3d.com/issues/ios-if-memory-of-device-is-2gb-over-systeminfo-dot-systemmemorysize-is-returned-wrong-value-on-ios
+            if (SystemInfo.systemMemorySize > 0)
             {
                 device.MemorySize = SystemInfo.systemMemorySize * 1048576L; // Sentry device mem is in Bytes
```
bruno-garcia, line 114: I didn't measure so this might not make sense but `systemMemorySize` is an `extern` method so to reduce to cost of P/Invoke we could do this:

```suggestion
            if (SystemInfo.systemMemorySize is {} systemMemorySize && systemMemorySize > 0)
            {
                device.MemorySize = mem * 1048576L; // Sentry device mem is in Bytes
```

There's probably a pattern to check directly the value being `>` instead of `is {} &&`.

bruno-garcia (approved):
Suggested one improvement but LGTM

src/Sentry.Unity/UnityEventProcessor.cs
semuserable: JFYI, this value will be stored in main thread data aka cached in my PR.

semuserable:
Let this PR (https://github.com/getsentry/sentry-unity/issues/238) supersede this fix. I'll implement it there.

Thanks!

PR #254, opened 2021-07-13
iOS: native support

bitsandfoxes:
First working iteration on our iOS Native Support:

- [x] Restore Cocoa submodule if missing
- [x] Build/Clean Cocoa SDK targets (dotnet msbuild /t:BuildCocoaSDK \ CleanCocoaSDK) 
- [x] Copy Sentry.framework into package-dev
- [x] Hook into Unity Xcode generation during build post process and add the Sentry SDK
- [x] Generate native options dictionary and fill it with the appropriate values read from ScriptableSentryOptions
- [x] Hook into main and initialize the SDK


src/Sentry.Unity.Editor.iOS/BuildPostprocess.cs
```diff
@@ -0,0 +1,86 @@
...
+                Path.Combine(FrameworkLocation, FrameworkName),
+                Path.Combine(FrameworkLocation, FrameworkName));
+
+            project.AddFileToBuild(targetGuid, fileGuid); // Ensures that the framework shows up on 'Link Binary with Libraries'
+            project.AddFileToEmbedFrameworks(targetGuid, fileGuid); // Embedding the framework because it's dynamic and needed at runtime
+
+            project.SetBuildProperty(targetGuid, "FRAMEWORK_SEARCH_PATHS", "$(inherited)");
+            project.AddBuildProperty(targetGuid, "FRAMEWORK_SEARCH_PATHS", $"$(PROJECT_DIR)/{FrameworkLocation}/");
+
+            // project.AddBuildProperty(targetGuid, "OTHER_LDFLAGS", "-ObjC");
```
bruno-garcia, line 52: I believe as we found out, we need this flag set:

```suggestion
            project.AddBuildProperty(targetGuid, "OTHER_LDFLAGS", "-ObjC");
```

src/Sentry.Unity.Editor.iOS/BuildPostprocess.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            project.WriteToFile(projectPath);
+        }
+
+        private static void ModifyMain(string pathToBuiltProject)
```
bruno-garcia, line 57: We could change this a bit but it works and if this isn't a reason for a build slow-down which is unlikely due to the size of `main.mm` I think more importantly is to have a test for it. 

Maybe we can add a `main.mm` file into a test in `Sentry.Unity.Edtior.iOS.Tests` and make sure this works. It could be done in a way that we name the `main.mm` file based on which version of Unity we took from. The assumption being that Unity might change the format/file in the future and we can just throw a new one and make sure the test passes while validating we're not regressing an older version. 

What do you think?

bitsandfoxes: Good idea. I'll make it so.

package-dev/Tests/Editor/TestFiles/main_2019_4_expected.txt
```diff
@@ -0,0 +1,33 @@
...
+        [ufw setExecuteHeader: &_mh_execute_header];
+    }
+    return ufw;
+}
+
+int main(int argc, char* argv[])
+{
+    @autoreleasepool
+    {
+		[SentrySDK startWithOptions:getSentryOptions()];
```
bruno-garcia, line 27: mix of tabs and spaces:

```suggestion
    {
		[SentrySDK startWithOptions:getSentryOptions()];
```

Can we replace our code to add `        ` instead of `\t\t`?

samples/unity-of-bugs/Assets/Plugins/Sentry/SentryOptions.txt
```diff
@@ -0,0 +1,15 @@
+#import <Foundation/Foundation.h>
+
+// IMPORTANT: DO NOT TOUCH! This file is generated during the Xcode project creation.
```
bruno-garcia, line 3:
```suggestion
// IMPORTANT: Changes to this file will be lost! 
// This file is generated during the Xcode project creation.
```

src/Sentry.Unity.Editor.iOS/MainModifier.cs
```diff
@@ -0,0 +1,71 @@
+using System.IO;
+using System.Text.RegularExpressions;
+using UnityEngine;
+
+namespace Sentry.Unity.Editor.iOS
+{
+    public static class MainModifier
+    {
+        private const string Include = "#include <Sentry/Sentry.h>\n#include \"SentryOptions.m\"\n";
+        private const string Init = "\t\t[SentrySDK startWithOptions:getSentryOptions()];\n\n";
```
bruno-garcia, line 10: What about (to replace the tabs with spaces):

```suggestion
        private const string Init = @"
        [SentrySDK startWithOptions:getSentryOptions()];

";
```

bruno-garcia (commented):
nice tests

test/Sentry.Unity.Editor.iOS.Tests/Sentry.Unity.Editor.iOS.Tests.csproj
```diff
@@ -0,0 +1,25 @@
...
+  </PropertyGroup>
+  <ItemGroup>
+    <None Update="TestFiles/*.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="../../src/Sentry.Unity/Sentry.Unity.csproj">
+      <Private>false</Private>
+    </ProjectReference>
```
bruno-garcia, line 14: What about this format? To have it in a single line:

```suggestion
    <ProjectReference Include="../../src/Sentry.Unity/Sentry.Unity.csproj" Private="false" />
```

src/Sentry.Unity.Editor.iOS/NativeMain.cs
```diff
@@ -0,0 +1,80 @@
...
+                return;
+            }
+
+            var sentryMain = AddSentryToMain(main);
+            if (sentryMain is null)
+            {
+                return;
+            }
+
+            File.WriteAllText(pathToMain, sentryMain);
```
bruno-garcia, line 40: There are 3 exit conditions before we add Sentry. It could be useful when troubleshooting to log Debug (or error on `AddSentryToMain` returning null?) to help us understand customers issues when they report the logs

src/Sentry.Unity.Editor.iOS/NativeMain.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            return true;
+        }
+
+        internal bool ContainsSentry(string main)
+        {
+            if (main.Contains(Include))
+            {
+                Debug.Log("Sentry already added to 'main.mm'.");
```
bruno-garcia, line 58: I wonder if we should pass `SentryOptions` in the constructor to this class (or just `IDiagnosticLogger`) via eh constructor so we can use our logger (and its configuration) instead of the Unity logger here?

src/Sentry.Unity.Editor.iOS/NativeOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        internal string Generate(SentryOptions options)
+        {
+            var nativeOptions = $@"#import <Foundation/Foundation.h>
+
+// IMPORTANT: Changes to this file will be lost!
+// This file is generated during the Xcode project creation.
+
+// TODO: make pretty with link to docs
```
bruno-garcia, line 27: 👍

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            _relativeFrameworkPath = GetRelativeFrameworkPath(_pathToProject);
+
+            _nativeMain = mainModifier ?? new NativeMain();
+            _sentryNativeOptions = sentryNativeOptions ?? new NativeOptions();
+        }
+
+        public static SentryXcodeProject Open(string path)
+        {
+            return new (path);
```
bruno-garcia, line 51: Hm, I thought the formatting for this was:

```suggestion
            return new(path);
```

Did Rider's 'auto format file' turned out as that?

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (string.IsNullOrEmpty(_relativeFrameworkPath))
+            {
+                Debug.LogWarning("Failed to locate the Sentry package within 'Frameworks'.");
+                return false;
+            }
+
+            var frameworkPath = Path.Combine(_pathToProject, _relativeFrameworkPath, FrameworkName);
+            if (!Directory.Exists(frameworkPath))
+            {
+                Debug.LogWarning($"Failed to locate 'Sentry.framework' within the Sentry package at '{frameworkPath}'.");
```
bruno-garcia, line 65: This happens if iOS support is disabled? I believe we avoid calling all of this.
In that case, shouldn't this be a LogError?

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            return true;
+        }
+
+        public void AddSentryFramework()
+        {
+            var targetGuid = _project.GetUnityMainTargetGuid();
+            var fileGuid = _project.AddFile(
+                Path.Combine(_relativeFrameworkPath, FrameworkName),
+                Path.Combine(_relativeFrameworkPath, FrameworkName));
```
bruno-garcia, line 77: Save up some string allocations:
```suggestion
            var filePath = Path.Combine(_relativeFrameworkPath, FrameworkName);
            var fileGuid = _project.AddFile(filePath, filePath);
```

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            // For dev purposes - The framework path contains the package name
+            relativeFrameworkPath += ".dev";
+            if (Directory.Exists(Path.Combine(pathToProject, relativeFrameworkPath)))
+            {
+                return Path.Combine(relativeFrameworkPath, "Plugins", "iOS");
+            }
+
+            return string.Empty;
```
bruno-garcia, line 121: is this a valid scenario? Should we throw an exception instead or log an error and return null?

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
bitsandfoxes: Yes! But maybe I got some wrong auto formatting settings?

bitsandfoxes: Nope, don't know how I got the space in between there.

src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
```diff
@@ -0,0 +1,36 @@
...
+            var options = ScriptableSentryUnityOptions.LoadSentryUnityOptions();
+            if (options is null || !options.IOSNativeSupportEnabled)
+            {
+                return;
+            }
+
+            var sentryXcodeProject = SentryXcodeProject.Open(pathToProject);
+            if (!sentryXcodeProject.ValidateFramework())
+            {
+                return;
```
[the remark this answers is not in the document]
bitsandfoxes: I thought I'd put the logging inside the Validate because we might want to check for a couple of things.

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -0,0 +1,128 @@
...
+            _relativeFrameworkPath = GetRelativeFrameworkPath(_pathToProject);
+
+            _nativeMain = mainModifier ?? new NativeMain();
+            _sentryNativeOptions = sentryNativeOptions ?? new NativeOptions();
+        }
+
+        public static SentryXcodeProject Open(string path)
+        {
+            return new(path);
+        }
```
bruno-garcia, line 52:
```suggestion
        public static SentryXcodeProject Open(string path) => new(path);
```

src/Sentry.Unity.Editor/SentryWindow.cs
```diff
@@ -159,16 +159,24 @@ private void DisplayCore()
...
-                Options.TracesSampleRate = (double)traceSampleRate;
+                Options.TracesSampleRate = (double) traceSampleRate;
             }
+
+            EditorGUILayout.Space();
+            EditorGUI.DrawRect(EditorGUILayout.GetControlRect(false, 1), Color.gray);
+            EditorGUILayout.Space();
+
+            Options.IOSNativeSupportEnabled = EditorGUILayout.Toggle(
+                new GUIContent("iOS Native Support", "Whether to enable Native Support."),
```
bruno-garcia, line 178:
```suggestion
                new GUIContent("iOS Native Support", "Whether to enable iOS Native Support."),
```

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -65,6 +65,13 @@ public new CompressionLevelWithAuto RequestBodyCompressionLevel
             }
         }
 
+        /// <summary>
+        /// Whether the SDK should add native support for iOS
+        /// </summary>
+        public bool IOSNativeSupportEnabled { get; set; } = true;
+
+        private SentryMonoBehaviour? _sentryMonoBehaviour;
+        internal Func<SentryMonoBehaviour> SentryMonoBehaviourGenerator;
```
bruno-garcia, line 74: Sorry I broke this when I merged main into ur branch

test/Sentry.Unity.Editor.iOS.Tests/NativeMainTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            var main = GetFileContents("main_2019_4.txt");
+            var nativeMain = new NativeMain();
+
+            var containsSentry = nativeMain.ContainsSentry(main);
+
+            Assert.IsFalse(containsSentry);
+        }
+
+        [Test]
+        public void AddSentryToMain_SentryNotAddedTo_2019_4_MatchesExceptedOutput()
```
bruno-garcia, line 54:
```suggestion
        public void AddSentryToMain_SentryNotAddedTo_2019_4_MatchesExpectedOutput()
```

test/Sentry.Unity.Editor.iOS.Tests/NativeMainTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            var expectedMain = GetFileContents("main_2019_4_expected.txt");
+            var nativeMain = new NativeMain();
+
+            var actualMain = nativeMain.AddSentryToMain(main);
+
+            Assert.AreEqual(expectedMain, actualMain);
+        }
+
+        [Test]
+        public void MainModifier_AddSentryToMain_SentryAdded()
```
bruno-garcia, line 66:
```suggestion
        public void AddSentry_AddSentryToMain_SentryAdded()
```

test/Sentry.Unity.Editor.iOS.Tests/NativeMainTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            File.Delete(workingMainPath);
+        }
+
+        private string GetFileContents(string fileName)
+        {
+            var assemblyPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var mainPath = Path.Combine(assemblyPath, "TestFiles", fileName);
+
+            return File.ReadAllText(mainPath);
+        }
```
bruno-garcia, line 87:
```suggestion
        private string GeTestFilePath(string fileName)
        {
            var assemblyPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
            return Path.Combine(assemblyPath, "TestFiles", fileName);
        }
        private string GetFileContents(string fileName)
        {
			var path = GeTestFilePath(fileName);
            return File.ReadAllText(mainPath);
        }
```

test/Sentry.Unity.Editor.iOS.Tests/Sentry.Unity.Editor.iOS.Tests.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <OutDir>$(PackageEditorTestsPath)</OutDir>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Update="TestFiles/*.txt">
```
bruno-garcia, line 7: what about we organize files by unity version as folder. Will be handy with the xcodeproj and other such files:

```suggestion
    <None Update="TestFiles/2021.1/*.txt">
```

test/Sentry.Unity.Editor.iOS.Tests/SentryXcodeProjectTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            var root = Path.GetTempPath() + Path.GetRandomFileName();
+            var expectedFrameworkPath = Path.Combine("Frameworks", "io.sentry.unity", "Plugins", "iOS");
+            Directory.CreateDirectory(Path.Combine(root, expectedFrameworkPath));
+            var xcodeProject = new SentryXcodeProject("", new NativeMainFixture(), new NativeOptionsFixture());
+
+            var actualFrameworkPath = xcodeProject.GetRelativeFrameworkPath(root);
+
+            Assert.AreEqual(expectedFrameworkPath, actualFrameworkPath);
+
+            Directory.Delete(root, true);
```
bruno-garcia, line 31: prob best with SetUp and TearDown methods

bruno-garcia (commented):
🚀

src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
```diff
@@ -0,0 +1,43 @@
...
+            }
+
+            // TODO: check for other criteria why we would stop touching the Xcode project
+            var options = ScriptableSentryUnityOptions.LoadSentryUnityOptions();
+            if (options is null)
+            {
+                return;
+            }
```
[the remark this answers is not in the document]
bitsandfoxes: I'll have to double check the "not calling twice". But it should (tm) work fine since we're using it 1x during runtime when we want to init and anytime else would be in the editor during building.

src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
```diff
@@ -0,0 +1,43 @@
...
+            try
+            {
+                using var sentryXcodeProject = SentryXcodeProject.Open(pathToProject, options);
+                sentryXcodeProject.AddSentryFramework();
+                sentryXcodeProject.AddNativeOptions();
+                sentryXcodeProject.AddSentryToMain();
+            }
+            catch (Exception e)
+            {
+                options.DiagnosticLogger?.LogError(e.Message);
```
bruno-garcia, line 39: What about passing the whole exception instance so it can format the stack trace?
```suggestion
                options.DiagnosticLogger?.LogError("Failed to add Sentry to the xcode project", e);
```

src/Sentry.Unity.Editor.iOS/NativeOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        // Native Diagnostic Level:
+        // None = 0
+        // Debug = 1
+        // Info = 2
+        // Warning = 3
+        // Error = 4
+        // Fatal = 5
+        internal static int ToNativeDiagnosticLevel(this SentryLevel diagnosticLevel) => (int)diagnosticLevel + 1;
+    }
```
bruno-garcia, line 54: These extension methods will be visible in this whole assembly. Since this is just used in `NativeOptions`, it' probably better to have it as a private method in that class.

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -0,0 +1,121 @@
...
+
+namespace Sentry.Unity.Editor.iOS
+{
+    internal class SentryXcodeProject : IDisposable
+    {
+        private const string FrameworkName = "Sentry.framework";
+
+        private const string MainPath = "MainApp/main.mm";
+        private const string OptionsPath = "MainApp/SentryOptions.m";
+        private string UnityPackageFrameworkRoot = Path.Combine("Frameworks", "io.sentry.unity");
```
bruno-garcia, line 15: It's a bit inconsistent, either go with `/` or use `Path.Combine` on all, like:

```suggestion
        private const string MainPath = Path.Combine("MainApp", "main.mm");
        private const string OptionsPath = Path.Combine("MainApp"," SentryOptions.m");
        private string UnityPackageFrameworkRoot = Path.Combine("Frameworks", "io.sentry.unity");
```

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        private readonly PBXProject _project;
+        private readonly string _projectPath;
+
+        internal string? RelativeFrameworkPath { get; set; }
+
+        private readonly INativeMain _nativeMain;
+        private readonly INativeOptions _nativeOptions;
+
+        internal SentryXcodeProject(
```
bruno-garcia, line 28: Communicates intent: This ctor is designed to be used by those with access to this class:

```suggestion
        public SentryXcodeProject(
```

If the class became public, nothing else needs to change.

And the other `internal` ctor is clear to be there for 'internal' (i.e: Tests or what not) reasons

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -0,0 +1,121 @@
...
+            xcodeProject.SetRelativeFrameworkPath();
+
+            return xcodeProject;
+        }
+
+        internal void ReadFromProjectFile()
+        {
+            if (!File.Exists(_projectPath))
+            {
+                throw new FileNotFoundException($"Could not locate generated Xcode project at {_projectPath}");
```
bruno-garcia, line 64: Second arg is the file name: https://docs.microsoft.com/en-us/dotnet/api/system.io.filenotfoundexception.-ctor?view=net-5.0#System_IO_FileNotFoundException__ctor_System_String_System_String_


```suggestion
                throw new FileNotFoundException($"Could not locate generated Xcode project", _projectPath);
```

src/Sentry.Unity/SentryUnityOptionsExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+using Sentry.Unity.Integrations;
+
+namespace Sentry.Unity
+{
+    public static class SentryUnityOptionsExtensions
+    {
+        public static bool ShouldInitializeSdk(this SentryUnityOptions? options)
+        {
+            return ShouldInitializeSdk(options, null);
+        }
```
bruno-garcia, line 11:
```suggestion
        public static bool ShouldInitializeSdk(this SentryUnityOptions? options)
            => ShouldInitializeSdk(options, null);
```

test/Sentry.Unity.Editor.iOS.Tests/NativeMainTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            var main = string.Empty;
+            var nativeMain = new NativeMain();
+
+            var x = Assert.Throws<ArgumentException>(() => nativeMain.AddSentryToMain(main));
+            Assert.AreEqual("main", x.ParamName);
+        }
+
+        [Test]
+        public void AddSentry_MainDoesNotExist_ThrowsFileNotFoundException()
```
bruno-garcia, line 55: Test name indicates an exception is thrown but test is a copy paste from test below

test/Sentry.Unity.Editor.iOS.Tests/NativeOptionsTests.cs
```diff
@@ -0,0 +1,67 @@
...
+namespace Sentry.Unity.Editor.iOS.Tests
+{
+    public class NativeOptionsTests
+    {
+        [Test]
+        public void GenerateOptions_NewSentryOptions_Compiles()
+        {
+            if (Environment.OSVersion.Platform != PlatformID.MacOSX)
+            {
+                Assert.Inconclusive("Skipping: Non MacOSX");
```
bruno-garcia, line 15: OSX is the old name, what about:

```suggestion
                Assert.Inconclusive("Skipping: Not on macOS");
```

test/Sentry.Unity.Editor.iOS.Tests/SentryXcodeProjectTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            public SentryXcodeProject GetSut() => new(ProjectRoot, Options, NativeMain, NativeOptions);
+        }
+
+        private Fixture _fixture = new();
+
+        [SetUp]
+        public void SetUp()
+        {
+            _fixture = new Fixture();
+        }
```
bruno-garcia, line 38:
```suggestion
        public void SetUp() => _fixture = new Fixture();
```

bruno-garcia (approved):
Awesome stuff! I left some minor notes but we could merge and address other things in follow up PRs

PR #256, opened 2021-07-15
DeviceModel set lazily (~ -20ms Sentry init); tests

semuserable:
#skip-changelog

### DIFF
--- a/src/Sentry.Unity/SystemInfoAdapter.cs
+++ b/src/Sentry.Unity/SystemInfoAdapter.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading;
+﻿using System;
+using System.Threading;
 using UnityEngine;
 
 namespace Sentry.Unity
@@ -13,7 +14,7 @@ namespace Sentry.Unity
         string? CpuDescription { get; }
         string? DeviceName { get; }
         string? DeviceUniqueIdentifier { get; }
-        string? DeviceModel { get; }
+        Lazy<string>? DeviceModel { get; }
         int? SystemMemorySize { get; }
         int? GraphicsDeviceId { get; }
         string? GraphicsDeviceName { get; }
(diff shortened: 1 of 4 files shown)


src/Sentry.Unity/SystemInfoAdapter.cs
```diff
@@ -48,7 +49,7 @@ private SentrySystemInfoAdapter()
         public string? CpuDescription => SystemInfo.processorType;
         public string? DeviceName => SystemInfo.deviceName;
         public string? DeviceUniqueIdentifier => SystemInfo.deviceUniqueIdentifier;
-        public string? DeviceModel => SystemInfo.deviceModel;
+        public Lazy<string> DeviceModel => new(() => SystemInfo.deviceModel);
```
bruno-garcia, line 52: Does it need to run in the UI thread?

bruno-garcia: Docs don't say anything about it. 
Maybe we could make all values that don't require running in the UI thread become Lazy<T>?

semuserable: > Does it need to run in the UI thread?
I also haven't found any info about `SystemInfo.deviceModel` to be called on UI thread.

> Maybe we could make all values that don't require running in the UI thread become Lazy?
We can, but I need to revisit all the properties... again.

bruno-garcia: > I also haven't found any info about SystemInfo.deviceModel to be called on UI thread.

Probably easy to test with `Task.Run(() => SystemInfo.deviceModel)` if that throws we know.

> We can, but I need to revisit all the properties... again)

Could be a new PR thingy, this solves already the main perf issue you've found right?

semuserable: > Could be a new PR thingy, this solves already the main perf issue you've found right?

Yes, when Sentry inits we no longer get `SystemInfo.deviceModel` call (~ 20ms), only when event is processed it's called and the result reused.

semuserable:
I also checked `Sentr.Dsn.Parse` and the idea to remove `new Uri`. I see no point of doing so as the whole logic relies on it. Otherwise, we need to re-implement the `Uri` parsing logic which is a strange thing to do. `20 ms` is not a deal breaker. At least I see no practical reason of doing so.

image

cc @bruno-garcia

bruno-garcia:
> I see no point of doing so as the whole logic relies on it.

Parsing this by hand is pretty straight forward for what we need and it was done in other SDKs. To brin that from 20 to 1 it might be worth it but work for the .NET SDK.

bruno-garcia (approved):

PR #262, opened 2021-07-17
log in single line

bruno-garcia:
To better support users who configure console to print a single line which currently shows like this:

image

image

Looks like this at the end: 
image


bruno-garcia:
With the default 2 lines:

image

With 1 line (pretty nice if u ask me):

image

PR #271, opened 2021-07-28
ci: verify build output

bruno-garcia:
#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,10 +137,10 @@ jobs:
           path: samples/artifacts/builds/Android
 
       - name: Build Standalone Player with IL2CPP
-        run: dotnet msbuild /t:UnityBuildPlayerIL2CPP /p:Configuration=Release
+        run: dotnet msbuild /t:UnityBuildStandalonePlayerIL2CPP /p:Configuration=Release
 
       - name: Run Smoke Tests
-        run: dotnet msbuild /t:UnitySmokeTestPlayerIL2CPP
+        run: dotnet msbuild /t:UnitySmokeTestStandalonePlayerIL2CPP
 
       - name: Prepare Sentry package for release
         run: |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ dotnet msbuild /t:"UnityPlayModeTest;UnityEditModeTest" /p:Configuration=Release
 ### Smoke test by building and running a player with IL2CPP:
 
 ```sh
- dotnet msbuild /t:"Build;UnitySmokeTestStandalonePlayerIL2CPP"
+ dotnet msbuild /t:"Build;UnityBuildStandalonePlayerIL2CPP;UnitySmokeTestStandalonePlayerIL2CPP"
 ```
 
 After this you can open your IDE (i.e: Visual Studio or Rider) and Unity Editor for development.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,10 +24,10 @@ Clone the repo `git clone https://github.com/getsentry/sentry-unity.git` and `cd
 dotnet msbuild /t:"UnityPlayModeTest;UnityEditModeTest" /p:Configuration=Release
 ```
 
-### Smoke test by building and running a player with IL2CPP::
+### Smoke test by building and running a player with IL2CPP:
 
 ```sh
- dotnet msbuild /t:"UnityBuildPlayerIL2CPP;UnitySmokeTestPlayerIL2CPP"
+ dotnet msbuild /t:"Build;UnitySmokeTestStandalonePlayerIL2CPP"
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -137,10 +137,10 @@ jobs:
           path: samples/artifacts/builds/Android
 
       - name: Build Standalone Player with IL2CPP
-        run: dotnet msbuild /t:UnityBuildPlayerIL2CPP /p:Configuration=Release
+        run: dotnet msbuild /t:UnityBuildStandalonePlayerIL2CPP /p:Configuration=Release
```
bruno-garcia, line 140: Now that we have "IL2CPP Players", naming it so it's easier to differentiate between Standalone, iOS, Android..

CONTRIBUTING.md
```diff
@@ -24,10 +24,10 @@ Clone the repo `git clone https://github.com/getsentry/sentry-unity.git` and `cd
 dotnet msbuild /t:"UnityPlayModeTest;UnityEditModeTest" /p:Configuration=Release
 ```
 
-### Smoke test by building and running a player with IL2CPP::
+### Smoke test by building and running a player with IL2CPP:
 
 ```sh
- dotnet msbuild /t:"UnityBuildPlayerIL2CPP;UnitySmokeTestPlayerIL2CPP"
+ dotnet msbuild /t:"Build;UnityBuildStandalonePlayerIL2CPP;UnitySmokeTestStandalonePlayerIL2CPP"
```
bruno-garcia, line 30: After `dotnet clean`, we need to `dotnet build` before running these custom targets (otherwise it fails due to missing `Sentry.Unity` and removes the `.meta`, etc.

One option is to set the `DependsOn` property so that we can run only the final target but that makes it harder to run them separately. That is, unless we add proper support for incremental builds:

https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-targets?view=vs-2019#incremental-builds

Directory.Build.targets
```diff
@@ -61,20 +62,19 @@ We need to run a unity build to restore the test packages and for that reason we
...
+  <Target Name="UnityBuildStandalonePlayerIL2CPP"
           Condition="'$(MSBuildProjectName)' == 'Sentry.Unity'">
     <Error Condition="$(UnityRoot) == ''" Text="Couldn't find Unity."></Error>
 
     <Message Importance="High" Text="Building Player with IL2CPP."></Message>
 
     <Exec Command="&quot;$(UnityExec)&quot; -quit -batchmode -nographics -logFile - -projectPath $(UnitySampleProjectPath) -executeMethod $(StandaloneBuildMethod) -buildPath $(StandaloneBuildPath)"
-          IgnoreStandardErrorWarningFormat="true"
-          IgnoreExitCode="true"></Exec>
+          IgnoreStandardErrorWarningFormat="true"></Exec>
```
bruno-garcia, line 73: Unity is returning `0` on successful builds on all my tests so I removed the `IgnoreExitCode="true"`.

So builds should break already based on Unity's return code now.

Directory.Build.targets
```diff
@@ -96,8 +96,9 @@ Related: https://forum.unity.com/threads/6572-debugger-agent-unable-to-listen-on
     <Message Importance="High" Text="Building Android Player with IL2CPP."></Message>
 
     <Exec Command="&quot;$(UnityExec)&quot; -quit -batchmode -nographics -logFile - -projectPath $(UnitySampleProjectPath) -executeMethod $(AndroidBuildMethod) -buildPath $(AndroidBuildPath)"
-          IgnoreStandardErrorWarningFormat="true"
-          IgnoreExitCode="true"></Exec>
+          IgnoreStandardErrorWarningFormat="true"></Exec>
+
+    <Error Condition="!Exists('$(AndroidBuildPath)')" Text="Android APK not found. Did something go wrong?"></Error>
```
bruno-garcia, line 101: Anyway a sanity check to make sure what we expect at the end is there.

Directory.Build.targets
```diff
@@ -108,8 +109,9 @@ Related: https://forum.unity.com/threads/6572-debugger-agent-unable-to-listen-on
     <Message Importance="High" Text="Building iOS Xcode project."></Message>
 
     <Exec Command="&quot;$(UnityExec)&quot; -quit -batchmode -nographics -logFile - -projectPath $(UnitySampleProjectPath) -executeMethod $(IOSBuildMethod) -buildPath $(IOSBuildPath)"
-          IgnoreStandardErrorWarningFormat="true"
-          IgnoreExitCode="true"></Exec>
+          IgnoreStandardErrorWarningFormat="true"></Exec>
+
+    <!-- <Error Condition="!Exists('$(IOSBuildPath)')" Text="iOS build not found. Did something go wrong?"></Error> -->
```
bruno-garcia, line 114: Gotta figure out where the app goes so we can fix this one

PR #272, opened 2021-07-29
Compile Initialization with the game

bitsandfoxes:
In preparation for the coming native support.
The aim is to have the Initialization compile with the game to have access to preprocessor directives.


src/Sentry.Unity/UnityLogger.cs
```diff
@@ -5,12 +5,12 @@
 
 namespace Sentry.Unity
 {
-    internal interface IUnityLoggerInterceptor
+    public interface IUnityLoggerInterceptor
```
bruno-garcia, line 8: This could be internal and the constructor that takes it too.

Lets add a single public constructor to `UnityLogger` that takes only `SentryOptions` instead of having the one with optional parameter:

```suggestion
    internal interface IUnityLoggerInterceptor
```

bruno-garcia (approved):
Sweet! Surprisingly simple change.
One nit but other than that, LGTM

bitsandfoxes:
> Sweet! Surprisingly simple change.
> One nit but other than that, LGTM

Should we add a changelog?

PR #281, opened 2021-08-20
fix: Passing Sentry diagnostic level to iOS native layer

bitsandfoxes:
We now pass the level as string to the native options.


bruno-garcia:
worth a changelog?

bruno-garcia (approved):
Nice! But could have a changelog item?

PR #288, opened 2021-08-23
fix: Moved NativeExample.c to Scripts/

bitsandfoxes:
The `NativeExample.c` does not have to be in `Plugins/` so to include it with our sample project I moved it to `Scripts/`.


bruno-garcia (approved):
Assuming because of the new path it gets copied out to the UPM package, LGTM

PR #296, opened 2021-08-26
feat: iOS native bridge for scope sync

bitsandfoxes:
Sync between .NET and iOS native via the scope observer interface and a native bridge.
Things that get synched:

- [x] Breadcrumbs
- [x] Extras
- [x] Tags
- [x] User

Remaining open: additional data for user (tracked by https://github.com/getsentry/sentry-unity/issues/314)


samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
```diff
@@ -0,0 +1,121 @@
...
+
+        if (type != NULL) {
+            breadcrumb.type = [NSString stringWithCString:type encoding:NSUTF8StringEncoding];
+        }
+        
+        if (category != NULL) {
+            breadcrumb.category = [NSString stringWithCString:category encoding:NSUTF8StringEncoding];
+        }
+        
+        if (level != NULL) {
```
philipphofmann, line 32: `m`: As level is an `int` it should never be `NULL`. Maybe it's worth checking if the passed `int` can be mapped to `SentryLevel`.

samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
```diff
@@ -0,0 +1,121 @@
...
+
+void SentryNativeBridgeAddBreadcrumb(const char* timestamp, const char* message, const char* type, const char* category, int* level) {
+    NSLog(@"Native Bridge: Adding breadcrumb.");
+    
+    if (timestamp == NULL && message == NULL && type == NULL && category == NULL && level == NULL) {
+        NSLog(@"Native Bridge: Breadcrumb empty. Dropping it.");
+        return;
+    }
+    
+    [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
```
philipphofmann, line 11: `m`: When you add `NS_ASSUME_NONNULL_BEGIN` below the header imports and `NS_ASSUME_NONNULL_END`, you don't need the `_Nonnull` specifiers as `_Nonnull` is default.

samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
```diff
@@ -0,0 +1,121 @@
...
+    NSLog(@"Native Bridge: Unsetting tag.");
+    [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
+        [scope removeTagForKey:[NSString stringWithUTF8String:key]];
+    }];
+}
+
+void SentryNativeBridgeSetUser(const char* email, const char* userId, const char* ipAddress, const char* username) {
+    NSLog(@"Native Bridge: Setting User.");
+    
+    if (email == NULL && userId == NULL && ipAddress == NULL && username == NULL) {
```
philipphofmann, line 86: `m`: Do we really need all of them? Or is it also just fine to use one of them? I guess it should be `||` instead of `&&`?

samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
```diff
@@ -0,0 +1,121 @@
...
+}
+
+void SentryNativeBridgeUnsetTag(const char* key) {
+    NSLog(@"Native Bridge: Unsetting tag.");
+    [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
+        [scope removeTagForKey:[NSString stringWithUTF8String:key]];
+    }];
+}
+
+void SentryNativeBridgeSetUser(const char* email, const char* userId, const char* ipAddress, const char* username) {
```
philipphofmann, line 83: `m`: Do you also have data, as in Cocoa? https://github.com/getsentry/sentry-cocoa/blob/7fb40d6e9dfb74fb444621b4475b866daabc9820/Sources/Sentry/Public/SentryUser.h#L32

philipphofmann (commented):
A quick first pass.

samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
```diff
@@ -0,0 +1,121 @@
+#import <Sentry/Sentry.h>
+
+void SentryNativeBridgeAddBreadcrumb(const char* timestamp, const char* message, const char* type, const char* category, int* level) {
```
bruno-garcia, line 3: This file should move into package-dev/Plugins/iOS so it's inferred to be iOS only and gets bundled with the packake

src/Sentry.Unity/UnityNativeScopeObserver.cs
```diff
@@ -0,0 +1,76 @@
+using System.Runtime.InteropServices;
+using Sentry.Extensibility;
+
+namespace Sentry.Unity
+{
+    public class UnityNativeScopeObserver : IScopeObserver
```
bruno-garcia, line 6: This will need to stay out of `Sentry.Unity` because the linker expects a native method declared in this file to exist and it won't in some cases (i.e Android, webGL etc)
We'll either need `Sentry.Unity.iOS` or this file goes as a single script

.github/workflows/ci.yml
```diff
@@ -43,28 +45,45 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2.3.3
+        with:
+          submodules: recursive
 
-      # Unity is about 3gb in size and takes 20 minutes to download and install.
+      # Unity is about 4gb in size and takes 20 minutes to download and install.
```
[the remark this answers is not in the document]
bruno-garcia: Technically it's `GiB` if we want to be correct here

samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
bitsandfoxes: Agreed, it belongs somewhere else. 
If we put it in `Plugins/iOS` tho it ends up in `Libraries/io.sentry.unity/Plugins/iOS/` right next to Sentry.framework. Does it belong there?

.github/workflows/ci.yml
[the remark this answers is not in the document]
bitsandfoxes: 🤓

package-dev/Plugins/iOS/SentryNativeBridge.m
```diff
@@ -0,0 +1,129 @@
+#import <Sentry/Sentry.h>
+#import <Sentry/SentryLog.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+void SentryNativeBridgeAddBreadcrumb(const char* timestamp, const char* message, const char* type, const char* category, int* level) {
```
bruno-garcia, line 6: Can we pass the value instead?

```suggestion
void SentryNativeBridgeAddBreadcrumb(const char* timestamp, const char* message, const char* type, const char* category, int level) {
```

samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
bitsandfoxes: It's just a safeguard for now as to not set a completely empty user.

samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
bitsandfoxes: We do have an `<string, string>` dictionary for additional data that we intend to sync down at a later point.
Gets tracked via: https://github.com/getsentry/sentry-unity/issues/314

samples/unity-of-bugs/Assets/Scripts/SentryNativeBridge.m
```diff
@@ -0,0 +1,121 @@
...
+    if (key == NULL) {
+        NSLog(@"Native Bridge: Extra key empty. Dropping it.");
+        return;
+    }
+    
+    [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
+        if (value != NULL) {
+            [scope setExtraValue:[NSString stringWithUTF8String:value] forKey:[NSString stringWithUTF8String:key]];
+        }
+        else {
```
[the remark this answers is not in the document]
bitsandfoxes: Very valid point. Tracking via https://github.com/getsentry/sentry-unity/issues/313

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -1,6 +1,10 @@
 using UnityEngine;
 using UnityEngine.Scripting;
 
+#if UNITY_IOS && !UNITY_EDITOR
```
bruno-garcia, line 4: No idea if this works:
```suggestion
#define ENABLE_ISO_BRIDGE = UNITY_IOS && !UNITY_EDITOR
#if ENABLE_ISO_BRIDGE
```

src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
```diff
@@ -24,7 +24,7 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
                 return;
             }
 
-            if (!options!.IOSNativeSupportEnabled)
+            if (!options!.IosNativeSupportEnabled)
```
bruno-garcia, line 27: Event though "we know it's checked in Validate" it's hard to tell from reading here.
So not against having the `null` check in the extension method but it looks a bit magic or possibly looks like a bug from the outside. So better have a `null` check before the use of `options` here

src/Sentry.Unity.iOS/IosNativeScopeObserver.cs
```diff
@@ -0,0 +1,108 @@
...
+                BreadcrumbLevel.Critical => 5,
+                _ => 0
+            };
+
+        public void SetExtra(string key, object? value)
+        {
+            string? extraValue = null;
+            if (value is not null)
+            {
+                extraValue = SerializeExtraValue(value);
```
bruno-garcia, line 62:
```suggestion
                extraValue = SerializeExtraValue(value);
                if (extraValue is null)
                {
                	// Failed to serialize, so lets just give up
                	return;
              	}
```

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -45,7 +45,7 @@ internal static string GetConfigPath(string? notDefaultConfigName = null)
         [field: SerializeField] internal int ShutdownTimeout { get; set; }
         [field: SerializeField] internal int MaxQueueItems { get; set; }
 
-        [field: SerializeField] internal bool IOSNativeSupportEnabled { get; set; }
+        [field: SerializeField] internal bool IosNativeSupportEnabled { get; set; }
```
bruno-garcia, line 48: breaking change?

bruno-garcia: Maybe add a new one `bool?` and make this one `[Obsolete]` and read from the old one so we can avoid losing the value when the users upgrade.

test/Sentry.Unity.iOS.Tests/IosNativeScopeObserverTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        private Fixture _fixture = new();
+
+        [SetUp]
+        public void SetUp() => _fixture = new Fixture();
+
+        [Test]
+        public void GetTimestamp_ReturnStringConformsToISO8601()
+        {
+            var timestamp = SystemClock.Clock.GetUtcNow();
```
bruno-garcia, line 29: SystemClock is used when we need to control which time is used at different places for testing.
Here you just need a timestamp, better to stick to the framework's API:

```suggestion
            var timestamp = DateTimeOffset.UtcNow;
```

test/Sentry.Unity.iOS.Tests/IosNativeScopeObserverTests.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        [Test]
+        public void SerializeExtraValue_ValueNotSerializable_ReturnsNull()
+        {
+            var sut = _fixture.GetSut();
+
+            var actualValue = sut.SerializeExtraValue(new SerializationTestClass());
+
+            Assert.AreEqual(null, actualValue);
```
bruno-garcia, line 68:
```suggestion
            Assert.IsNull(actualValue);
```

bruno-garcia (commented):
Just small items, looks like we can merge this soon

philipphofmann (commented):
Just to confirm the Objective-C code LGTM.

Issue #265, opened 2021-07-20
Errors on new install


michaeleconomy:
### Environment

How do you use Sentry?
to catch and report exceptions

Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)
saas

Which SDK and version?
e.g: JavaScript 5.11.1, .NET 1.2.0

unity 2020 lts

### Steps to Reproduce

1 follow install instructions

### Expected Result

no errors

### Actual Result
errors:

Assembly for Assembly Definition File 'Packages/io.sentry.unity/Editor/io.sentry.unity.editor.asmdef' will not be compiled, because it has no scripts associated with it.
UnityEditor.Scripting.ScriptCompilation.EditorCompilationInterface:TickCompilationPipeline (UnityEditor.Scripting.ScriptCompilation.EditorScriptCompilationOptions,UnityEditor.BuildTargetGroup,UnityEditor.BuildTarget,string[])

Assembly for Assembly Definition File 'Packages/io.sentry.unity/Runtime/io.sentry.unity.runtime.asmdef' will not be compiled, because it has no scripts associated with it.
UnityEditor.Scripting.ScriptCompilation.EditorCompilationInterface:TickCompilationPipeline (UnityEditor.Scripting.ScriptCompilation.EditorScriptCompilationOptions,UnityEditor.BuildTargetGroup,UnityEditor.BuildTarget,string[])

(this next one appears 5 times)
File couldn't be read
UnityEditor.AssetDatabase:Refresh ()
Sentry.Unity.Editor.SentryWindow:CopyLinkXmlToPlugins ()
Sentry.Unity.Editor.SentryWindow:OnEnable ()
UnityEditor.EditorWindow:GetWindow (System.Type)
Sentry.Unity.Editor.SentryWindow:OpenSentryWindow ()

An infinite import loop has been detected. The following Assets were imported multiple times, but no changes to them have been detected. Please check if any custom code is trying to import them:
Assets/Plugins/Sentry/link.xml

UnityEditor.AssetDatabase:Refresh ()
Sentry.Unity.Editor.SentryWindow:CopyLinkXmlToPlugins ()
Sentry.Unity.Editor.SentryWindow:OnEnable ()
UnityEditor.EditorWindow:GetWindow (System.Type)
Sentry.Unity.Editor.SentryWindow:OpenSentryWindow ()

michaeleconomy:
Sentry does appear to work.

bitsandfoxes:
[...]

PR #315, opened 2021-09-16
fix: Importing link.xml instead of refreshing AssetDatabase

bitsandfoxes:
Fixes: https://github.com/getsentry/sentry-unity/issues/265
Instead of creating the file in OnEnable and (possibly) triggering an entire AssetDatabase refresh that ends up in an endless loop we just import that one XML.

Fixed the weird formatting that happened a couple of PRs earlier as well, sry for the noise.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Import link.xml infinity loop fix ([#315](https://github.com/getsentry/sentry-unity/pull/315))
```
bruno-garcia, line 5:
```suggestion
## Unreleased

### Fixes

- Import link.xml caused an infinite loop ([#315](https://github.com/getsentry/sentry-unity/pull/315))
```

bruno-garcia (approved):

Issue #292, opened 2021-08-25
Remove console warning `Assembly Definition .. io.sentry.unity.editor.asmdef' will not be compiled ..`


bruno-garcia:
Since we don't add any scripts (we add a DLL) to Sentry.Unity.Editor, Unity warns there are no scripts to be compiled:

```
Assembly for Assembly Definition File 'Packages/io.sentry.unity/Editor/io.sentry.unity.editor.asmdef' will not be compiled, because it has no scripts associated with it.
```

Ideally we'd get rid of this warning. If that's not possible, we should add to the troubleshooting guide an explanation why this happens and that this isn't a real issue.

PR #316, opened 2021-09-17
fix: Removed unused .asmdefs

bitsandfoxes:
> Asmdefs are assets that define how scripts get compiled into DLLs

Except for `SentryInitialization.cs` we don't have any scripts. So the one for Runtime stays, the rest goes.

Resolves https://github.com/getsentry/sentry-unity/issues/292


bruno-garcia (approved):
brilliant!

bruno-garcia:
Oh, CI was green because it didn't run the build. And it didn't because the file changed doesn't trigger a build.

PR #322, opened 2021-09-18
run smoke test in test script

bruno-garcia:
#skip-changelog


bruno-garcia:
Sort of testing out a build with all caches in `main` being hot. It took 27 minutes with all cache hits.

Issue #309, opened 2021-09-12
[Windows] Unable to resolve reference 'UnityEditor.iOS.Extensions.Xcode'


ArnauKokoro:
### Environment

How do you use Sentry?
Sentry SaaS

Which SDK and version?
Windows Editor
Unity 0.5.1

### Steps to Reproduce

1. Play on Editor

### Expected Result

No error occurred

### Actual Result

```
Assembly 'Packages/io.sentry.unity/Editor/iOS/Sentry.Unity.Editor.iOS.dll' will not be loaded due to errors:
Unable to resolve reference 'UnityEditor.iOS.Extensions.Xcode'. Is the assembly missing or incompatible with the current platform?
```

This error occurs on Unity Editor Windows and Android platform selected. On Unity Editor macOS did not occur.

bruno-garcia:
Seems like it'll depend on the iOS build tools regardless of what platform you've currently selected.
As a work around, even though it doesn't make sense on Windows, is it possible to install the iOS build tools?

Does it result in a console error only? Does this block you from building an Android player or playing in the editor?

ArnauKokoro:
Hi @bruno-garcia unfortunately we cannot install iOS build tools on Windows.

Yes, it results as a console error and makes a crash so it stops the playing editor.

Thanks!!

bruno-garcia:
Thanks for the info. We're trying to find a solution for this

bitsandfoxes:
Hey @ArnauKokoro, thanks again for bringing this to our attention. We've excluded everything iOS related from the Win platform and the next release will contain a fix for this issue!

PR #325, opened 2021-09-20
fix: Resolving iOS reference on Windows

bitsandfoxes:
Fixes: https://github.com/getsentry/sentry-unity/issues/309
The *dll.meta files allow us fine-grained control of which DLLs do what on which platform.

Screenshot 2021-09-20 at 18 42 13

So all of our iOS related DLL are exclusive to OSX.
I added all of our dll metas in this PR because this issue keeps coming up and I don't want to manually exclude each and every file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,19 +26,14 @@ package-dev/**/*.xml
 package-dev/**/TestSentryOptions.json
 package-dev/Tests/Editor/TestFiles/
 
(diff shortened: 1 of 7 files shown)


bruno-garcia (approved):
Nice. Shouldn't some new `.meta` show up on git for you to check in?

bruno-garcia:
Nevermind I see all of them now

Issue #304, opened 2021-09-06
Library not loaded: @rpath/Sentry.framework/Sentry, Reason: image not found


ketourneau:
### Environment

How do you use Sentry?
Sentry SaaS

Which SDK and version?
sentry-unity 0.5.1

### Steps to Reproduce

1. Create new empty project with Unity 2020.3.17f1 LTS (target ios platform)
2. Install sentry-unity 0.5.1 with upm
3. Build for ios
4. Open build result with XCode
5. Deploy on ipad

### Expected Result

App start and can use sentry-unity library.

### Actual Result

App crash and got "Library not loaded: @rpath/Sentry.framework/Sentry ... Reason: image not found".

`2021-09-06 16:13:31.972514+0200 DebugSentry[869:160732] Error loading /var/containers/Bundle/Application/49BD94D0-0A4E-41DF-9BBA-4865582FCE56/DebugSentry.app/Frameworks/UnityFramework.framework/UnityFramework:  dlopen(/var/containers/Bundle/Application/49BD94D0-0A4E-41DF-9BBA-4865582FCE56/DebugSentry.app/Frameworks/UnityFramework.framework/UnityFramework, 265): Library not loaded: @rpath/Sentry.framework/Sentry
  Referenced from: /private/var/containers/Bundle/Application/49BD94D0-0A4E-41DF-9BBA-4865582FCE56/DebugSentry.app/Frameworks/UnityFramework.framework/UnityFramework
  Reason: image not found`

bruno-garcia:
Is this building for a real device or simulator? I ask because simulator isn't supported yet (#297)

ketourneau:
I build and run it in real device

bruno-garcia:
Would you please be able to provide us with a small repro? 

I've added the SDK to [this game](https://github.com/getsentry/sentry-defenses), same version 0.5.1 with Unity 2020.3.15f2 LTS and it's working properly.

ketourneau:
@bruno-garcia It works in your game sample but not when I create new unity project.

[I add a repo project](https://github.com/ketourneau/DebugSentry/tree/master).

ketourneau:
I found the problem, the app crash if dsn is empty.
We set sentry dsn at runtime from script.

bitsandfoxes:
Thanks for the repo! I'll take a look. 
And double thanks for digging further and the DSN hint. That does sound like a bug.

PR #330, opened 2021-09-22
fix: Properly embedded iOS native support

bitsandfoxes:
Fixes: https://github.com/getsentry/sentry-unity/issues/304

Unity copies the `.framework` and our `bridge.m` in any case. But with invalid options, we bail too soon and the Xcode project gets left in a broken state.

### DIFF
--- a/src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
+++ b/src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
@@ -30,7 +30,7 @@ namespace Sentry.Unity.Editor.iOS
 
                 if (!options.IosNativeSupportEnabled)
                 {
-                    options.DiagnosticLogger?.LogDebug("iOS Native support disabled. Native support disabled.");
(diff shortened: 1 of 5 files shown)


src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
```diff
@@ -16,29 +16,30 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
+
+                if (options?.Validate() != true)
+                {
+                    new UnityLogger(new SentryOptions()).LogWarning("Failed to validate Sentry Options. Native support disabled.");
+                    return;
+                }
+
+                if (!options.IosNativeSupportEnabled)
+                {
+                    options.DiagnosticLogger?.LogDebug("iOS Native support disabled. Native support disabled.");
```
bruno-garcia, line 33:
```suggestion
                    options.DiagnosticLogger?.LogDebug("iOS Native support disabled through the options.");
```

bruno-garcia (approved):

PR #331, opened 2021-09-23
add android native libraries to artifact

bruno-garcia:
#skip-changelog

The UPM package contains `sentry-native` already bundled in the `-ndk.aar` package. This PR will add these directly in the GHA artifact which craft downloads. That allows the symbol collector target to find them.

See also: https://github.com/getsentry/sentry-java/blob/7726602f22cc29db736cc687606549d1663ba4b6/.github/workflows/build.yml#L60

Currently, only the UPM package exists. And within it:
image

The Android SDK on the other hand, contains not only the .aar that gets uploaded to maven central, but the individual `.so` that symbol collector is looking for:

image


.github/workflows/ci.yml
```diff
@@ -204,7 +204,10 @@ jobs:
         uses: actions/upload-artifact@v1
         with:
           name: ${{ github.sha }}
-          path: package-release.zip
+          # Adding the native libraries so the symbol collector craft target can find/upload them
+          path: | 
+            package-release.zip
+            ./src/sentry-java/sentry-android-ndk/build/intermediates/merged_native_libs/release/out/lib/*
```
bitsandfoxes, line 210: We're caching the build output in `/Plugins`. So the submodule build output does not exist.

bruno-garcia:
Added to artifact:

bruno-garcia:
Also checking files are not stripped of debug info:
image

/cc @marandaneto

.github/workflows/ci.yml
bruno-garcia: Added path to be cached too.

PR #338, opened 2021-09-24
fix: android crumb level

bruno-garcia:
Technically these JNI classes should be disposed on each use. Docs says subsequent calls to the same method are faster, but doesn't specify if that's on the same instance.

Since these would be used for the lifetime of the app, I believe we can allocate them statically (perhaps Lazily instead, on the first use) and go without disposing them altogether. Would be nice to get some validation of these thoughts before we merge this

image

image


bruno-garcia:
Replaces by: #339

bitsandfoxes:
I agree with the sentiment of holding those statically. Especially since their lifetime equals the apps.

PR #339, opened 2021-09-24
android: re-installation of native backend

bruno-garcia:
Android reinitialization of the `sentry-native` backend after C# init to compensate Unity taking the signal handler and not calling back into Sentry's one.

Additionally, other changes happened in the process that were hard to split into different PRs:

* Adding breadcrumb level for Android (replacing #338)
* Aligned behavior (mainly logging) between iOS and Android C# bridge
* Moved any logic/complex code away from the SentryInitialization class since we don't get that to compile with the solution. There's now a single method to configure the native layer through C# and less code to document on the 'manual initialization'
* Enabled NDK scope propagation by default (we can possibly add an opt-out in the future if this becomes an issue in terms of overhead)
* Unit tests where possible
* `SendMessage` button changed to a `SentrySdk.CaptureMessage` since Debug.Log only tracks breadcrumbs
* Changed the `CaptureExceptionToString` which was used for testing the parsing code originally to a `SetUser` so we can have a sample of scope propagation: User set in C# should end up in errors in Cocoa, Java, C:

Example [event in Sentry](https://sentry.io/organizations/sentry-sdks/issues/2668868634/?project=5439417&query=is%3Aunresolved):

image


bitsandfoxes (approved):
Thanks for the additional cleanup!

PR #342, opened 2021-09-27
c, obj-c and cpp samples

bruno-garcia:
In both new samples I wasn't able to get a symbolicated stack trace properly. Even though I ran `sentry-cli` on the directory I configured Unity to create the xcode project and also on the original Unity project (so it can get anything under Temp). In both situations I got:

image

It could symbolicate frames from the SDK itself (since the `.framework` is copied to the xcode project and `sentry-cli` found it there) but I didn't find the `UnityFramework.framework`.

Sounds like it's something we need to find out and properly document so users don't get confused.

In any case, here we have 2 more samples: C++ Exceptions and C++ segfault

The Objective-C example requires an additional configuration which requires documenting: https://github.com/getsentry/sentry-unity/issues/343

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Android Native Support ([#307](https://github.com/getsentry/sentry-unity/pull/307))
 - iOS native bridge for scope sync ([#296](https://github.com/getsentry/sentry-unity/pull/296))
 - Android native bridge for scope sync ([#308](https://github.com/getsentry/sentry-unity/pull/308))
+- Sample: C++ Exception and Segfault ([#342](https://github.com/getsentry/sentry-unity/pull/342))
 
 ### Fixes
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Android Native Support ([#307](https://github.com/getsentry/sentry-unity/pull/307))
 - iOS native bridge for scope sync ([#296](https://github.com/getsentry/sentry-unity/pull/296))
 - Android native bridge for scope sync ([#308](https://github.com/getsentry/sentry-unity/pull/308))
+- Sample: C++ Exception and Segfault ([#342](https://github.com/getsentry/sentry-unity/pull/342))
```
bruno-garcia, line 10:
```suggestion
- Sample: Throw exceptions in C++ and Objective-C. C++ segfault ([#342](https://github.com/getsentry/sentry-unity/pull/342))
```

samples/unity-of-bugs/Assets/Scenes/1_BugfarmScene.unity
```diff
@@ -170,8 +170,8 @@ MonoBehaviour:
   m_Enabled: 1
   m_EditorHideFlags: 0
   m_Script: {fileID: 11500000, guid: 4e29b1a8efbd4b44bb3f3716e73f07ff, type: 3}
-  m_Name:
-  m_EditorClassIdentifier:
+  m_Name:
```
bruno-garcia, line 173: Not sure what caused all these white spaces. Tricky to revert the file while keeping the new buttons only.

bitsandfoxes (approved):

PR #347, opened 2021-09-28
feat: android mark sessions as crashed

bruno-garcia:


marandaneto:
what should we do about:

> Logs for the case where it doesn't work as expected:

bruno-garcia:
> what should we do about:
> 
> 
> 
> > Logs for the case where it doesn't work as expected:
> 
> 
> 
> 

Nothing. Thst was happening before I changed the onQuitting.

bitsandfoxes (approved):

marandaneto (approved):

bruno-garcia:
We'll take other changes in and do a final submodule update before we release (to take the tags)

PR #380, opened 2021-10-21
Windows native support

bruno-garcia:
Bringing sentry-native as a submodule which we can use to build native Linux and Windows.

Following install instructions from: https://docs.sentry.io/platforms/native/
Couldn't use the configuration listed there. Had to use `Release`.

~Needed to add Directory.Build.props to `src/sentry-native` since CMake creates a `.sln` which ends up finding our Unity `.prop` files and fails to build due to not finding Unity installs. A possible solution for this that won't require changing `sentry-native` itself is to add the submodules to a directory (such as `src/modules/sentry-native`) and add the empty .prop and .targets in the modules directory.~

Using crashpad means we need to distribute the game with the handler sitting next to it (or somehow provide the path to it when initializing the SDK). ~Alternatively we can use a backend such as the Android SDK and use the .NET layer to pick up those envelopes and send it to Sentry.~

Also to figure out: We can't build and ship the package with building in a single CI agent now. We'll must build on Android + macOS and then assemble a package using bits taken from both builds.

~Seems like assemblies such as `Sentry.Unity.Android` should be renamed to simply `Sentry.Android` since all they do is wrap our Android SDK without any Unity specific logic. Technically this code could be used by .NET Mobile in the future and made part of the core .NET SDK. Naming `Sentry.Unity.Native` for now but same here, should be `Sentry.Native`.~


vaind:
* [ ] TODO: from the native docs:
> Warning
> Calling sentry_close() before exiting the application is critical. It ensures that events can be sent to Sentry before execution stops. Otherwise, event data may be lost.

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -81,6 +81,11 @@ public new CompressionLevelWithAuto RequestBodyCompressionLevel
         /// </summary>
         public bool AndroidNativeSupportEnabled { get; set; } = true;
 
+        /// <summary>
+        /// Whether the SDK should add native support for Windows
+        /// </summary>
+        public bool WindowsNativeSupportEnabled { get; set; } = true;
```
bitsandfoxes, line 87: There is a test at `Sentry.Unity.Tests.ScriptableSentryUnityOptions` where we check the content of the actual `.asset` file so we don't accidentally drop any properties.

bitsandfoxes (commented):
Looks really good to me!

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -3,6 +3,8 @@
 #define SENTRY_NATIVE_IOS
 #elif UNITY_ANDROID
 #define SENTRY_NATIVE_ANDROID
+#elif UNITY_STANDALONE_WIN && ENABLE_IL2CPP
```
bitsandfoxes, line 6:
```suggestion
#elif UNITY_STANDALONE_WIN
```
Is IL2CPP a requirement for native support? What about native plugins on mono?

.github/workflows/ci.yml
```diff
@@ -170,12 +170,18 @@ jobs:
       - name: Build Android SDK
         env:
           UNITY_VERSION: ${{ matrix.unity-version }}
-        run: dotnet msbuild /t:BuildAndroidSDK /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
+        run: dotnet msbuild /v:d /t:BuildAndroidSDK /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
```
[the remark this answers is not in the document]
bruno-garcia: I'm afraid this will bloat the output log quite a bit, can we do this only when we need to debug the build in CI?

package-dev/Runtime/SentryInitialization.cs
bruno-garcia: @bitsandfoxes we'd be taking the signal handler from Mono so I believe that's not what we want

bruno-garcia: Also makes me realize we need to be clear about minidump support requiring IL2CPP

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -0,0 +1,76 @@
...
+        [PostProcessBuild(1)]
+        public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
+        {
+            if (target is not (BuildTarget.StandaloneWindows or BuildTarget.StandaloneWindows64))
+            {
+                return;
+            }
+
+            var options = ScriptableSentryUnityOptions.LoadSentryUnityOptions(BuildPipeline.isBuildingPlayer);
+            var logger = options?.DiagnosticLogger ?? new UnityLogger(new SentryUnityOptions());
```
bruno-garcia, line 21: Is `options` of type SentryUnityOptions? Can we instead do:

```suggestion
            var logger = options?.DiagnosticLogger ?? new UnityLogger(options);
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            try
+            {
+                if (options?.Validate() != true)
```
bruno-garcia, line 25: Since we're using pattern matching above, to keep consistent (and since it's preferred over an operator which can be overriden):

```suggestion
                if (options?.Validate() is not true)
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -0,0 +1,76 @@
...
+
+                if (!options.WindowsNativeSupportEnabled)
+                {
+                    logger.LogDebug("Windows Native support disabled through the options.");
+                    return;
+                }
+
+                var crashpadPath = Path.GetFullPath(Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins",
+                    "Windows", "Sentry", "crashpad_handler.exe"));
+                var targetPath = Path.Combine(Path.GetDirectoryName(pathToProject)!, Path.GetFileName(crashpadPath));
```
bruno-garcia, line 39: `Path.GetDirectoryName(pathToProject)!` when can this method return null?
Should we assign to a variable and check for null and log/bail instead of `!` in here?

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <summary>
+    /// Access to the Sentry native support of the sentry-native SDK.
+    /// </summary>
+    public static class SentryNative
+    {
+        /// <summary>
+        /// Configures the native SDK.
+        /// </summary>
+        /// <param name="options">The Sentry Unity options to use.</param>
+        public static void Configure(SentryUnityOptions options)
```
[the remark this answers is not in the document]
bruno-garcia: What if we just don't enable native support when running on Mono?

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            if (options.WindowsNativeSupportEnabled)
+            {
+                SentryNativeBridge.Init(options.Dsn!, options.Release!);
```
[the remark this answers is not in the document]
bruno-garcia: if `!` is set here I assume the method doesn't accept nullable. So either we need to check for null before calling, or make the `Init` method take null and deal with it later.
If we can't have `Init` do its thing with null arguments, we should just check for those values here before we call and log in case we're not calling due to missing options.

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -0,0 +1,45 @@
...
+                //     {
+                //         options.DiagnosticLogger?
+                //             .LogDebug("Native Android SDK reported: 'crashedLastRun': '{0}'", crashedLastRun);
+                //     }
+
+                //     return crashedLastRun.Value;
+                // };
+                // At this point Unity has taken the signal handler and will not invoke the original handler (Sentry)
+                // So we register our backend once more to make sure user-defined data is available in the crash report.
+                SentryNativeBridge.ReinstallBackend();
```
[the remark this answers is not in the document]
bruno-garcia: I wonder why do we even init the native SDK in this case, since it won't capture crashes without reinstalling the backend. Perhaps best is to avoid calling native init altogether

bruno-garcia (commented):
Looking great!

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: options may be null here. This is a line I've copy-pasted from another place (same pattern is used many times)

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: ok

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: actually, it cannot be null. fixed

src/Sentry.Unity.Native/SentryNative.cs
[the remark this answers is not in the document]
vaind: Note: DSN is not null because `options.Validate()` must have returned true in SentryInitialization.cs for this to be called at all. I've added a comment.

package-dev/Runtime/SentryInitialization.cs
vaind: I've also added a warning to the PostBuild to warn the user when he has the Native SDK support enabled in settings but then uses non-IL2CPP backend.

src/Sentry.Unity.Native/SentryNative.cs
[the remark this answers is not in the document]
vaind: that's what currently happens. I don't see an issue with that - can be changed later if needed.

src/Sentry.Unity/SentryUnityOptions.cs
vaind: added there.

bitsandfoxes (approved):

bruno-garcia:
Thanks for doing this @vaind

PR #394, opened 2021-10-30
build on linux

bruno-garcia:


bruno-garcia:
The only problem here was that the license JSON file we set as env var was failing to be read on Linux. I tried different escapings without success.

Since we don't have enough licenses to add this and have 2 PRs running in paralllel. This will be put on hold

PR #397, opened 2021-11-02
messing with assembly aliasing

SimonCropp:
#skip-changelog


bitsandfoxes:
I used the modified pack script and imported the package-release into a fresh Unity project but I'm getting stuck on 
`SentryInitialization.cs(28,25): error CS0012: The type 'SentryOptions' is defined in an assembly that is not referenced. You must add a reference to assembly 'Sentry, Version=3.10.0.0, Culture=neutral, PublicKeyToken=fba2ec45388e2af0'.`

scripts/pack.ps1
```diff
@@ -1,6 +1,9 @@
 New-Item "package-release" -ItemType Directory
 
 # Copy `package-dev` stuff
+dotnet tool install --global Alias --version 0.1.0
```
bruno-garcia, line 4: We'll need to have this run through msbuild so that when we run `pack.ps1`, assemblies are already aliased.
The reason is that when we load up Unity with the sample project, we'd things already aliased, so unit and integration tests will be loaded aliased. And when we build players for testing or smoke testing, they'll be validating the bits we'll bundle through this script

SimonCropp: @bruno-garcia in the after build of the test project?

bitsandfoxes: `install` skips even if the provided differs from the already installed one?

PR #402, opened 2021-11-03
feat: Android automated symbol upload

bitsandfoxes:
The idea is to upload symbols if the automated symbol-upload is enabled.

- [ ] Make sentry-cli work with the unity package
  - [ ] Add sentry-cli to package-dev
  - [ ] Add restoring/loading specified version via script to msbuild (i.e. https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/download-sentry-cli.sh)
  - [ ] Make sure properties are set (chmod +x)
- [ ] Store sentry-cli options somewhere 
  - [ ] Create on sentry window open with options & link XML?
  - [ ] Split Sentry Scriptable Options into SDK and Editor related options?
  - [x] Contain Auth Token, Org-Slug, Proj-Name, Upload toggle
- [x] Hook into the BuildPostProcessor
  - [x] Check if `Create symbols.zip` is enabled
  - [x] Check if `Development Build` is disabled
  - [x] Check for `BuiltName-ApplicationVersion-v1.symbols.zip`
  - [x] Use platform specific version of sentry-cli for uploading

### DIFF
--- a/src/Sentry.Unity.Editor/Android/AndroidBuildPostProcessor.cs
+++ b/src/Sentry.Unity.Editor/Android/AndroidBuildPostProcessor.cs
@@ -18,14 +18,33 @@ namespace Sentry.Unity.Editor.Android
         [MenuItem("Tools/SentryUploadSymbols")]
         public static void PostProcess() => UploadSymbols(pathToBuiltProject);
 
+        [MenuItem("Tools/Test")]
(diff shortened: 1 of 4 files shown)


bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-unity/pull/408

PR #412, opened 2021-11-11
Add Android Devices Test on mac-latest.

lucas-zimerman:
Add smoke tests with Android emulators.
Tests will run only if the main tests are completed on all devices to allow better usage of job parallelism.

NOTE: In addition to the test, the IL2CPP apk required to support the X86 ABI in order to run the tests.
Additionally, the tests only ran on macOS due to Virtualization limitations on Github being exclusively supported on the macOS host.

#skip-changelog.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    name: Unity ${{ matrix.unity-version }} - ${{ matrix.os }} - Android ${{  matrix.api-level }}
+    name: Unity ${{ matrix.unity-version }} - ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     env:
       DOTNET_CLI_TELEMETRY_OPTOUT: 1
@@ -73,15 +73,6 @@ jobs:
           path: samples/artifacts/builds
           # using hash of package.json to bust cache on release builds which modify it
           key: ${{ matrix.os }}-${{ matrix.unity-version }}-${{ matrix.unity-version-changeset }}-${{ hashFiles('package/package.json') }}
-
-      - name: AVD cache
-        uses: actions/cache@v2
-        id: avd-cache
-        with:
-          path: |
-            ~/.android/avd/*
-            ~/.android/adb*
-          key: avd-${{ matrix.api-level }}
 
       - name: Restore Unity Packages
         uses: actions/cache@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -188,6 +198,19 @@ jobs:
           UNITY_VERSION: ${{ matrix.unity-version }}
         run: dotnet msbuild /t:UnitySmokeTestStandalonePlayerIL2CPP
 
+      - name: Setup Android Emulators
+        if: matrix.os == 'macos-latest'
+        uses: reactivecircus/android-emulator-runner@v2
```
bruno-garcia, line 203: Not a verified action. Please pin this to a commit hash:

```suggestion
        uses: reactivecircus/android-emulator-runner@9216c92cd9935b2aea47751a50e686a9c884b2dc
```

bruno-garcia (commented):
Looking very promising 👍

scripts/smoke-test-droid.ps1
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    $stdout = adb -s $device logcat -d  | select-string SMOKE
+    if ($stdout -ne $null)
+    {
+        Write-Output "$stdout"
+    }
+    else
+    {
+        Start-Sleep -Seconds 2
```
bitsandfoxes, line 105: Why the sleep?

scripts/smoke-test-droid.ps1
```diff
@@ -0,0 +1,112 @@
...
+    $checkStarted = 'False'
+
+    Write-Output "Clearing logcat from $device."
+    adb -s $device logcat -c
+
+    Write-Output "Starting Test..."
+
+    adb -s $device shell am start -n $TestActivityName -e test smoke
+
+    for ($i = 30; $i -gt 0; $i--) {
```
bitsandfoxes, line 68: What about something like this:
```
$timeout = 30
while($timeout -gt 0){
    $smokeTestId = (adb -s $device shell ps)
    $smokeTestId = $smokeTestId | select-string $ActivityName

    If ($smokeTestId -eq $null)
    {
        break
    }

    Write-Output "Waiting Process on $device to complete, waiting $timeout seconds"
    Start-Sleep -Seconds 1
    $timeout--
}
```
With the `checkStarted` you skip ahead in the for without waiting so the timeout of 30 is not really 30s?

scripts/smoke-test-droid.ps1
```diff
@@ -0,0 +1,112 @@
...
+Write-Output "#                       SCRIPT                  #"
+Write-Output "#################################################"
+
+Set-Variable -Name "ApkPath" -Value "samples/artifacts/builds/Android"
+Set-Variable -Name "ApkFileName" -Value "IL2CPP_Player.apk"
+Set-Variable -Name "ActivityName" -Value "io.sentry.samples.unityofbugs"
+Set-Variable -Name "TestActivityName" -Value "io.sentry.samples.unityofbugs/com.unity3d.player.UnityPlayerActivity"
+
+# Filter device List
+$RawAdbDeviceList = adb devices
```
bitsandfoxes, line 14: I'm not familiar with adb devices, is there a way to install additional ones?

scripts/smoke-test-droid.ps1
lucas-zimerman: Some Android versions seem to only exit the adb callback for `am start` when the app is open, while others don't.
So checkStarted is there to inform if the app was started or not.

If the app was not initialized, smokeTestId is going to be set to null, and checkStarted will avoid going out of the loop since the app wasn't initialized.

Once it detects the app was initialized, it'll wait for smokeTestId  to be come null.

About smokeTestId, it'll get the list of active processes, and doing the `$smokeTestId | select-string $ActivityName`
 will return null if it doesn't find `io.sentry.samples.unityofbugs` otherwise, it'll return ugh, `io.sentry.samples.unityofbugs` 😅.

It's a silly way of checking if the process is running, there're better ways of doing that, but those methods may run on version X and not on Y where the PS one seems to work on any Android version.

scripts/smoke-test-droid.ps1
lucas-zimerman: `adb devices` gives the list of devices. if you have one device connected, it'll display only one item.
In the case of multiple devices, it'll display all active devices.
You need that info in case of having multiple active simulators because if you have more than one device connected, you need to specify which device you wish to execute the command.

scripts/smoke-test-droid.ps1
lucas-zimerman: Removed since it wasn't necessary for grabbing more info after the app was closed.

.github/workflows/ci.yml
```diff
@@ -225,3 +231,40 @@ jobs:
         with:
           name: Test results (editmode)
           path: artifacts/test/editmode
+
+  android-smoke-test:
+    needs: [build]
+    name: Unity ${{ matrix.unity-version }} - macos-latest - Android ${{ matrix.api-level }} smoke tests
```
bruno-garcia, line 237:
```suggestion
    name: Smoke Test - Android ${{ matrix.api-level }}  Unity ${{ matrix.unity-version }}
```

.github/workflows/ci.yml
```diff
@@ -225,3 +231,40 @@ jobs:
...
           path: artifacts/test/editmode
+
+  android-smoke-test:
+    needs: [build]
+    name: Unity ${{ matrix.unity-version }} - macos-latest - Android ${{ matrix.api-level }} smoke tests 
+    runs-on: macos-latest
+    strategy:
+      matrix:
+         api-level: [21, 26, 29]
+         # 2021.1.26f1 is not being generated.
```
bruno-garcia, line 242: This is odd. Is it broken on this version?

bruno-garcia, line 242: We should only skip this version on Windows due o the PathTooLong issue (which is a Unity issue). Since it builds on macOS, we should have an artifact here to test

.github/workflows/ci.yml
```diff
@@ -142,6 +142,12 @@ jobs:
         shell: pwsh
         run: |
           Write-Output "${{ matrix.unity-root }}${{ matrix.unity-version }}/${{ matrix.unity-path }}" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          
+      - name: Make symbolic link for Unity (Windows)
```
bruno-garcia, line 146: This was merged on to main, maybe pull and resolve conflicts?

bruno-garcia (commented):
Sounds like we can merge this. when resolving `main` merge?

PR #414, opened 2021-11-12
feat: Added sentry-cli to package

bitsandfoxes:
Adding sentry-cli for convenience for now. And because we'll need it for the upcoming automated symbols upload.


scripts/download-sentry-cli.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+cd $(dirname "$0")
+REPO=getsentry/sentry-cli
+VERSION=1.70.0
```
bruno-garcia, line 4: There's already 1.71.0: https://github.com/getsentry/sentry-cli/releases

```suggestion
VERSION=1.71.0
```

But we don't need to update right now, nothing on that release we need here.

bruno-garcia (approved):

PR #433, opened 2021-11-24
Fix Android 27 smoke test

lucas-zimerman:
Retry for 3 times if detected Unity timed out to initialize.

#skip-changelog.

### DIFF
--- a/scripts/smoke-test-droid.ps1
+++ b/scripts/smoke-test-droid.ps1
@@ -58,37 +58,54 @@ foreach ($device in $DeviceList)
     {
         Throw "Failed to Install APK: $stdout."
     }
-    $AppStarted = 'False'
 
-    Write-Output "Clearing logcat from $device."
-    adb -s $device logcat -c
-
-    Write-Output "Starting Test..."
-
-    adb -s $device shell am start -n $TestActivityName -e test smoke
-    #despite calling start, the app might not be started yet.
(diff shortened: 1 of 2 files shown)


scripts/smoke-test-droid.ps1
```diff
@@ -58,37 +58,54 @@ foreach ($device in $DeviceList)
...
             break
         }
-        ElseIf ($processIsRunning -ne $null -And $AppStarted -eq 'False')
+        Else
         {
-            # Some devices might take a while to start the test, so we wait for the activity to start before checking if it was closed.
-            $AppStarted = 'True'
+            Write-Warning "Test was flaky, retrying."
+            $FlakyRetry--
+            Start-Sleep -Seconds 3
```
bitsandfoxes, line 107: Can you really just restart it like that? Is it dead and stuck and hitting the timeout because of that? You can run the process twice? Am I missing the cleanup?

bitsandfoxes (approved):
Other than those open questions: LGTM!

scripts/smoke-test-droid.ps1
lucas-zimerman: The app is already closed when entering this if condition. So all it does is open the test app again and repeat the tests.
What's happening seems to be Unity is killing the app before even running the tests.

PR #445, opened 2021-12-03
move submodules outside of src

bruno-garcia:
One of the issues to compile `sentry-native` on Windows is that it CMake creates a `.sln` on Windows and that inherits our Directory.Build.props.

A solution is to not have that flow through upwards. See details here #380

#skip-changelog

### DIFF
--- a/src/Sentry.Unity/Sentry.Unity.csproj
+++ b/src/Sentry.Unity/Sentry.Unity.csproj
@@ -17,7 +17,7 @@
   </Target>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\modules\sentry-dotnet\src\Sentry\Sentry.csproj" />
+    <ProjectReference Include="../sentry-dotnet/src/Sentry/Sentry.csproj" Private="false" />
   </ItemGroup>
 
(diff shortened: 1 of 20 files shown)


SimonCropp (approved):

src/Sentry.Unity/Sentry.Unity.csproj
```diff
@@ -15,9 +15,9 @@
     </ItemGroup>
     <Delete Files="@(FilesToDelete)" ContinueOnError="true" />
   </Target>
-  
+
   <ItemGroup>
-    <ProjectReference Include="..\sentry-dotnet\src\Sentry\Sentry.csproj" />
+    <ProjectReference Include="..\..\modules\sentry-dotnet\src\Sentry\Sentry.csproj" />
```
bitsandfoxes: The slashes are killing me. It's funny how Rider just doesn't care and mixes them up.

bruno-garcia: msbuild that supports either way. It was necessary to go from Windows to x-plat

PR #450, opened 2021-12-06
fix: Symbol upload requires IL2CPP

bitsandfoxes:
With automatic symbols upload enabled and scripting backend not IL2CPP we currently break builds because we're looking for symbol files that don't exist.
That and we were missing a bunch of upload feature related tests.

### DIFF
--- a/src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
+++ b/src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
@@ -49,7 +49,10 @@ namespace Sentry.Unity.Editor.Android
         public void OnPostGenerateGradleAndroidProject(string basePath)
         {
             ModifyManifest(basePath);
-            SetupSymbolsUpload(basePath);
+
+            var unityProjectPath = Directory.GetParent(Application.dataPath).FullName;
+            var gradleProjectPath = Directory.GetParent(basePath).FullName;
+            SetupSymbolsUpload(unityProjectPath, gradleProjectPath);
         }
 
         internal void ModifyManifest(string basePath)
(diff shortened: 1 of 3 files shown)


src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
```diff
@@ -133,17 +147,22 @@ internal void SetupSymbolsUpload(string basePath)
...
-            var sentryCliOptions = SentryCliOptions.LoadCliOptions();
-            if (!sentryCliOptions.UploadSymbols)
+            if (_scriptingImplementation != ScriptingImplementation.IL2CPP)
+            {
+                logger.LogDebug("Automated symbols upload requires the IL2CPP scripting backend.");
+                return;
+            }
+
+            var sentryCliOptions = _getSentryCliOptions();
+            if (sentryCliOptions is null || !sentryCliOptions.UploadSymbols)
```
bruno-garcia, line 157: In the event `null` is returned, it'll be hard to tell from the log message. Ideally we'd have to `if`s here, one could be a `Warn` or `Err`? Why/When would we return `null`?

bruno-garcia (approved):
nit otherwise LGTM

src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
bitsandfoxes: There should not be a case where it's `null`. If we can't find it at the default location we create a new one and return that one. Which might be its own issue.
I'll separate the null-check to have it covered anyway as I intend to refactor parts of our scriptable object creation and the editor config window when we improve our onboarding experience.

PR #452, opened 2021-12-06
fix: CI uploading test results on failure

bitsandfoxes:
I might be misunderstanding the use-case for test results but right now: If the tests fail the upload gets skipped.

#skip-changelog


bruno-garcia (approved):

lucas-zimerman:
failure() will be skipped if the previous steps didn't fail.
if you intend to make the uploas if the build passes or when it fails you could use `always()` instead of `failure()`

bitsandfoxes:
> failure() will be skipped if the previous steps didn't fail. if you intend to make the uploads if the build passes or when it fails you could use `always()` instead of `failure()`

I think this is what I want? I don't particularly care about passing tests but about failing ones. `always` would be an option tho. I'll keep that in mind. Thanks!

PR #455, opened 2021-12-07
fix: CI ignores CHANGELOG

bitsandfoxes:
CI should ignore changes to the changelog.

#skip-changelog


bruno-garcia:
Odd that it runs CI even after adding the ignore

bitsandfoxes:
> Odd that it runs CI even after adding the ignore

Got me with that one! I didn't change the changelog file but CI itself! 😅

Issue #456, opened 2021-12-08
Error building iOS player when using Append instead of Replace


bruno-garcia:
Affected version 0.8.0.

```
Sentry: (Debug) 'Sentry.framework' has already copied to '/Users/bruno/sentry-defenses-build/1.0.3/ios/Frameworks/Sentry.framework' 
UnityEngine.Debug:Log (object)
Sentry.Unity.UnityLogger:Log (Sentry.SentryLevel,string,System.Exception,object[]) (at /Users/runner/work/sentry-unity/sentry-unity/src/Sentry.Unity/UnityLogger.cs:41)
Sentry.Extensibility.DiagnosticLoggerExtensions:LogIfEnabled<string> (Sentry.Extensibility.IDiagnosticLogger,Sentry.SentryLevel,System.Exception,string,string)
Sentry.Extensibility.DiagnosticLoggerExtensions:LogDebug<string> (Sentry.Extensibility.IDiagnosticLogger,string,string)
Sentry.Unity.Editor.iOS.BuildPostProcess:CopyFrameworkToBuildDirectory (string,Sentry.Extensibility.IDiagnosticLogger)
Sentry.Unity.Editor.iOS.BuildPostProcess:OnPostProcessBuild (UnityEditor.BuildTarget,string)
UnityEngine.GUIUtility:ProcessEvent (int,intptr,bool&) (at /Users/bokken/buildslave/unity/build/Modules/IMGUI/GUIUtility.cs:189)

Sentry: (Info) 'main.mm' already contains Sentry. 
UnityEngine.Debug:Log (object)
Sentry.Unity.UnityLogger:Log (Sentry.SentryLevel,string,System.Exception,object[]) (at /Users/runner/work/sentry-unity/sentry-unity/src/Sentry.Unity/UnityLogger.cs:41)
Sentry.Extensibility.DiagnosticLoggerExtensions:LogIfEnabled (Sentry.Extensibility.IDiagnosticLogger,Sentry.SentryLevel,System.Exception,string)
Sentry.Extensibility.DiagnosticLoggerExtensions:LogInfo (Sentry.Extensibility.IDiagnosticLogger,string)
Sentry.Unity.Editor.iOS.NativeMain:ContainsSentry (string,Sentry.Extensibility.IDiagnosticLogger)
Sentry.Unity.Editor.iOS.NativeMain:AddSentry (string,Sentry.Extensibility.IDiagnosticLogger)
[...]

PR #457, opened 2021-12-09
fix: Added a file exists check for sentry-cli exec

bitsandfoxes:
Fixes: https://github.com/getsentry/sentry-unity/issues/456

Now there is a check in place to prevent us from trying to override an already existing sentry-cli executable.

### DIFF
--- a/samples/unity-of-bugs/Assets/Resources/Sentry/SentryOptions.asset
+++ b/samples/unity-of-bugs/Assets/Resources/Sentry/SentryOptions.asset
@@ -35,4 +35,4 @@ MonoBehaviour:
   <AndroidNativeSupportEnabled>k__BackingField: 1
   <Debug>k__BackingField: 1
   <DebugOnlyInEditor>k__BackingField: 0
-  <DiagnosticLevel>k__BackingField: 0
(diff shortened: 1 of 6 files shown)


samples/unity-of-bugs/Assets/Resources/Sentry/SentryOptions.asset
```diff
@@ -35,4 +35,4 @@ MonoBehaviour:
   <AndroidNativeSupportEnabled>k__BackingField: 1
   <Debug>k__BackingField: 1
   <DebugOnlyInEditor>k__BackingField: 0
-  <DiagnosticLevel>k__BackingField: 2
+  <DiagnosticLevel>k__BackingField: 0
```
bruno-garcia, line 38: Intentional?

bruno-garcia (approved):

samples/unity-of-bugs/Assets/Resources/Sentry/SentryOptions.asset
bitsandfoxes: Thanks, missed that.

PR #459, opened 2021-12-09
Move assembly aliasing into the MsBuild pipeline

SimonCropp:
#skip-changelog


bruno-garcia:
As we agreed to drop the msbuild nuget package, I believe this PR can be closed.

Ideally a new PR calling the alias tool as a post-build script or something would land instead

PR #462, opened 2021-12-10
fix: Sample OOM button iOS exclusive

bitsandfoxes:
#skip-changelog

### DIFF
--- a/samples/unity-of-bugs/Assets/Scripts/MobileNativeSupport/NativeSupportButtons.cs
+++ b/samples/unity-of-bugs/Assets/Scripts/MobileNativeSupport/NativeSupportButtons.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using System.Collections.Generic;
 #if ENABLE_IL2CPP || PLATFORM_IOS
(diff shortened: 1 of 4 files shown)


samples/unity-of-bugs/Assets/Scripts/MobileNativeSupport/NativeSupportButtons.cs
```diff
@@ -8,6 +9,18 @@
...
+
+    private void Start()
+    {
+#if !UNITY_IOS
+        if (_outOfMemoryButton != null)
+        {
+            Destroy(_outOfMemoryButton);
+        }
+#endif
+    }
```
bruno-garcia, line 22: Would this work?

```suggestion
#if UNITY_IOS
    [SerializeField] private GameObject _outOfMemoryButton;
#endif
```

bruno-garcia (approved):

samples/unity-of-bugs/Assets/Scripts/MobileNativeSupport/NativeSupportButtons.cs
bitsandfoxes: The GameObject `OOM Sample Button` exists and is hooked up for sending click events in the Editor. We just destroy it on any platform that is not iOS.

PR #474, opened 2021-12-15
fix: bump submodules

bruno-garcia:
NOTE: I had to push a branch to the .NET SDK since the addition of a .targets there broken the inheritance flow to this repo:

https://github.com/getsentry/sentry-dotnet/pull/1397

I suggest we move on with this to unblock the next release. And iterate from there

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,8 @@ jobs:
         with:
           submodules: recursive
 
+      - run: echo "::add-mask::${{ secrets.LICENSE_SERVER_URL }}"
+
       - name: Setup .NET SDK (v6.0)
         uses: actions/setup-dotnet@v1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,7 +173,7 @@ jobs:
         if: ${{ matrix.unity-version != '2022.1.0a12' }}
         env:
           UNITY_VERSION: ${{ matrix.unity-version }}
-        run: dotnet msbuild /t:UnityPlayModeTest /p:Configuration=Release /p:OutDir=other
+        run: dotnet msbuild /t:UnityPlayModeTest /p:Configuration=Release /p:OutDir=other test/Sentry.Unity.Tests
 
       - name: Upload test artifacts (playmode)
         if: ${{ failure() }}
@@ -188,7 +188,7 @@ jobs:
         if: ${{ matrix.unity-version != '2022.1.0a12' }}
         env:
           UNITY_VERSION: ${{ matrix.unity-version }}
-        run: dotnet msbuild /t:UnityEditModeTest /p:Configuration=Release /p:OutDir=other
+        run: dotnet msbuild /t:UnityEditModeTest /p:Configuration=Release /p:OutDir=other test/Sentry.Unity.Editor.Tests
 
       - name: Upload test artifacts (editmode)
         if: ${{ failure() }}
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -152,6 +157,21 @@ jobs:
         shell: pwsh
         run: New-Item -ItemType SymbolicLink -Path "C:\${{ matrix.unity-version }}" -Target "C:\Program Files\Unity\Hub\Editor\${{ matrix.unity-version }}"
 
+      - name: Install Unity UPM Packages
```
bruno-garcia, line 160: These steps are all called by Build but to get their timing more clearly, I'm adding them here explicitly here.
They only run if their output hasn't being built yet so the time spent on it will move from the "Build Unity" step to here

.github/workflows/ci.yml
```diff
@@ -168,7 +188,7 @@ jobs:
         if: ${{ matrix.unity-version != '2022.1.0a12' }}
         env:
           UNITY_VERSION: ${{ matrix.unity-version }}
-        run: dotnet msbuild /t:UnityPlayModeTest /p:Configuration=Release /p:OutDir=other
+        run: dotnet msbuild /t:UnityPlayModeTest /p:Configuration=Release /p:OutDir=other test/Sentry.Unity.Tests
```
bruno-garcia, line 191: Added the project path explicitly so we're not running this target against the solution, and hence, against all projects in the solution. They are designed to spawn Unity once only

Directory.Build.targets
```diff
@@ -6,16 +6,19 @@
     <AndroidBuildMethod>Builder.BuildAndroidIl2CPPPlayer</AndroidBuildMethod>
     <AndroidBuildPath>$(PlayerBuildPath)Android/IL2CPP_Player.apk</AndroidBuildPath>
     <!-- Assumes running `dotnet` from the root of the repo: -->
-    <UnitySampleProjectUnityVersion>$(SolutionDir)samples/unity-of-bugs/ProjectSettings/ProjectVersion.txt</UnitySampleProjectUnityVersion>
+    <RepoRoot>$([System.IO.Path]::GetDirectoryName($([MSBuild]::GetPathOfFileAbove('.gitignore', '$(MSBuildThisFileDirectory)'))))/</RepoRoot>
```
bruno-garcia, line 9: So we can build individual projects (`dotnet build src/Sentry.Unity`), we couldn't rely on `SolutionDir` because that didn't expand when not buiilding the `.sln` (which it's what we're doing when we do `dotnet build` on a directory with a .sln in it).

So changing things here to be relative to the root of the repo based on this new property

Directory.Build.targets
```diff
@@ -212,7 +214,8 @@ Related: https://forum.unity.com/threads/6572-debugger-agent-unable-to-listen-on
   <!-- If Unity Library Project doesn't exist, create a Unity project. We use this project to restore packages needed to build
   this solution without using the sample project which depends on the output of this build. -->
   <Target Name="UnityCreatePackages"
-    Condition="!Exists('$(UnityPackageProject)') AND '$(MSBuildProjectName)' == 'Sentry.Unity'">
+    Condition="!Exists('$(UnityPackageProject)') AND '$(MSBuildProjectName)' == 'Sentry.Unity'"
+    AfterTargets="FindUnity">
```
bruno-garcia, line 218: `$UnityPackageProject` depends on `FindUnity` so this was missing

test/Directory.Build.props
```diff
@@ -26,6 +22,7 @@
         <Private>false</Private>
       </Reference>
     </ItemGroup>
+    <Error Condition="!Exists('$(UnityTestPath)/UnityEngine.TestRunner.dll')" Text="TestRunner not found. Expected: $(UnityTestPath)/UnityEngine.TestRunner.dll"></Error>
```
bruno-garcia, line 25: To help debug when `UnityTestPath` didnt' expand properly.

test/Sentry.Unity.Tests/UnityEventProcessorTests.cs
```diff
@@ -176,42 +176,6 @@ static SentryEvent CreateSentryEvent()
                     }
                 };
         }
-
-        [Test]
-        public void SentrySdkCaptureEvent_WrongDsn_CorrectException()
```
bruno-garcia, line 181: This test isn't really doing anything other than testing DNS and our internal logging. Also, it started failing because of a change in the .NET SDK.

bitsandfoxes (approved):

Issue #390, opened 2021-10-27
Set CrashedLastRun outside observer


bruno-garcia:
The Android and iOS integrations are inconsistent in terms of hooking to the native SDK's `CrashedLastRun`:

On Android, it's done in the `Configure` method:

https://github.com/getsentry/sentry-unity/blob/9c36b5b420aeaf74407247a0b48bf834b08f48b9/src/Sentry.Unity.Android/SentryNativeAndroid.cs#L20-L37

On iOS it's done inside the constructor of the Observer:

https://github.com/getsentry/sentry-unity/blob/9c36b5b420aeaf74407247a0b48bf834b08f48b9/src/Sentry.Unity.iOS/IosNativeScopeObserver.cs#L14

Ideally this wouldn't be done inside the observer, so aligning with the Android approach is ideal

PR #488, opened 2021-12-21
ref: Crashed last run for iOS

bitsandfoxes:
Solves: https://github.com/getsentry/sentry-unity/issues/390

#skip-changelog


SimonCropp:
should a test be change in relation to this ?

bruno-garcia (approved):
Just a refactor, I think it's OK to skip a changelog entry, not relevant to consumers

PR #495, opened 2021-12-22
Tag whether event was created in the UI thread

### LINKED ISSUE #62: Tag whether event was created in the UI thread
We should add a tag that indicates whether the event was raised from the UI thread or not

SimonCropp:
fixes  #62

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Tag whether event was created in the UI thread ([#495](https://github.com/getsentry/sentry-unity/pull/495))
+
 ## 0.9.1
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


lucas-zimerman:
PR looks ok for me.

@bruno-garcia , do you know if it's possible to add the main thread info next to the mechanism? picture for reference
image

So we could have a new item called [main_thread] or [main thread] and set True/False.

Event for reference: https://sentry.io/organizations/sentry-sdks/issues/2866761831/?project=5428557&query=&statsPeriod=90d

SimonCropp:
are the broken builds expected?

bitsandfoxes:
> are the broken builds expected?

No, not really. Main is breaking too without apparent changes.

bruno-garcia:
> PR looks ok for me.
> 
> 
> 
> @bruno-garcia , do you know if it's possible to add the main thread info next to the mechanism? picture for reference
> 
> image
> 
> 
> 
> So we could have a new item called [main_thread] or [main thread] and set True/False.
> 
> 
> 
> Event for reference: https://sentry.io/organizations/sentry-sdks/issues/2866761831/?project=5428557&query=&statsPeriod=90d

Yeah this is something that makes sense to align across all sdks. @philipphofmann I believe had a DACI about this. Or possibly @marandaneto who is away until beginning of January in vacation. Let's figure this out before we release this.

Issue #465, opened 2021-12-13
Sentry/Sentry.h file not found


ChrisColeAttensi:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
Unity 0.8.0

### Steps to Reproduce

1. Build for iOS on windows PC
2. zip and copy to mac
3. unzip and open in xcode
4. archive

### Expected Result

Build not failing

### Actual Result

Build fails with an error: 'Sentry/Sentry.h' file not found.

bitsandfoxes:
Hey @ChrisColeAttensi, sorry to see you running into this issue. Could you take a look if you locate Sentry at `Frameworks/Sentry.framework`? Just to narrow it down?

ChrisColeAttensi:
Doesn't look like its there

![image](https://user-images.githubusercontent.com/86221754/145823292-70650727-539b-445b-888f-85cd07dde4b4.png)

bitsandfoxes:
Thanks for the confirmation. So it's not even copying the framework. I'll take a look and get back to you.

fahall:
We're running into the same problem in our CI/CD pipeline after upgrading from the old sentry unity lite package to the new package on v0.8.0. 

We're running on a Github Actions Mac-OS runner using: https://github.com/kuler90/build-unity , so I suspect this isn't unique to building on Windows.

bruno-garcia:
I believe this was resolved with this PR: https://github.com/getsentry/sentry-unity/pull/457

fahall:
I tried building/deploying today with version 0.9.1 with the same result. I don't think the issue is resolved.

bitsandfoxes:
Could you provide us with the build logs? Copying the framework is the very first thing it does.
![Screenshot 2021-12-21 at 11 28 39](https://user-images.githubusercontent.com/25725783/146914793-0264950c-7188-471d-80db-b704a47f0113.png)

fahall:
[logs_11136.zip](https://github.com/getsentry/sentry-unity/files/7758838/logs_11136.zip)

Logs attached. 

The `Build` action succeeds. It fails during the `Deploy` phase, which includes all the XCode/Testflight steps.

fahall:
Any updates on this?

ChrisColeAttensi:
Yes please. This is a showstopper for us and I've had to downgrade all the way to the old Sentry-Unity package.

fahall:
[...]

PR #496, opened 2021-12-22
fix: Added check for Sentry.framework copy success

bitsandfoxes:
Issues (e.g. https://github.com/getsentry/sentry-unity/issues/465) about successful Unity builds but missing Sentry.framework keep coming up.
If we failed to copy the framework to the target we throw and log the paths so at least we know what's going on.

#skip-changelog

### DIFF
--- a/src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
+++ b/src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
@@ -21,7 +21,10 @@ namespace Sentry.Unity.Editor.iOS
 
             try
             {
-                CopyFrameworkToBuildDirectory(pathToProject, options?.DiagnosticLogger);
+                var frameworkDirectory = PlayerSettings.iOS.sdkVersion == iOSSdkVersion.DeviceSDK ? "Device" : "Simulator";
+                var pathToFramework = Path.GetFullPath(Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins", "iOS", frameworkDirectory, "Sentry.framework"));
+
+                CopyFrameworkToBuildDirectory(pathToProject, pathToFramework, options?.DiagnosticLogger);
 
                 using var sentryXcodeProject = SentryXcodeProject.Open(pathToProject);
                 sentryXcodeProject.AddSentryFramework();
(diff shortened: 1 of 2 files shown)


lucas-zimerman (approved):

src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
```diff
@@ -84,20 +87,17 @@ private static void CopyFrameworkToBuildDirectory(string pathToProject, IDiagnos
...
-
-            var frameworkPath = Path.GetFullPath(Path.Combine("Packages", packageName, "Plugins", "iOS", frameworkDirectory, "Sentry.framework"));
-            if (Directory.Exists(frameworkPath))
+            if (Directory.Exists(pathToSentryFramework))
             {
-                logger?.LogDebug("Copying Sentry.framework from '{0}' to '{1}'", frameworkPath, targetPath);
+                logger?.LogDebug("Copying 'Sentry.framework' from '{0}' to '{1}'", pathToSentryFramework, targetPath);
 
-                Directory.CreateDirectory(Path.Combine(pathToProject, "Frameworks"));
-                FileUtil.CopyFileOrDirectoryFollowSymlinks(frameworkPath, targetPath);
```
SimonCropp, line 96: is there a reason for the `FollowSymlinks` removal?

test/Sentry.Unity.Editor.iOS.Tests/BuildPostProcessorTests.cs
```diff
@@ -0,0 +1,70 @@
...
+using NUnit.Framework;
+using Sentry.Unity.Tests.SharedClasses;
+
+namespace Sentry.Unity.Editor.iOS.Tests
+{
+    public class BuildPostProcessorTests
+    {
+        public class Fixture
+        {
+            public string TestDirectoryPath { get; set; }
```
SimonCropp, line 13: i dont understand the `Fixture` pattern. why not make these fields on `BuildPostProcessorTests`?

bitsandfoxes: Very valid point. I had something else in mind with the fixture and then just stuck with it.

src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
bitsandfoxes: According to the [docs](https://docs.unity3d.com/ScriptReference/FileUtil.CopyFileOrDirectoryFollowSymlinks.html) FollowSymlinks is `macOS only` and issues keep coming up with the copy-process failing on other platforms than macOS. We're not relying on symlinks so I don't see a reason why we should special case it.

bruno-garcia (approved):

src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
SimonCropp: thanks :)

SimonCropp (approved):

PR #497, opened 2021-12-22
ci: Add Default Platform Label to PRs and Issues

philipphofmann:
#skip-changelog


lucas-zimerman (approved):

bruno-garcia (changes requested):
I believe we're changing the approach

PR #503, opened 2022-01-04
Update sentry-dotnet module

SimonCropp:


SimonCropp:
@bruno-garcia i must be missing something. didnt we expect this to fail the build?

PR #511, opened 2022-01-11
Programmatic initialization

bitsandfoxes:


bitsandfoxes:
This approach is dropped due to partial classes being required to exist in the same assembly.

PR #515, opened 2022-01-12
Remove jcenter repository

lucas-zimerman:
Unity is failing by having Jcenter repository by default.

This PR uses the default baseProjectTemplate.gradle and replaces JCenter by mavenCentral.

#skip-changelog.


romtsn:
Note, there's an ongoing incident on jcenter/gradle https://status.gradle.com/, should be eventually resolved. But if unity doesn't rely on anything on jcenter, it's a good improvement anyway :)

bitsandfoxes:
I'm not sure I agree with that one. It would mean CI no longer tests against a default Unity setup? For what seems to be a temporary issue?

bruno-garcia:
things are working again without this PR so lets close it for now

Issue #499, opened 2021-12-23
Oculus Quest 2: Crashes (sigsegv) the app on null dereferences


johanhelsing-attensi:
### Environment

How do you use Sentry?
sentry.io

Which SDK and version?
0.9.1

### Steps to Reproduce

Built an android app, installed it on oculus quest 2 and threw an exception.

### Expected Result

Error message is sent to sentry.io, app continues running

### Actual Result

No error sent. App quits instantaneously. Logcat shows this:

```
12-23 10:07:33.472 11702 11702 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
12-23 10:07:33.472 11702 11702 F DEBUG   : Build fingerprint: 'oculus/hollywood/hollywood:10/QQ3A.200805.001/21985310426300000:user/release-keys'
12-23 10:07:33.472 11702 11702 F DEBUG   : Revision: '0'
12-23 10:07:33.472 11702 11702 F DEBUG   : ABI: 'arm'
12-23 10:07:33.473 11702 11702 F DEBUG   : Timestamp: 2021-12-23 10:07:33+0100
12-23 10:07:33.473 11702 11702 F DEBUG   : pid: 11572, tid: 11604, name: UnityMain  >>> com.MyCompany.MyProject <<<
12-23 10:07:33.473 11702 11702 F DEBUG   : uid: 10095
12-23 10:07:33.473 11702 11702 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
12-23 10:07:33.473 11702 11702 F DEBUG   : Cause: null pointer dereference
12-23 10:07:33.473 11702 11702 F DEBUG   :     r0  00000000  r1  bb5583c0  r2  00000000  r3  00000000
12-23 10:07:33.473 11702 11702 F DEBUG   :     r4  00000000  r5  00000000  r6  00000000  r7  aaeec9c0
12-23 10:07:33.473 11702 11702 F DEBUG   :     r8  00000000  r9  cdafc360  r10 00000000  r11 cdafbce8
12-23 10:07:33.473 11702 11702 F DEBUG   :     ip  ebae6928  sp  cdafbcd0  lr  00000000  pc  a011dac0
12-23 10:07:33.921 11702 11702 F DEBUG   :
12-23 10:07:33.921 11702 11702 F DEBUG   : backtrace:
[...]

PR #524, opened 2022-01-19
fix: Reinitializing native backend for IL2CPP only

bitsandfoxes:
Aims to fix: https://github.com/getsentry/sentry-unity/issues/499

On Mono, when an actual crash happens and we reinitialized the native backend the app does what it's supposed to do when it crashes. It crashes. To prevent this we only reinizialize when building with IL2CPP.

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -31,7 +31,11 @@ namespace Sentry.Unity
 #if SENTRY_NATIVE_IOS
                 SentryNativeIos.Configure(options);
 #elif SENTRY_NATIVE_ANDROID
+#if ENABLE_IL2CPP
+                SentryNativeAndroid.Configure(options, true);
+#else
                 SentryNativeAndroid.Configure(options);
+#endif
 #endif
 
                 SentryUnity.Init(options);
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -31,11 +31,11 @@ namespace Sentry.Unity
 #if SENTRY_NATIVE_IOS
                 SentryNativeIos.Configure(options);
 #elif SENTRY_NATIVE_ANDROID
-var il2cpp =
+                var il2cpp =
 #if ENABLE_IL2CPP
-        true;
+                true;
 #else
-        false;
+                false;
 #endif
                 SentryNativeAndroid.Configure(options, il2cpp);
 #endif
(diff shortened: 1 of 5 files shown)


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -31,7 +31,11 @@ public static void Init()
 #if SENTRY_NATIVE_IOS
                 SentryNativeIos.Configure(options);
 #elif SENTRY_NATIVE_ANDROID
+#if ENABLE_IL2CPP
+                SentryNativeAndroid.Configure(options, true);
+#else
                 SentryNativeAndroid.Configure(options);
+#endif
```
bruno-garcia, line 38: What about we remove the default argument and always pass a value?

```suggestion
                var il2cpp = false;
#if ENABLE_IL2CPP
               il2cpp = true;
#endif
                SentryNativeAndroid.Configure(options, il2cpp);
```

src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -11,7 +12,7 @@ public static class SentryNativeAndroid
         /// Configures the native Android support.
         /// </summary>
         /// <param name="options">The Sentry Unity options to use.</param>
-        public static void Configure(SentryUnityOptions options)
+        public static void Configure(SentryUnityOptions options, bool isIL2CPP = false)
```
bruno-garcia, line 15: If you take suggestion above, then we'd always require the value:

```suggestion
        public static void Configure(SentryUnityOptions options, bool il2cpp)
```

src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -35,9 +36,13 @@ public static void Configure(SentryUnityOptions options)
 
                     return crashedLastRun.Value;
                 };
-                // At this point Unity has taken the signal handler and will not invoke the original handler (Sentry)
-                // So we register our backend once more to make sure user-defined data is available in the crash report.
-                SentryNative.ReinstallBackend();
+
+                if (isIL2CPP)
```
bruno-garcia, line 40:
```suggestion
                // When running on Mono, we shouldn't take over the signal handler because its used to propagate exceptions into the VM.
                // If we take over, a C# null reference ends up crashing the app.
                if (isIL2CPP)
```

package-dev/Runtime/SentryInitialization.cs
lucas-zimerman: What about
```csharp
var il2cpp = 
#if ENABLE_IL2CPP
        true;
#else
        false;
#endif
``` 
?

bruno-garcia (approved):

lucas-zimerman:
```
           > Could not HEAD 'https://jcenter.bintray.com/org/jetbrains/kotlinx/binary-compatibility-validator/org.jetbrains.kotlinx.binary-compatibility-validator.gradle.plugin/0.7.1/org.jetbrains.kotlinx.binary-compatibility-validator.gradle.plugin-0.7.1.jar'.
              > Read timed out
```
CI error is unrelated to the PR so I'm restarting the test.

PR #531, opened 2022-01-25
feat: SentryUnityInfo for build-time variables at runtime

bitsandfoxes:
The idea is to "cache" build time variables and make them available during runtime like `IL2CPP` or `PlayerSettings.forceSingleInstance`.

#skip-changelog


bruno-garcia (approved):
Nice!

Issue #521, opened 2022-01-18
Import error when creating link.xml on Windows


bitsandfoxes:
When opening the config window for the first time (the link.xml does not yet exist) we copy the file in OnEnable. This seems to create an endless import-loop that gets stopped by the editor.
This does not impact the functionality but just looks bad.

PR #539, opened 2022-02-01
fix: Importing link.xml no longer causes an infinite loop

bitsandfoxes:
The config window class was getting quite big and hard to navigate.
Also fixes https://github.com/getsentry/sentry-unity/issues/521 by calling `Awake` instead of `OnEnable` as OnEnable can retrigger during the import process.


src/Sentry.Unity.Editor/ConfigurationWindow/SentryEditorWindowInstrumentation.cs
```diff
@@ -4,7 +4,7 @@
 using UnityEditor;
 using UnityEngine;
 
-namespace Sentry.Unity.Editor
+namespace Sentry.Unity.Editor.ConfigurationWindow
```
lucas-zimerman, line 7: line 170 of Directoryu.Build.Targets needs to be changed for:
```xml
    <Exec Command="&quot;$(UnityExec)&quot; -quit -batchmode -nographics -logFile - -projectPath $(UnitySampleProjectPath) -executeMethod Sentry.Unity.Editor.ConfigurationWindow.SentryEditorWindowInstrumentation.ConfigureOptions -sentryOptions.Dsn $(TestDsn)" IgnoreStandardErrorWarningFormat="true">
        <Output TaskParameter="ExitCode" PropertyName="UnityConfigureSentryOptionsExitCode"/>
    </Exec>
```

bitsandfoxes: thanks for pointing that out!

PR #540, opened 2022-02-01
draft: Added an Embed package button 

bitsandfoxes:
Instead of pointing users to the docs telling them to read up on how to embed a package so they can modify it, we can add a button at the bottom of the `Advanced` tab that does that for them.


bruno-garcia:
Is there a screenshot of what this looks like? seems to include changes from anotther PR

bitsandfoxes:
> Is there a screenshot of what this looks like? seems to include changes from anotther PR

It sits on top of the config window split PR.
Screenshot 2022-02-01 at 19 07 59

PR #545, opened 2022-02-04
use dotnet format of net6

bruno-garcia:
#skip-changelog

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -26,9 +26,6 @@ jobs:
       - name: Format C# Code Whitespace
         run: dotnet format whitespace Sentry.Unity.sln --exclude src/sentry-dotnet --verbosity diag
 
-      - name: Format C# Code Style
-        run: dotnet format style Sentry.Unity.sln --exclude src/sentry-dotnet --verbosity diag
-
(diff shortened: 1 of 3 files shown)


bruno-garcia:
Not sure why it fails... need to run the script (or command by command) in the terminal to see what's going on.

Might be that's not happy that there are changes in the submodule, since dotnetformat ends up running inside the dotnet SDK. If that's the case we'll need to filter it (don't run it under src/sentry-dotnet)

.github/workflows/format-code.yml
```diff
@@ -20,11 +23,12 @@ jobs:
...
-        run: dotnet tool install -g dotnet-format
-      - name: Format C# Code
-        run: dotnet format
-      
+      - name: Format C# Code Whitespace
+        run: dotnet format whitespace Sentry.Unity.sln --exclude src/sentry-dotnet --verbosity diag
+
+      - name: Format C# Code Style
+        run: dotnet format style Sentry.Unity.sln --exclude src/sentry-dotnet --verbosity diag
+
```
bruno-garcia, line 31:
```suggestion
```

bitsandfoxes (approved):

Issue #527, opened 2022-01-21
Windows native support


bitsandfoxes:
## Requirements: 

- Capturing native errors
- Capturing minidumps: https://github.com/getsentry/sentry-unity/issues/94
- Sync scope from Unity SDK to native SDK
- `crashed last run` functionality to track healthy sessions
Currently not supported by the native SDK 

## Things to consider:

- Breakpad
  - Less robust than Crashpad but does not require additional executable shipped with the build output
- Crashpad
  - Requires shipping with a `crashpad_handler.exe` with the build output
  - Users might not want to do that 
- We could offer a toggle between those two in addition to `Enable Window Native Support`? 

## TODOs and open questions:
* [x] Initial support for x86_64 IL2CPP + crasphad: #380
* [x] implement NativeScopeObserver (scope, breadcrumbs, extra, ...) - #546
* [x] do we want to support x86 & ARM64? -- wait for the real interest from the userbase
* [x] does it make sense to run with Mono scripting backend?
* [x] logging from native to the debug logger in the managed code? I don't think I've seen that in other integrations either... #577 
* [x] unit tests for the native crashes? #577 
* [x] upload symbols using CLI #607
* [ ] add call to `sentry_init` in the `main` function at compile time
* [x] `bool CrashedLastRun()` #615 
* [x] Packaging requires change in CI: We need artifacts from Mac and Windows so packaging requires a new job dependant on those - #547 
* [ ] Documentation

bruno-garcia:
Regarding initializing the code in the Launcher instead of after the engine runs, we're investigating (#622) recompiling the playback engine (game executable). That's documented here: https://docs.unity.cn/Manual/WindowsStandaloneBinaries.html

bruno-garcia:
Opened an issue to the follow up change

PR #546, opened 2022-02-04
Feat/native scope observer

vaind:
Originally this was supposed to implement native scope observer (for windows support #527), but I've noticed the scope observer implementations share some code and didn't want to duplicate it further so instead went for a little refactoring.

#skip-changelog

Native crash breadcrumbs shown here: https://sentry.io/organizations/sentry-sdks/issues/2988392337/events/53be944157b54ee59586dbb5f8fdd8c8/?project=5439417


src/Sentry.Unity/IScopeObserver.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        /// <summary>
+        /// Adds a breadcrumb.
+        /// </summary>
+        void AddBreadcrumb(Breadcrumb breadcrumb);
+
+        /// <summary>
+        /// Sets an extra.
+        /// </summary>
+        void SetExtra(string key, string value);
```
bruno-garcia, line 16: This interface is defined in the Sentry SDK (added as a submodule) are you redefining it here because of `string` value on Extra?

bruno-garcia: Perhaps best we just change the one in the .NET SDK to avoid having these two interface 'look-alikes' live in the two assemblies

bruno-garcia (commented):
This is really great work!

src/Sentry.Unity.Native/NativeScopeObserver.cs
```diff
@@ -1,59 +1,104 @@
+using System;
+using System.Runtime.InteropServices;
 using Sentry.Extensibility;
-using Sentry.Unity.Json;
-using UnityEngine;
 
-namespace Sentry.Unity.Native
+namespace Sentry.Unity
```
bruno-garcia, line 5: Having the type on this namespace is intentional?

bruno-garcia (commented):
Nice approach!

src/Sentry.Unity/IScopeObserver.cs
vaind: reworked as discussed yesterday.

src/Sentry.Unity.Native/NativeScopeObserver.cs
vaind: unintentional, fixed now
```suggestion
namespace Sentry.Unity.Native
```

src/Sentry.Unity.Android/AndroidJavaScopeObserver.cs
```diff
@@ -8,22 +6,16 @@ namespace Sentry.Unity.Android
...
     /// <see href="https://github.com/getsentry/sentry-java"/>
-    public class AndroidJavaScopeObserver : IScopeObserver
+    public class AndroidJavaScopeObserver : ScopeObserver
     {
-        private readonly SentryOptions _options;
-
-        public AndroidJavaScopeObserver(SentryOptions options) => _options = options;
+        public AndroidJavaScopeObserver(SentryOptions options) : base("Android", options) { }
 
+        // TODO can we use a local var `AndroidJavaClass _sentryJava` initialized in a constructor instead?
```
bruno-garcia, line 13: I believe the instance can't be shared across threads. I tried on my first attempt and learned with some errors, if my memory doesn't fail me. Also, we need to lazily instantiate things when possible to add less overhead during "init", so if possible to cache the reference, we could, but not in the ctor in this case

```suggestion
```

src/Sentry.Unity.Android/AndroidJavaScopeObserver.cs
```diff
@@ -34,76 +26,49 @@ public void AddBreadcrumb(Breadcrumb breadcrumb)
...
-                extraValue = SafeSerializer.SerializeSafely(value);
-                if (extraValue is null)
-                {
-                    return;
-                }
-            }
-
-            using var sentry = GetSentryJava();
-            sentry.CallStatic("setExtra", key, extraValue);
+            GetSentryJava().CallStatic("setExtra", key, value);
```
bruno-garcia, line 32: Sematics changed here because the Java objects need to be disposed. Before we had:

**using** var sentry = GetSentryJava();

Now the dispose is not called.

We need to go back to:

```
            using var sentry = GetSentryJava();
            sentry.CallStatic("setExtra", key, extraValue);
```

bruno-garcia (changes requested):

src/Sentry.Unity.Android/AndroidJavaScopeObserver.cs
vaind: I see... somehow my mind (non-c# developer) hasn't expected `using` to do anything like that... I'll roll back the change, thanks

src/Sentry.Unity.Native/SentryNativeBridge.cs
```diff
@@ -68,4 +68,4 @@ public static void Init(SentryUnityOptions options)
         // Testing
         internal static Action ReinstallSentryNativeBackendStrategy = sentry_reinstall_backend;
     }
-}
+}
```
bruno-garcia, line 71:
```suggestion
}

```

bruno-garcia (approved):

Issue #522, opened 2022-01-18
Stacktrace does not match Unity log when switching from 'Script Only' to 'Full'


bitsandfoxes:
Came from: https://discord.com/channels/621778831602221064/621783582628642822/930258591452123147

1. Create new Unity 3D project with Unity 2021.2.4f1.
2. Import Sentry Unity 0.9.2 package via Git URL.
3. Add High Definition RP through Unity Package Manager.
4. Duplicate "Directional Light" once in the default "SampleScene" so that there are now two.
5. For unrelated reasons, be forced to restart Unity once or twice and have it crash on you twice. (this happened to me, your mileage may vary)
6. Set a valid DSN in Sentry options window.
7. Go to "Player Settings" and in the "Stack Trace" options checkboxes, switch it from "ScriptOnly" to "Full" for errors and exceptions.
8. Run the game in editor and see the  "Cascade Shadow atlasing has failed, only one directional light can cast shadows at a time" error in Console output.
9. View this issue in Sentry dashboard, and see that the call stack only has the addresses, but the callstack seen in the editor's console output had much more information. 

[unity-error-sentry-repro (1).zip](https://github.com/getsentry/sentry-unity/files/7891672/unity-error-sentry-repro.1.zip)

bitsandfoxes:
Some context information:
When switching to 'Full' Unity provides additional string as stacktrace that we do not yet support by parsing.
 
![Screenshot 2022-01-18 at 20 37 41](https://user-images.githubusercontent.com/25725783/150006615-8b59e9e3-5776-47c1-8184-979355aa3797.png)

```
Cascade Shadow atlasing has failed, only one directional light can cast shadows at a time
 #0 GetStacktrace(int)
 #1 DebugStringToFile(DebugStringToFileData const&)
 #2 DebugLogHandler_CUSTOM_Internal_Log(LogType, LogOption, ScriptingBackendNativeStringPtrOpaque*, ScriptingBackendNativeObjectPtrOpaque*)
 #3  (Mono JIT Code) (wrapper managed-to-native) UnityEngine.DebugLogHandler:Internal_Log (UnityEngine.LogType,UnityEngine.LogOption,string,UnityEngine.Object)
[...]

PR #549, opened 2022-02-07
Add support for converting full stack trace into SentryException

foriequal0:
Closes: #522


src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -90,6 +94,15 @@ internal void OnLogMessageReceived(string? condition, string? stackTrace, LogTyp
             _hub.AddBreadcrumb(message: condition, category: "unity.logger", level: ToBreadcrumbLevel(type));
         }
 
+        private static StackTraceLogType GetStackTraceLogType()
+        {
+#if UNITY_EDITOR
+            return PlayerSettings.GetStackTraceLogType(logType);
+#else
+            return StackTraceLogType.ScriptOnly;
```
bruno-garcia, line 102: Couldn't it be Full outside of the editor?

src/Sentry.Unity/UnityLogException.cs
```diff
@@ -149,6 +165,216 @@ public SentryException ToSentryException()
...
+                        Module = module,
+                        Function = function,
+                        FileName = file,
+                        LineNumber = line,
+                        InApp = GuessInApp(module, function, wrapper, file),
+                    });
+                }
+            }
+
+            frames.Reverse();
```
bruno-garcia, line 244: We could use a reversed for here and avoid iterating twice

bruno-garcia (commented):
Thank you for adding this. We'd need some tests that Dover the new code paths

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
foriequal0: Oh I didn't know that `Application.GetStackTraceLogType` exists. I've changed it.

src/Sentry.Unity/UnityLogException.cs
```diff
@@ -149,6 +164,215 @@ public SentryException ToSentryException()
...
+            if (LogStackTrace.StartsWith("0x")
+                && LogStackTrace.Contains("(Unity) StackWalker::GetCurrentCallstack"))
+            {
+                return ConvertWindowsFullStackTraceImpl();
+            }
+
+            return ConvertFullStackTraceImpl();
+        }
+
+        private static readonly Regex WindowsStackFramePattern = new Regex(
```
bruno-garcia, line 178: Ideally we'd use the overload that takes a timeout (and passing the Compile flag is probably a good idea too):

```
public Regex (string pattern, System.Text.RegularExpressions.RegexOptions options, TimeSpan matchTimeout);
```
Timeout here for a second should be way more than enough, but a safenet for some bad input causing some really bad thing to happen

src/Sentry.Unity/UnityLogException.cs
```diff
@@ -149,6 +164,215 @@ public SentryException ToSentryException()
...
+                if (string.IsNullOrWhiteSpace(item))
+                {
+                    continue;
+                }
+
+                var match = WindowsStackFramePattern.Match(item);
+                if (!match.Success)
+                {
+                    // we did something wrong, failed the check
+                    Debug.Log("failed parsing " + item);
```
bruno-garcia, line 201: We avoid calling into `Debug.Log` here because the SDK has its own control for logging. The users are allowed to turn it on or off and control the level. If we're left with Debug.Log in the other implementation that this was based from, we should probably change there too

src/Sentry.Unity/UnityLogException.cs
```diff
@@ -149,6 +164,215 @@ public SentryException ToSentryException()
...
+                Stacktrace = stacktrace,
+                Type = LogType.ToString(),
+                Value = LogString,
+                Mechanism = new Mechanism
+                {
+                    Handled = false,
+                    Type = "unity.log"
+                }
+            };
+        }
```
bruno-garcia, line 343: I didn't go through in detail to point out where the opportunities for refactor are but it seems to me there's a lot of duplication here. Particularly in areas that are candidate for change like mechanism.type. Ideally this would be refactored a bit.

bruno-garcia (commented):
Thanks again for the PR. I haven't had a chance to go into details just yet but I did a quick pass and here are my initial thoughts.

bruno-garcia:
@vaind @bitsandfoxes is this something you can help review, please?

src/Sentry.Unity/UnityLogExceptionConverter.cs
```diff
@@ -0,0 +1,343 @@
...
+            => filename.Equals("<00000000000000000000000000000000>", StringComparison.OrdinalIgnoreCase)
+                ? string.Empty
+                : filename;
+
+        // TODO: discuss
+        private static string TryResolveFileNameForMono(string fileName)
+        {
+            try
+            {
+                // throws on Mono for <1231231231> paths
```
bruno-garcia, line 333: best we check somehow the pattern instead of relying on the try catch. It can stay just a last resort though

bruno-garcia:
Not sure why CI isn't running (just formatting and changlog checks). Tried close/open but didn't work.

Please pull `main` into this branch so we can get the new CI pipeine

bitsandfoxes:
Hey @foriequal0. Sorry, it took so long to get back to you and we let your PR go stale but the good news is that we kinda fixed the way we report exceptions, and the SDK no longer relies on parsing strings in the first place! https://github.com/getsentry/sentry-unity/pull/731

Issue #535, opened 2022-01-29
Sample should only show buttons applicable the platform where it runs


bruno-garcia:
The "Native Plugin" example should only display buttons when running on a platform that they work.
This is in preparation to add Windows and WebGL, as the current format will be quite busy if we just add more buttons to it as-is

PR #551, opened 2022-02-07
feat: Made native samples be platform sensitive

bitsandfoxes:
Closes https://github.com/getsentry/sentry-unity/issues/535
The native sample scene buttons now have a top part for the platform-agnostic `Requires IL2CPP`. They turn inactive if not on IL2CPP.

Followed by a platform-specific section for`Android` and `iOS` that is easily expandable for i.e. `Windows`.

nativesample


bruno-garcia (approved):
Looks great

PR #547, opened 2022-02-04
Split CI

vaind:
see #449 

#skip-changelog

PR #556, opened 2022-02-11
Regression: collect android native libraries as an artifact

vaind:
Fixes a regression introduced by #547 - native libs were not collected anymore

TODO: where are these libs used? Do we need them in the final artifact [as they were before](https://github.com/getsentry/sentry-unity/blob/e1ddb54972935499fbca95c1558916b941d7e0c2/.github/workflows/ci.yml#L286) or can we just have a separate artifact "Android-libraries" (or a similar name)

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,11 @@ jobs:
 
       - uses: actions/download-artifact@v2
         with:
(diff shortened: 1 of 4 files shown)


bruno-garcia:
We need them in the same artifact because craft will download thst one zip to work with

PR #558, opened 2022-02-13
Refactored Smoke test on Android

lucas-zimerman:
- Merged both smoke test jobs into a single job.
- Add retry for new flaky case on Android 30

#skip-changelog.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -328,8 +328,7 @@ jobs:
       max-parallel: 5
       fail-fast: false
       matrix:
-        api-level: [21, 27, 29]
-        avd-target: [default]
+        api-level: [21, 27, 29, 30]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -332,11 +332,10 @@ jobs:
     name: Smoke Test - Android ${{ matrix.api-level }} Unity ${{ matrix.unity-version }}
     runs-on: macos-latest
     strategy:
-      max-parallel: 5
+      max-parallel: 7
```
bruno-garcia, line 335: Should we just drop this and let it use as many agents as there are available?

vaind (approved):

.github/workflows/ci.yml
lucas-zimerman: ok

Issue #525, opened 2022-01-20
Sentry.Unity.Editor.iOS and AppDomain.CurrentDomain.GetAssemblies() for Android target platform


PeraDllGD:
### Environment

Which version of the SDK?
0.9.3 (imported with package manager)

Which version of Unity?
2019.4.16f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Happening only in Editor

### Steps to Reproduce

After implementing Sentry problem appeared with custom inspector drawing. One of assets we are using is Lip Sync Pro and when you select game object with lipsync pro script on it (or eye controller script), its not drawing things properly (throwing exception). 
After debugging we found that when its looping through AppDomain.CurrentDomain.GetAssemblies(), and GetTypes() is called for each assembly, its throwing exception and main culprit is Sentry.Unity.Editor.iOS.dll which if is deleted, no exception is thrown. 

Our target platform is Android, so this dll should be excluded from assemblies. 

(temporal solution we found is to update code on multiple places to make it work - with try/catch - some of other assets are using same approach - but this is just for preventing consequences)

bitsandfoxes:
Thanks @PeraDllGD, for raising this. We just recently changed the import settings for `Sentry.Unity.Editor.iOS.dll` and we are aware that this might lead to issues on some of the older Unity versions (we tested it on 2019.4.33f1 and newer).
Do you have the Unity iOS module installed? Are you on Windows or on macOS?

PeraDllGD:
Thanks for reply. Its on Windows, and I don't have iOS module installed for this Unity version, so probably it wasn't able to find it and changed to default one which triggered problem.

bitsandfoxes:
I see that you're on a pretty old Unity version as well. I'm closing this since it a Unity update or iOS module installation will fix it.
Please let us know if you're running into any issues.

PeraDllGD:
Just tried with latest 2019 LTS version (34f1) and problem is still there, and as iOS module is not required for development purpose, do you think its safe deleting that file? Maybe in future update we can have option what to include and what not depending on target platform. Thanks in advance

bitsandfoxes:
Yes, if you don't require the iOS build t target then you can just remove the `Sentry.Unity.Editor.iOS.dll`.
We'll take another look. Could you provide us with your `YourProject/Packages/manifest.json` file? I'd like to be able to reproduce this issue.

PeraDllGD:
[...]

PR #559, opened 2022-02-14
fix: Sentry.Unity.Editor.iOS is 'OSX only' by default

bitsandfoxes:
Closes: https://github.com/getsentry/sentry-unity/issues/525
Instead of requiring users to have the iOS module installed or embed the package and delete the `Sentry.Unity.Editor.iOS.dll` to get rid of the error, we set the default import settings to `OSX only`. 
If a user targets iOS on Windows we log the problem with a solution.

### DIFF
--- a/src/Sentry.Unity.Editor/SentryPreprocessBuild.cs
+++ b/src/Sentry.Unity.Editor/SentryPreprocessBuild.cs
@@ -17,7 +17,7 @@ namespace Sentry.Unity.Editor
             switch (report.summary.platform)
             {
                 case BuildTarget.iOS:
-                    CheckIOSEditorImportSettingsOnWindows(Path.Combine("Packages", SentryPackageInfo.GetName(), "Editor", "iOS",
+                    CheckIOSEditorImportSettings(Path.Combine("Packages", SentryPackageInfo.GetName(), "Editor", "iOS",
                         "Sentry.Unity.Editor.iOS", ".meta"));
                     break;
                 default:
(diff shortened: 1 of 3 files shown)


src/Sentry.Unity.Editor/SentryPreprocessBuild.cs
```diff
@@ -10,32 +14,39 @@ public class SentryPreprocessBuild : IPreprocessBuildWithReport
...
                     break;
             }
-            // file: summary.outputPath
-            // guid: summary.guid
+        }
+
+        internal void CheckIOSEditorImportSettingsOnWindows(string metaFilePath, IApplication? application = null)
+        {
+            application ??= ApplicationAdapter.Instance;
+            if (application.Platform != RuntimePlatform.WindowsEditor)
```
bruno-garcia, line 32: Should this be Windows or Linux? This is anywhere one can be developing with Unity outside of Mac, right?

I guess 0.001% are on Linux but...

src/Sentry.Unity.Editor/SentryPreprocessBuild.cs
```diff
@@ -10,32 +14,39 @@ public class SentryPreprocessBuild : IPreprocessBuildWithReport
...
+
+            if (!File.Exists(metaFilePath))
+            {
+                return;
+            }
+
+            var meta = File.ReadAllText(metaFilePath);
+            if (meta.Contains("OS: OSX"))
+            {
+                new UnityLogger(new SentryUnityOptions()).LogWarning("You're trying to build for iOS on " +
```
bruno-garcia, line 45: The idea of a link to the docs was a good one.
Maybe a sub page on Native Support just about this?

bruno-garcia:
Link to docs can come later btw if you want to expedite this fix

bruno-garcia:
odd, CI didn't run

Issue #449, opened 2021-12-06
Improve CI performance


bruno-garcia:
CI has been continuously built with a single/serial pipeline. That adds to the overall build time as we add new validations, etc.

The goal is to run in parallel what makes sense. And to avoid installing Unity on each run, which can be done with a customer Docker file on Linux:

![image](https://user-images.githubusercontent.com/1633368/144891294-e9abd194-1d2c-4ace-bfd2-4cbca779929b.png)

* [x] SDK split to separate jobs - #547
* [x] create a docker image for the linux build - https://github.com/getsentry/unity-docker
* [ ] prepare the package and run tests in a linux container
* [ ] build & smoke-test players on supported platforms

lucas-zimerman:
@bruno-garcia  not sure how efficient this could be but what if we map a remote disk to Github actions and instead of installing Unity all the time, we just simply use the one that was installed on the remote disk?

vaind:
@lucas-zimerman that's an interesting idea. If we could map a read-only drive from some cloud provider with the unity pre-installed that could be good. 

I was originally thinking about something similar - instead of using github cache (which isn't employed correctly because it constantly hits the size limit) - use a "manual" cache implementation, fetching pre-installed unity zip from AWS s3 bucket and extracting it (while streaming; GH cache seems to download, then extract). So that could be another option if the remote drive didn't work

bruno-garcia:
Consider also: https://github.com/getsentry/sentry-unity/issues/312

bruno-garcia:
@vaind seems like this issue can be closed?

vaind:
> @vaind seems like this issue can be closed?

I'll still do some minor improvements here and there but yes, the overall goal of this issue seems fulfilled

PR #563, opened 2022-02-15
Build in linux docker, test on supported platforms

vaind:
See #449 

Uses a linux+docker based Unity to prepare "build projects" for all platforms and then runs the actual build on the target platforms

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     name: Build - ${{ matrix.unity-version }}
     runs-on: ubuntu-latest
     container:
-      image: dlugos/unity-editor:ubuntu-${{ matrix.unity-version }}-main
+      image: ghcr.io/getsentry/unity-docker:editor-ubuntu-${{ matrix.unity-version }}-main
     # TODO add a "wait" step just before the artifacts are really needed - waiting for the SDK job to complete.
     #      That way, we can start downloading the docker container while the native SDKs are being built.
     #      There's no existing action to do that but a simple polling wait action could use the following API
@@ -158,7 +158,8 @@ jobs:
       - name: Configure Sentry via the Editor Window
         run: dotnet msbuild /t:UnityConfigureSentryOptions /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
 
-      # TODO: can we make unity only prepare the project, without compiling? Same as what it does for iOS.
+      # TODO: can we make unity only prepare the project, without compiling it, to unlock the unity license faster? iOS is already built like that.
+      #       This option should do it: https://docs.unity3d.com/ScriptReference/BuildOptions.AcceptExternalModificationsToPlayer.html
       - name: Build Android Player with IL2CPP
         run: dotnet msbuild /t:UnityBuildPlayerAndroidIL2CPP /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,11 +77,17 @@ jobs:
           # using hash of package.json to bust cache on release builds which modify it
           key: samples/artifacts/builds|${{ matrix.unity-version }}-${{ hashFiles('package/package.json') }}
 
-      - name: Restore Unity Packages
+      - name: Restore Unity Packages (samples)
         uses: actions/cache@v2
         with:
           path: samples/unity-of-bugs/Library
           key: samples/unity-of-bugs/Library|${{ matrix.unity-version }}-${{ hashFiles('samples/unity-of-bugs/Packages/packages-lock.json') }}
+
+      - name: Restore Unity Packages (for target UnityCreatePackages)
+        uses: actions/cache@v2
+        with:
+          path: temp/unity-packages
+          key: temp/unity-packages|${{ matrix.unity-version }}-${{ hashFiles('samples/unity-of-bugs/Packages/packages-lock.json') }}
 
       - uses: actions/download-artifact@v2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,14 +80,15 @@ jobs:
       - name: Restore Unity Packages (samples)
         uses: actions/cache@v2
         with:
-          path: samples/unity-of-bugs/Library
-          key: samples/unity-of-bugs/Library|${{ matrix.unity-version }}-${{ hashFiles('samples/unity-of-bugs/Packages/packages-lock.json') }}
+          path: samples/unity-of-bugs/Library/Packages
+          key: samples/unity-of-bugs/Library/Packages|${{ matrix.unity-version }}-${{ hashFiles('samples/unity-of-bugs/Packages/packages-lock.json') }}
 
-      - name: Restore Unity Packages (for target UnityCreatePackages)
+      # avoids the need to run target UnityCreatePackages
+      - name: Restore TestRunner library
         uses: actions/cache@v2
         with:
-          path: temp/unity-packages
-          key: temp/unity-packages|${{ matrix.unity-version }}-${{ hashFiles('samples/unity-of-bugs/Packages/packages-lock.json') }}
+          path: temp/unity-packages/Library/ScriptAssemblies/*.TestRunner.*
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -232,14 +154,15 @@ jobs:
         run: dotnet msbuild /t:UnityBuildPlayerAndroidIL2CPP /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
 
       - name: Build iOS Player
-        if: matrix.os == 'macos-latest'
         run: dotnet msbuild /t:UnityBuildPlayerIOS /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
 
-      - name: Build Standalone Player with IL2CPP
-        run: dotnet msbuild /t:UnityBuildStandalonePlayerIL2CPP /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
+      # TODO: build for all supported desktop platforms
```
bruno-garcia, line 159: Not sure we can build a Windows standalone player on Linux

vaind: You were right, on Linux you can only target Windows-mono, not IL2CPP... I'm thinking of running a pre-built windows unity container... Should still be much faster then installing unity

.github/workflows/ci.yml
```diff
@@ -88,32 +70,13 @@ jobs:
...
-        with:
-          path: ${{ matrix.unity-installation-path }}
-          key: ${{ matrix.os }}-${{ matrix.unity-version }}-${{ matrix.unity-version-changeset }}
-
-      - name: Restore previous build
-        uses: actions/cache@v2
-        with:
-          path: samples/artifacts/builds
-          # using hash of package.json to bust cache on release builds which modify it
-          key: samples/artifacts/builds|${{ matrix.os }}-${{ matrix.unity-version }}-${{ matrix.unity-version-changeset }}-${{ hashFiles('package/package.json') }}
```
vaind, line 110: This was caching the android (smoke-test) player build but that doesn't make sense (with the key based only on the package.json) - we actually weren't testing code the changes made in the PRs if the cache key hit...

.github/workflows/ci.yml
```diff
@@ -88,32 +70,13 @@ jobs:
...
-
       - name: Restore Unity Packages
         uses: actions/cache@v2
         with:
-          path: samples/unity-of-bugs/Library
-          key: samples/unity-of-bugs/Library|${{ matrix.os }}-${{ matrix.unity-version }}-${{ matrix.unity-version-changeset }}-${{ hashFiles('samples/unity-of-bugs/Packages/packages-lock.json') }}
+          path: |
+            samples/unity-of-bugs/Library/Packages
+            temp/unity-packages/Library/ScriptAssemblies/*.TestRunner.*
```
vaind, line 78: TestRunner wasn't cached - caching it speeds up (skips) `UnityRestorePackages` target

.github/workflows/ci.yml
```diff
@@ -252,43 +165,24 @@ jobs:
...
-        run: ./test/Scripts.Integration.Test/integration-update-sentry.ps1 "${{ env.UNITY_PATH }}"
-
-      - name: Integration Test - Build Standalone Player Sentry SDK
-        shell: pwsh
-        run: ./test/Scripts.Integration.Test/integration-build-project.ps1 "${{ env.UNITY_PATH }}"
-
-      - name: Integration Test - Run Player - Smoke Test
-        shell: pwsh
-        run: ./test/Scripts.Integration.Test/integration-run-smoke-test.ps1
-
```
vaind, line 274: moved to a separate `windows-smoke-test` job below

.github/workflows/ci.yml
```diff
@@ -327,6 +220,73 @@ jobs:
...
+        with:
+          # Artifact name is the commit sha. Which is what craft uses to find the relevant artifact.
+          name: ${{ github.sha }}
+
+      - name: Setup Unity
+        uses: getsentry/setup-unity@46c2e082d98cc3a825a5b59038cb31705fe9ff56
+        with:
+          unity-version: ${{ matrix.unity-version }}
+          unity-version-changeset: ${{ matrix.unity-version-changeset }}
+          unity-modules: windows-il2cpp
```
vaind, line 258: Windows runners can't use containers because of licensing & the limitations of Windows Server 2022 (no support for WSL2), see https://github.com/actions/runner/issues/904

Therefore we stay with the standard unity setup here

samples/unity-of-bugs/Assets/Editor/Builder.cs
```diff
@@ -60,6 +60,7 @@ public static void BuildIl2CPPPlayer(BuildTarget buildTarget)
 
     public static void BuildWindowsIl2CPPPlayer() => BuildIl2CPPPlayer(BuildTarget.StandaloneWindows64);
     public static void BuildMacIl2CPPPlayer() => BuildIl2CPPPlayer(BuildTarget.StandaloneOSX);
+    public static void BuildLinuxIl2CPPPlayer() => BuildIl2CPPPlayer(BuildTarget.StandaloneLinux64);
```
vaind, line 63: this ended up unused but I guess we can keep it for now. Later we can fix the smoke test on linux (currently times out)

vaind (commented):
Some info for a future reviewer

.github/workflows/ci.yml
vaind: This ended up as a separate `windows-smoke-test` job

.github/workflows/ci.yml
```diff
@@ -228,19 +146,14 @@ jobs:
       - name: Configure Sentry via the Editor Window
         run: dotnet msbuild /t:UnityConfigureSentryOptions /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
 
+      # TODO: can we make unity only prepare the project, without compiling it, to unlock the unity license faster? iOS is already built like that.
```
bruno-garcia, line 149: Yes please. Export gradle project.

Downside is that we don't cover some code path the one that uploads symbols when unity builds an apk. But we anyway have 2 paths to test

.github/workflows/ci.yml
bruno-garcia: We need a Mac and Linux too for standalone. Less relevant until we have native support

.github/workflows/ci.yml
```diff
@@ -252,43 +165,24 @@ jobs:
...
         with:
           name: testapp-ios-project-${{ matrix.unity-version }}
           path: samples/artifacts/builds/iOS/Xcode
+          if-no-files-found: error
 
       - name: Upload release artifacts
         uses: actions/upload-artifact@v2
         # Only on macOS which contain all native plugin needed
-        if: ${{ matrix.os == 'macos-latest' && matrix.unity-version == env.LOWEST_SUPPORTED_UNITY_VERSION }}
+        if: ${{ matrix.unity-version == env.LOWEST_SUPPORTED_UNITY_VERSION }}
```
bruno-garcia, line 185: Comment two lines above is outdated

.github/workflows/ci.yml
```diff
@@ -327,6 +220,73 @@ jobs:
...
+          name: ${{ github.sha }}
+
+      - name: Setup Unity
+        uses: getsentry/setup-unity@46c2e082d98cc3a825a5b59038cb31705fe9ff56
+        with:
+          unity-version: ${{ matrix.unity-version }}
+          unity-version-changeset: ${{ matrix.unity-version-changeset }}
+          unity-modules: windows-il2cpp
+
+      - run: echo "::add-mask::${{ secrets.LICENSE_SERVER_URL }}"
```
bruno-garcia, line 260: Make sure this is also on the Linux step. Anytime we use the license

bruno-garcia (approved):

.github/workflows/ci.yml
vaind: We could do the other (=current) path later in another job later on. In any case I'd do this change in a followup PR.

.github/workflows/ci.yml
vaind: it is (still) there - unchanged code

.github/workflows/ci.yml
vaind: yes, seemed a bit pointless to run on those while they're not officially supported yet. Would just take away from the available github-actions runners' pool. So if you don't mind I'd also add them later when needed.

PR #564, opened 2022-02-15
feat: Support for programmatic option configuration

bitsandfoxes:
The idea is to allow users to programmatically interact with the Sentry options i.e. interact with BeforeSend.
We provide the users with a snippet that lets them create a scriptable object. The path `Assets/Resources/Sentry/` is provided.

ConfigurableOptions

```
using Sentry.Unity;
using UnityEngine;

[CreateAssetMenu(fileName = "Assets/Resources/Sentry/OptionsConfiguration", menuName = "Sentry/OptionsConfiguration", order = 999)]
public class MyOptionsConfiguration : ScriptableOptionsConfiguration
{
    public override void Configure(SentryUnityOptions options)
    {
        // Your code here
    }
}
```

### DIFF
--- a/src/Sentry.Unity/ScriptableSentryUnityOptions.cs
+++ b/src/Sentry.Unity/ScriptableSentryUnityOptions.cs
@@ -44,14 +44,11 @@ namespace Sentry.Unity
         [field: SerializeField] public bool IsEnvironmentUser { get; set; }
 
         [field: SerializeField] public bool EnableOfflineCaching { get; set; }
-
-        // At least 1 item must be allowed in the cache.
-        [field: SerializeField] public int MaxCacheItems { get; set; } = 1;
+        [field: SerializeField] public int MaxCacheItems { get; set; } = 30;
         [field: SerializeField] public int InitCacheFlushTimeout { get; set; }
         [field: SerializeField] public float? SampleRate { get; set; }
         [field: SerializeField] public int ShutdownTimeout { get; set; }
-        // At least 1 item must be allowed in the queue.
-        [field: SerializeField] public int MaxQueueItems { get; set; } = 1;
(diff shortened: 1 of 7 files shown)


src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -44,15 +44,20 @@ public static string GetConfigPath(string? notDefaultConfigName = null)
...
-        [field: SerializeField] public int MaxCacheItems { get; set; }
+
+        // At least 1 item must be allowed in the cache.
+        [field: SerializeField] public int MaxCacheItems { get; set; } = 1;
         [field: SerializeField] public int InitCacheFlushTimeout { get; set; }
         [field: SerializeField] public float? SampleRate { get; set; }
         [field: SerializeField] public int ShutdownTimeout { get; set; }
-        [field: SerializeField] public int MaxQueueItems { get; set; }
+        // At least 1 item must be allowed in the queue.
+        [field: SerializeField] public int MaxQueueItems { get; set; } = 1;
```
bruno-garcia, line 54: This field is here twice?

Default value is 30 btw, in the .NET SDK

bitsandfoxes:
> opening the Options Configuration (o) window revealed this: image Should the user have some message inside of it instead of 'none' on what todo in order to have the options available?

Good catch! I'm working on a follow-up PR that aims to automate the creation of that scriptable object. The idea is to have a button that creates the snippet-script from a template, instantiates the actual scriptable object, and fills the field if it's empty.

bruno-garcia (approved):
what about we use it in the sample?

bitsandfoxes:
> what about we use it in the sample?

Good idea!

PR #568, opened 2022-02-16
feat: Added custom options configuration to samples

bitsandfoxes:
Unfortunately, we ship our sample with just the scenes and scripts. So the user has to add an instance of the scriptable object to their own configuration window by hand.

### DIFF
--- a/samples/unity-of-bugs/Assets/Scripts/CustomOptionsConfiguration.cs
+++ b/samples/unity-of-bugs/Assets/Scripts/CustomOptionsConfiguration.cs
@@ -4,6 +4,7 @@ using UnityEngine;
 [CreateAssetMenu(fileName = "Assets/Resources/Sentry/CustomOptionsConfiguration", menuName = "Sentry/CustomOptionsConfiguration", order = 999)]
 public class CustomOptionsConfiguration : ScriptableOptionsConfiguration
 {
+    // This method gets called when you instantiated the scriptable object and added it to the configuration window
     public override void Configure(SentryUnityOptions options)
     {
         // NOTE: You have complete access to the options object here but changes to the options will not make it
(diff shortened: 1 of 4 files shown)


samples/unity-of-bugs/Assets/Scripts/CustomOptionsConfiguration.cs
```diff
@@ -0,0 +1,23 @@
...
+using UnityEngine;
+
+[CreateAssetMenu(fileName = "Assets/Resources/Sentry/CustomOptionsConfiguration", menuName = "Sentry/CustomOptionsConfiguration", order = 999)]
+public class CustomOptionsConfiguration : ScriptableOptionsConfiguration
+{
+    // This method gets called when you instantiated the scriptable object and added it to the configuration window
+    public override void Configure(SentryUnityOptions options)
+    {
+        // NOTE: You have complete access to the options object here but changes to the options will not make it
+        // to the native layer because the native layer is configured during build time.
```
bruno-garcia, line 11: What about?

```suggestion
        // NOTE: Changes to the options object done here will not affect native crashes. The native SDKs only take 
        // options defined in the Sentry editor configuration window. 
        // Learn more at: https://docs.sentry.io/platforms/unity/native-support/configuration/
```

samples/unity-of-bugs/Assets/Scripts/CustomOptionsConfiguration.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        options.BeforeSend = sentryEvent =>
+        {
+            if (sentryEvent.Tags.ContainsKey("SomeTag"))
+            {
+                return null;
```
bruno-garcia, line 17:
```suggestion
                // Don't send events with a tag SomeTag to Sentry
                return null;
```

bruno-garcia (approved):
LGTM with nits

PR #572, opened 2022-02-17
SmokeTest fixes & improvements

vaind:
#skip-changelog

### DIFF
--- a/samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
+++ b/samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
@@ -100,9 +100,12 @@ public class SmokeTester : MonoBehaviour
             void Verify(HttpRequestMessage message)
             {
                 var msgText = message.Content.ReadAsStringAsync().Result;
-                Debug.Log($"SMOKE TEST: Intercepted HTTP Request #{requests.Count} = {msgText}");
-                requests.Add(msgText);
-                evt.Set();
+                lock (requests)
+                {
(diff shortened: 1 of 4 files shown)


samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
```diff
@@ -75,19 +75,62 @@ public void Start()
...
+                {
+                    Debug.Log($"SMOKE TEST: Quitting due to a failed test case #{testNumber}");
+                    Exit(testNumber);
+                }
+            }
+
+            void ExpectMessage(int index, String substring)
+            {
+                Debug.Log($"SMOKE TEST: Checking if the HTTP request #{index} contains: '{substring}'.");
+                while (requests.Count < index + 1)
```
bruno-garcia, line 123: This will end up throwing, we need to swap the `List` with a `ConcurrentBag`

bruno-garcia (commented):
This is huge improvement

samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
vaind: yes, this wasn't finished yet. I'm now locking manually. `ConcurrentBag` wouldn't do because the checks rely on index-based access for messages. Alternatively, I could use a `SyncrhonizedDictionary` with an atomic increment but don't see much benefit in that, at least not code-wise. And I don't know enough about the available c# synchronization primitives & mechanisms to tell whether it would make any difference.

bruno-garcia (approved):

PR #577, opened 2022-02-22
Windows native config and logging

vaind:
#skip-changelog


bruno-garcia:
Good news?

image

Directory.Build.targets
```diff
@@ -255,6 +255,19 @@ Related: https://forum.unity.com/threads/6572-debugger-agent-unable-to-listen-on
 
     <Error Text="Smoke test failed with code $(SmokeTestExitCode)" Condition="'$(SmokeTestExitCode)' != '200'" />
     <Message Importance="High" Text="Smoke test PASSED" Condition="'$(SmokeTestExitCode)' == '200'" />
+
+    <!-- WIP
```
bruno-garcia, line 259: This getting removed before we merge?

samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
```diff
@@ -65,83 +67,97 @@ public void Start()
...
-            // TODO: Must be set explicitly for the time being.
-            options.RequestBodyCompressionLevel = CompressionLevelWithAuto.Auto;
-            options.DiagnosticLogger = new ConsoleDiagnosticLogger(SentryLevel.Debug);
-            options.CreateHttpClientHandler = () => t;
-
-            var sentryUnityInfo = new SentryUnityInfo();
+        options.Dsn = "http://publickey@localhost:8000/12345";
+        options.Debug = true;
+        options.DebugOnlyInEditor = false;
+        options.DiagnosticLevel = SentryLevel.Debug;
```
bruno-garcia, line 97: Debug is the default
```suggestion
```

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -36,9 +36,6 @@ public static void Configure(SentryUnityOptions options)
 
                 //     return crashedLastRun.Value;
                 // };
-                // At this point Unity has taken the signal handler and will not invoke the original handler (Sentry)
-                // So we register our backend once more to make sure user-defined data is available in the crash report.
-                SentryNativeBridge.ReinstallBackend();
```
bruno-garcia, line 41: This was added to resolve an issue where Unity would take the signal handler after Sentry had taken it and it would not call us after doing its thing. 
Is removing this really intended?

src/Sentry.Unity.Native/SentryNativeBridge.cs
```diff
@@ -29,6 +30,40 @@ public static void Init(SentryUnityOptions options)
                 sentry_options_set_release(cOptions, options.Release);
             }
 
+            if (options.Environment is not null)
```
bruno-garcia, line 33: Was it not possible to add the init code in the main function of the Windows executable instead of initializing here?
This approach is different than Android and iOS and would require further clarification on the docs. i.e:

> In order to provide native crash support, the Sentry SDK for Unity includes platform-specific (that is, Native) SDKs, such as [Android](https://docs.sentry.io/platforms/android/), [Apple](https://docs.sentry.io/platforms/apple/guides/ios/), and [Native](https://docs.sentry.io/platforms/native/).
> 
> By default, we initialize the native SDKs before the Unity engine itself. That means it also runs before the C# layer that relies on [BeforeSceneLoad RuntimeInitializeOnLoadMethodAttribute](https://docs.unity3d.com/ScriptReference/RuntimeInitializeLoadType.BeforeSceneLoad.html) to get initialized. This allows us to capture bugs/crashes of the engine itself, any native plugin, and anything written in C#. The setup for the native SDKs happens during build time, which is why we rely on the options being set in the Sentry editor configuration window and saved to Assets/Resources/Sentry/SentryOptions.asset. To provide a way to modify options programmatically, we've added ScriptableOptionsConfiguration to the Options Config tab in the Sentry editor window.

From: https://docs.sentry.io/platforms/unity/configuration/options/

src/Sentry.Unity.Native/SentryNativeBridge.cs
```diff
@@ -42,6 +77,26 @@ public static void Init(SentryUnityOptions options)
...
+        private static extern void sentry_options_set_environment(IntPtr options, string environment);
+
+        [DllImport("sentry")]
+        private static extern void sentry_options_set_sample_rate(IntPtr options, double rate);
+
+
+        [DllImport("sentry")]
+        private static extern void sentry_options_set_database_path(IntPtr options, string path);
+
+
```
bruno-garcia, line 93:
```suggestion

        [DllImport("sentry")]
        private static extern void sentry_options_set_database_path(IntPtr options, string path);

```

src/Sentry.Unity.Native/SentryNativeBridge.cs
vaind: as discussed, that would be the next PR

Directory.Build.targets
vaind: Didn't manage to get it to work. 
I'll extract to a another branch if you don't want it here.

vaind:
@bruno-garcia @bitsandfoxes please have a/another look. Finally, the crash test passes, hopefully it stays so

Directory.Build.targets
vaind: updated

Directory.Build.targets
```diff
@@ -244,17 +244,11 @@ Related: https://forum.unity.com/threads/6572-debugger-agent-unable-to-listen-on
...
-
-    <Message Condition="'$(StandaloneDataPath)' != ''" Importance="High" Text="Runtime data is found in $(StandaloneDataPath)" />
-    <Exec Condition="$(StandaloneDataPath) != ''" Command="type &quot;$(StandaloneDataPath)/Player.log&quot;"/>
+    <PropertyGroup>
+      <myscript>pwsh "$(RepoRoot)/test/Scripts.Integration.Test/integration-run-smoke-test.ps1" -Smoke -Crash "$(StandaloneExecutablePath)" -AppDataDir "$(StandaloneDataPath)"</myscript>
+    </PropertyGroup>
 
-    <Error Text="Smoke test failed with code $(SmokeTestExitCode)" Condition="'$(SmokeTestExitCode)' != '200'" />
-    <Message Importance="High" Text="Smoke test PASSED" Condition="'$(SmokeTestExitCode)' == '200'" />
+    <Exec Command="$(myscript)" />
```
bitsandfoxes, line 251: Is there a difference between creating `myscript` inside a property group vs exec command?

vaind: just `"` vs `quot;` but I've changed it to the latter (i.e. inlined)

bitsandfoxes (commented):
This looks amazing!

src/Sentry.Unity.Native/SentryNativeBridge.cs
```diff
@@ -42,9 +96,85 @@ public static void Init(SentryUnityOptions options)
...
+        private static Mutex _loggerMutex = new Mutex();
+
+        private static IDiagnosticLogger? _getLogger()
+        {
+            _loggerMutex.WaitOne();
+            var result = _logger;
+            _loggerMutex.ReleaseMutex();
+            return result;
+        }
+        private static void _setLogger(IDiagnosticLogger? logger)
```
bruno-garcia, line 134: C# getter and setter is done through Properties so we can:

```
private static Logger { get; set;}
```
You can declare the body of `get` and `set` if you prefer, but to there's no need to synchronize just for setting/reading

src/Sentry.Unity.Native/SentryNativeBridge.cs
```diff
@@ -42,9 +96,85 @@ public static void Init(SentryUnityOptions options)
...
+            _loggerMutex.ReleaseMutex();
+        }
+
+        // This method is called from the C library
+        [MonoPInvokeCallback(typeof(sentry_logger_function_t))]
+        private static void nativeLog(int cLevel, string message, IntPtr args, IntPtr userData)
+        {
+            var logger = _getLogger();
+            if (logger is null)
+                return;
```
bruno-garcia, line 147: nit: Please always use `{` in blocks
```suggestion
            if (logger is null)
            {
                return;
            }
```

bruno-garcia (commented):
I took a look at the C# files and have a couple notes

samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
```diff
@@ -166,6 +183,49 @@ public static void SmokeTest()
...
     }
 
+    public static void SmokeTestCrash()
+    {
+        Debug.Log("SMOKE TEST: Start");
+
+        InitSentry(new SentryUnityOptions());
+
+        const int sleepMs = 5000;
+        Debug.Log($"SMOKE TEST: Sleep for {sleepMs} ms to avoid failure caused by the background data sync during sentry init.");
```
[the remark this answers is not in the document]
bruno-garcia: Agreed to merge this right now and follow up with figuring out why this is needed. Possibly an SDK bug

samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
```diff
@@ -65,83 +67,98 @@ public void Start()
...
+        }
+        else
+        {
+            Debug.Log($"Unknown command line argument: {arg}");
+            Application.Quit(-1);
+        }
     }
 
-    public static void SmokeTest()
+    public static void InitSentry(SentryUnityOptions options, bool requireNative = true)
```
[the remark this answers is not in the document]
vaind: If you look at the call sites, there are two: one from SmokeTest() which is plain .net and doesn't require native to be configured. The other is from the crash that does require the native part. 
Basically, this was originally meant to be able to run the smoke test (.net) also on Linux/Mac, but that was removed in the end so is actually not necessary per se, but may still come in handy.

samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
```diff
@@ -166,6 +186,49 @@ public static void SmokeTest()
...
 
+    public static void SmokeTestCrash()
+    {
+        Debug.Log("SMOKE TEST: Start");
+
+        InitSentry(new SentryUnityOptions());
+
+        // const int sleepMs = 5000;
+        // Debug.Log($"SMOKE TEST: Sleep for {sleepMs} ms to avoid failure caused by the background data sync during sentry init.");
+        // Thread.Sleep(sleepMs);
```
bruno-garcia, line 197: Consistently working without it now? Can we remove the deadcode?

bruno-garcia (approved):
:shipit:

bruno-garcia:
@vaind looks ready to merge?

bruno-garcia:
Renamed the PR to clarify what user-facing value it adds. Smoke testing was the strategy used to validate/test the feature so removing from the title as it's a tech detail

vaind:
> @vaind looks ready to merge?

Not sure if you've noticed but I've changed the crash test to use a python server instead of a powershell one. Seems to have helped. I guess I can remove the ps1 server script?

bruno-garcia:
> > @vaind looks ready to merge?
> 
> Not sure if you've noticed but I've changed the crash test to use a python server instead of a powershell one. Seems to have helped. I guess I can remove the ps1 server script?
> 
> edit: done so in the latest commit.

I didn't notice. I was hoping we could stick to a single scripting language in the repo and since there's a lot of pwsh that would be the one but if that solved the problem lets go with it. HttpListener is an old bit, we can give it a stab later with a newer approach in pwsh. Or move all the way to python eventually

vaind:
> I was hoping we could stick to a single scripting language

yes, me too. that's why I've tried powershell but doesn't seem to be a very good option. Seems OK to me to have a single python script for this - it's quite simple and limited in scope.

PR #592, opened 2022-02-25
fix: Explicitly copy SentryNativeBridge during build post process

bitsandfoxes:
Due to the `.meta` import settings for the `SentryNativeBridge.m` Unity copied it over to the generated Xcode project whether NativeSupport was enabled/available or not.
We now exclude it on any platform and copy it over during build post-processing.

Stops https://github.com/getsentry/sentry-unity/issues/588 from breaking.

Issue #400, opened 2021-11-02
"Library not loaded: @rpath/Sentry.framework/Sentry, Reason: image not found" if project built on Windows


TigerHix:
### Environment

Sentry Unity SDK 0.6.0, installed with UPM
Unity Editor 2019.4.32f1 Personal

### Steps to Reproduce

1. Build iOS project with Sentry on Windows.
2. Copy project to a Mac and build with Xcode.
3. The app builds successfully; however, when run on a device (not an emulator), the app crashes on launch. Xcode shows this message: `Library not loaded: @rpath/Sentry.framework/Sentry, Reason: image not found`

I then copied the Unity project to the Mac, build it in Unity and Xcode - this time it works flawlessly.

Another plugin that I use, [Lunar Console](https://github.com/SpaceMadness/lunar-unity-console), also had [this issue](https://github.com/SpaceMadness/lunar-unity-console/wiki/Troubleshooting#xcode-interface-builder-could-not-open-the-document-because-it-does-not-exist) (i.e., framework not copied to output project if building on Windows instead of Mac). This is how they [fixed it](https://github.com/SpaceMadness/lunar-unity-console/blob/67f0e241a243024b572dd9d53a4e1d5e9cdab967/Project/Assets/LunarConsole/Editor/BuildPostProcessor.cs#L60). Hope that helps!

TigerHix:
> @TigerHix Are you building an iOS player on Windows? Where do you get this message, in Unity's console? What version of Unity are you using?

I am updating the post now with relevant details :) I accidentally pressed on the submit button right after I typed the title.

TigerHix:
Updated the post.

TigerHix:
I would also mention that although the current Sentry Unity SDK [troublshooting] page mentions this issue, the solution it provides did not work in my case (I already installed Sentry with UPM).

bruno-garcia:
Thanks for all the details, we'll look into it asap

bitsandfoxes:
Thanks for reporting this and also providing a link to the fix @TigerHix. I was able to reproduce this and a PR will be up shortly.
I'm curious, is that your regular dev workflow? You export the Xcode project on Windows and just build on a mac?

TigerHix:
> Thanks for reporting this and also providing a link to the fix @TigerHix. I was able to reproduce this and a PR will be up shortly.
> I'm curious, is that your regular dev workflow? You export the Xcode project on Windows and just build on a mac?

[...]

PR #601, opened 2022-02-28
draft: Resolving PBXProject dependency with reflection

bitsandfoxes:
#skip-changelog

### DIFF
--- a/src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
+++ b/src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
@@ -1,9 +1,8 @@
 using System;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using Sentry.Extensibility;
-using UnityEditor.iOS.Xcode;
-using UnityEditor.iOS.Xcode.Extensions;
 using UnityEngine;
 
 namespace Sentry.Unity.Editor.iOS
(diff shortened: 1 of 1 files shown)


src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -37,8 +58,13 @@ public SentryXcodeProject(string projectRoot) : this(projectRoot, new NativeMain
             INativeOptions sentryNativeOptions)
         {
             _projectRoot = projectRoot;
-            _projectPath = PBXProject.GetPBXProjectPath(projectRoot);
-            _project = new PBXProject();
+
+            var xcodeAssembly = Assembly.Load("UnityEditor.iOS.Extensions.Xcode");
+            var pbxProjectClass = xcodeAssembly.GetType("UnityEditor.iOS.Xcode.PBXProject");
```
bitsandfoxes, line 63: Best practice question: Should I hold that as a member instead of having to call `_project.GetType()` all the time?

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -59,41 +85,68 @@ internal void ReadFromProjectFile()
...
+                .Invoke(_project, new object[] { relativeFrameworkPath, relativeFrameworkPath, 1 }); // 1 is PBXSourceTree.Source
+
+            var addFrameworkToProjectMethod = _project.GetType()
+                .GetMethod("AddFrameworkToProject", BindingFlags.Public | BindingFlags.Instance);
+            addFrameworkToProjectMethod.Invoke(_project, new object[] { _mainTargetGuid, FrameworkName, false });
+            addFrameworkToProjectMethod.Invoke(_project, new object[] { _unityFrameworkTargetGuid, FrameworkName, false });
+
+            // Embedding the framework because it's dynamic and needed at runtime
+            // _project.AddFileToEmbedFrameworks(_mainTargetGuid, frameworkGuid);
+            var xcodeAssembly = Assembly.Load("UnityEditor.iOS.Extensions.Xcode");
```
bitsandfoxes, line 107: The only place where we need the extension of the Xcode extension. Should that loading code move into the constructor?

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
bruno-garcia: Yes, that'd be best

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -17,9 +16,31 @@ internal class SentryXcodeProject : IDisposable
...
+        echo ""Uploading symbols and bcsymbolmaps""
+        process_upload_symbols
+    else
+        echo ""Skipping symbol upload on bitcode enabled and non-install builds""
+    fi
+fi";
 
         private readonly string _projectRoot;
-        private readonly PBXProject _project;
+        private readonly object _project = null!;
```
bruno-garcia, line 43: Sounds like you only need the type:

```suggestion
        private readonly Type _pbxProjectType = null!; // Set in the ctor or throws
```

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -59,41 +85,68 @@ internal void ReadFromProjectFile()
...
+        // Also used for testing
         internal void SetSearchPathBuildProperty(string path)
         {
-            _project.AddBuildProperty(_mainTargetGuid, "FRAMEWORK_SEARCH_PATHS", path);
-            _project.AddBuildProperty(_unityFrameworkTargetGuid, "FRAMEWORK_SEARCH_PATHS", path);
+            // _project.AddBuildProperty(_mainTargetGuid, "FRAMEWORK_SEARCH_PATHS", path);
+            // _project.AddBuildProperty(_unityFrameworkTargetGuid, "FRAMEWORK_SEARCH_PATHS", path);
+            _project.GetType()
+                .GetMethod("AddBuildProperty", new[] { typeof(string[]), typeof(string), typeof(string) })
+                .Invoke(_project, new object[] { new[] { _mainTargetGuid, _unityFrameworkTargetGuid }, "FRAMEWORK_SEARCH_PATHS", path });
```
bruno-garcia, line 149: Does the method we called before have an overload that takes an array of Guids?

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -1,9 +1,8 @@
 using System;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using Sentry.Extensibility;
-using UnityEditor.iOS.Xcode;
-using UnityEditor.iOS.Xcode.Extensions;
 using UnityEngine;
 
 namespace Sentry.Unity.Editor.iOS
```
bruno-garcia, line 8: Worth adding a note like:

// The decision to use reflection here came after many attempts to avoid it
// Allowing customers to export an xcode project outside a Mac while not
// requiring users to install the iOS tools on Windows and Linux became a challenge
// More context: https://github.com/getsentry/sentry-unity/pull/592 and https://github.com/getsentry/sentry-unity/issues/400

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
SimonCropp: should this stuff be done in a static constructor?

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
bitsandfoxes: Yes! I only now realized that due to an `AmbiguousMatchException`.

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -28,7 +57,14 @@ internal class SentryXcodeProject : IDisposable
         private readonly INativeMain _nativeMain;
         private readonly INativeOptions _nativeOptions;
 
-        public SentryXcodeProject(string projectRoot) : this(projectRoot, new NativeMain(), new NativeOptions())
+        static SentryXcodeProject()
+        {
+            var xcodeAssembly = Assembly.Load("UnityEditor.iOS.Extensions.Xcode");
```
bruno-garcia, line 62: I believe this will throw at first access to this type (`SentryXcodeProject`).
In which case we need to handle error here and make sure `Type` isn't null in the instance constructor.

Do we have a test in CI that validates this with and without the iOS build tools? I'd expect a broken build

bitsandfoxes: We run the smoke test on Windows without the iOS module but no one is accessing the type. I've created a follow up issue so we won't hit that issue again: https://github.com/getsentry/sentry-unity/issues/610

bruno-garcia (approved):
Change looks good but I'm afraid of the fact no tests failed, and no tests were added. I expect the call site to require some exception handling when instantiating `SentryXcodeProject` to deal with the reflection call failing when the iOS Tools for Unity are not installed.

fahall:
Can we get a new release with this fix included? v0.11.0 does not appear to include it.

bitsandfoxes:
It's on its way!

PR #607, opened 2022-03-02
Feat: windows symbol upload

vaind:
Part of windows native support #527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@
 
 ### Features
 
-- Windows build-time debug symbol upload ([#607](https://github.com/getsentry/sentry-unity/pull/607))
+- Windows native support (64-bit)
+  - native crash handler ([#380](https://github.com/getsentry/sentry-unity/pull/380))
+  - configuration & log forwarding ([#577](https://github.com/getsentry/sentry-unity/pull/577))
+  - scope synchronization ([#546](https://github.com/getsentry/sentry-unity/pull/546))
+  - symbol upload while building through Unity ([#607](https://github.com/getsentry/sentry-unity/pull/607))
 
 ## 0.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - iOS builds no longer break when native support disabled or not available ([#592](https://github.com/getsentry/sentry-unity/pull/592))
 
+### Features
+
+- Windows build-time debug symbol upload ([#607](https://github.com/getsentry/sentry-unity/pull/607))
+
 ## 0.11.0
 
 ### Features
--- a/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
+++ b/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
@@ -70,8 +70,6 @@ namespace Sentry.Unity.Editor.Native
                 return;
             }
 
-            var propertiesFile = SentryCli.CreateSentryProperties(projectDir, cliOptions);
-
             logger.LogInfo("Uploading debugging information using sentry-cli in {0}", projectDir);
 
             var paths = "";
@@ -113,7 +111,10 @@ namespace Sentry.Unity.Editor.Native
                     CreateNoWindow = true
                 }
             };
-            process.StartInfo.EnvironmentVariables["SENTRY_PROPERTIES"] = propertiesFile;
+
+            process.StartInfo.EnvironmentVariables["SENTRY_ORG"] = cliOptions.Organization;
+            process.StartInfo.EnvironmentVariables["SENTRY_PROJECT"] = cliOptions.Project;
+            process.StartInfo.EnvironmentVariables["SENTRY_AUTH_TOKEN"] = cliOptions.Auth;
             process.OutputDataReceived += (sender, args) => logger.LogDebug($"sentry-cli: {args.Data.ToString()}");
             process.ErrorDataReceived += (sender, args) => logger.LogError($"sentry-cli: {args.Data.ToString()}");
             process.Start();
--- a/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
+++ b/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
@@ -42,8 +42,8 @@ namespace Sentry.Unity.Editor.Native
                 }
 
                 var projectDir = Path.GetDirectoryName(executablePath);
-                addCrashHandler(logger, projectDir);
-                uploadDebugSymbols(logger, projectDir, Path.GetFileName(executablePath));
+                AddCrashHandler(logger, projectDir);
+                UploadDebugSymbols(logger, projectDir, Path.GetFileName(executablePath));
 
             }
             catch (Exception e)
@@ -53,24 +53,26 @@ namespace Sentry.Unity.Editor.Native
             }
         }
 
-        private static void addCrashHandler(IDiagnosticLogger logger, string projectDir)
+        private static void AddCrashHandler(IDiagnosticLogger logger, string projectDir)
         {
             var crashpadPath = Path.GetFullPath(Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins",
                 "Windows", "Sentry", "crashpad_handler.exe"));
             var targetPath = Path.Combine(projectDir, Path.GetFileName(crashpadPath));
-            logger.LogInfo($"Copying the native crash handler '{Path.GetFileName(crashpadPath)}' to the output directory");
+            logger.LogInfo("Copying the native crash handler '{0}' to the output directory", Path.GetFileName(crashpadPath));
             File.Copy(crashpadPath, targetPath, true);
         }
 
-        private static void uploadDebugSymbols(IDiagnosticLogger logger, string projectDir, string executableName)
+        private static void UploadDebugSymbols(IDiagnosticLogger logger, string projectDir, string executableName)
         {
             var cliOptions = SentryCliOptions.LoadCliOptions();
             if (!cliOptions.Validate(logger))
+            {
                 return;
+            }
 
             var propertiesFile = SentryCli.CreateSentryProperties(projectDir, cliOptions);
 
-            logger.LogInfo($"Uploading debuging information using sentry-cli in {projectDir}");
+            logger.LogInfo("Uploading debugging information using sentry-cli in {0}", projectDir);
 
             var paths = "";
             Func<string, bool> addPath = (string name) =>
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - iOS builds no longer break when native support disabled or not available ([#592](https://github.com/getsentry/sentry-unity/pull/592))
 
+### Features
+
+- Windows build-time debug symbol upload ([#607](https://github.com/getsentry/sentry-unity/pull/607))
```
vaind, line 11: actually, I see we haven't had a generic windows-support changelog entry so maybe we should do that instead

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
+            addPath(executableName);
+            addPath("GameAssembly.dll");
+            addPath("UnityPlayer.dll");
+            addPath(Path.GetFileNameWithoutExtension(executableName) + "_BackUpThisFolder_ButDontShipItWithYourGame");
+
+            // no debug info in sentry.dll but sentry.io will complain about a missing file...
+            // addPath(Path.GetFileNameWithoutExtension(executableName) + "_Data/Plugins/x86_64/sentry.dll");
+
+            // Configure the process using the StartInfo properties.
+            var process = new Process
```
bitsandfoxes, line 101: Ideally, we would like to avoid spawning a blocking process from within the Unity editor and call Sentry CLI from a msbuild target during the build.

vaind: While @bitsandfoxes is working on the integration into the unity build process for standalone windows, I'd say this PR still makes sense to merge as is now? Already brings support for the "build-in-unity" workflow

CHANGELOG.md
vaind: done

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
             }
             catch (Exception e)
             {
-                logger.LogError("Failed to add the Sentry crash handler to the built application", e);
+                logger.LogError("Failed to add the Sentry native integration to the built application", e);
                 throw new BuildFailedException("Sentry Native BuildPostProcess failed");
             }
         }
+
+        private static void addCrashHandler(IDiagnosticLogger logger, string projectDir)
```
bruno-garcia, line 56:
```suggestion
        private static void AddCrashHandler(IDiagnosticLogger logger, string projectDir)
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
+        {
+            var crashpadPath = Path.GetFullPath(Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins",
+                "Windows", "Sentry", "crashpad_handler.exe"));
+            var targetPath = Path.Combine(projectDir, Path.GetFileName(crashpadPath));
+            logger.LogInfo($"Copying the native crash handler '{Path.GetFileName(crashpadPath)}' to the output directory");
+            File.Copy(crashpadPath, targetPath, true);
+        }
+
+        private static void uploadDebugSymbols(IDiagnosticLogger logger, string projectDir, string executableName)
```
bruno-garcia, line 65:
```suggestion
        private static void UploadDebugSymbols(IDiagnosticLogger logger, string projectDir, string executableName)
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
                     return;
                 }
 
-                var crashpadPath = Path.GetFullPath(Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins",
-                    "Windows", "Sentry", "crashpad_handler.exe"));
-                var targetPath = Path.Combine(Path.GetDirectoryName(pathToProject), Path.GetFileName(crashpadPath));
-                File.Copy(crashpadPath, targetPath, true);
+                var projectDir = Path.GetDirectoryName(executablePath);
+                addCrashHandler(logger, projectDir);
+                uploadDebugSymbols(logger, projectDir, Path.GetFileName(executablePath));
```
bruno-garcia, line 46:
```suggestion
                AddCrashHandler(logger, projectDir);
                UploadDebugSymbols(logger, projectDir, Path.GetFileName(executablePath));
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
                 throw new BuildFailedException("Sentry Native BuildPostProcess failed");
             }
         }
+
+        private static void addCrashHandler(IDiagnosticLogger logger, string projectDir)
+        {
+            var crashpadPath = Path.GetFullPath(Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins",
+                "Windows", "Sentry", "crashpad_handler.exe"));
+            var targetPath = Path.Combine(projectDir, Path.GetFileName(crashpadPath));
+            logger.LogInfo($"Copying the native crash handler '{Path.GetFileName(crashpadPath)}' to the output directory");
```
bruno-garcia, line 61: The logger takes a positional format (standard in C#) and avoids allocations if logging is disabled

```suggestion
            logger.LogInfo("Copying the native crash handler '{0}' to the output directory", Path.GetFileName(crashpadPath));
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
+            File.Copy(crashpadPath, targetPath, true);
+        }
+
+        private static void uploadDebugSymbols(IDiagnosticLogger logger, string projectDir, string executableName)
+        {
+            var cliOptions = SentryCliOptions.LoadCliOptions();
+            if (!cliOptions.Validate(logger))
+                return;
```
bruno-garcia, line 69:
```suggestion
            if (!cliOptions.Validate(logger))
            {
                return;
            }
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
+
+            var propertiesFile = SentryCli.CreateSentryProperties(projectDir, cliOptions);
+
+            logger.LogInfo($"Uploading debuging information using sentry-cli in {projectDir}");
```
bruno-garcia, line 73:
```suggestion
            logger.LogInfo("Uploading debugging information using sentry-cli in {0}", projectDir);
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
+                    Arguments = "upload-dif " + paths,
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true,
+                    RedirectStandardError = true,
+                    CreateNoWindow = true
+                }
+            };
+            process.StartInfo.EnvironmentVariables["SENTRY_PROPERTIES"] = propertiesFile;
+            process.OutputDataReceived += (sender, args) => logger.LogDebug($"sentry-cli: {args.Data.ToString()}");
+            process.ErrorDataReceived += (sender, args) => logger.LogWarning($"sentry-cli: {args.Data.ToString()}");
```
bruno-garcia, line 116: Why not level error?

```suggestion
            process.ErrorDataReceived += (sender, args) => logger.LogError($"sentry-cli: {args.Data.ToString()}");
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
+                    return true;
+                }
+                else
+                {
+                    logger.LogWarning($"Coudn't find '{name}' - debug symbol upload will be incomplete");
+                    return false;
+                }
+            };
+
+            addPath(executableName);
```
bruno-garcia, line 92: Aren't we better off running sentry-cli on the root and let it find the relevant files? It'll also be more future proof if Unity introduces other native shared libraries

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -40,43 +41,83 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToProject)
...
+                    FileName = SentryCli.SetupSentryCli(),
+                    WorkingDirectory = projectDir,
+                    Arguments = "upload-dif " + paths,
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true,
+                    RedirectStandardError = true,
+                    CreateNoWindow = true
+                }
+            };
+            process.StartInfo.EnvironmentVariables["SENTRY_PROPERTIES"] = propertiesFile;
```
bruno-garcia, line 114: Why not pass the individual values as variables instead of creating a file?
If sentry-cli is never spawned outside of this C# call, we don't need the dangling file

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: I've thought about that but didn't like that it would also upload:
* UnityCrashHandler64.exe
* crashpad_handler.exe

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: true... I've just followed the other bindings but you're right we don't need the file

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
bitsandfoxes: Do you mean it uploads the executables? My understanding was that `upload-dif` would only scan for and upload debug information files.

vaind: executables also have some debug info. See for example here:

image

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: BTW, this is changed now for a couple of days

bitsandfoxes (approved):

PR #608, opened 2022-03-03
Close sentry instance when quitting the app

### LINKED ISSUE #543: Call native `sentry_close()` when the app quits
The native API docs say the following: 
> Calling sentry_close() before exiting the application is critical. It ensures that events can be sent to Sentry before execution stops. Otherwise, event data may be lost.

See:
* https://docs.sentry.io/platforms/native/#configure
* https://docs.sentry.io/platforms/native/configuration/draining/

Calling it inside `OnApplicationQuit()` seems like a good candidate ([unity docs](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnApplicationQuit.html))

vaind:
closes #543 

Native integrations
* [x] standalone native (windows)
* [x] iOS
* [x] Android

### DIFF
--- a/src/Sentry.Unity.Android/SentryNativeAndroid.cs
+++ b/src/Sentry.Unity.Android/SentryNativeAndroid.cs
@@ -47,10 +47,9 @@ namespace Sentry.Unity.Android
                 }
                 ApplicationAdapter.Instance.Quitting += () =>
                 {
+                    // Sentry Native is initialized and closed by the Java SDK, no need to call into it directly
                     options.DiagnosticLogger?.LogDebug("Closing the sentry-java SDK");
                     SentryJava.Close();
-                    options.DiagnosticLogger?.LogDebug("Closing the sentry-native SDK");
-                    SentryNative.Close();
                 };
             }
         }
--- a/src/Sentry.Unity.Android/SentryNativeAndroid.cs
+++ b/src/Sentry.Unity.Android/SentryNativeAndroid.cs
@@ -1,4 +1,5 @@
 using Sentry.Extensibility;
+using Sentry.Unity.Integrations;
 
 namespace Sentry.Unity.Android
 {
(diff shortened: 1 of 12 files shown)


bitsandfoxes:
The `SentryMonoBehaviour` was added to give us access to callbacks on the `MonoBehaviour` and things from inside Unity itself. It's an actual game object that receives the OnApplicationQuit inherently to destroy itself.

In your case, I'd go with the `IApplication` and use the `OnQuitting` there.

src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -44,6 +45,13 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
                     // So we register our backend once more to make sure user-defined data is available in the crash report.
                     SentryNative.ReinstallBackend();
                 }
+                ApplicationAdapter.Instance.Quitting += () =>
+                {
+                    options.DiagnosticLogger?.LogDebug("Closing the sentry-java SDK");
+                    SentryJava.Close();
+                    options.DiagnosticLogger?.LogDebug("Closing the sentry-native SDK");
+                    SentryNative.Close();
```
bruno-garcia, line 53: I would expect sentry-android to close sentry-native:

https://github.com/getsentry/sentry-java/blob/530d17dec68a4689d4455941a11f2445d8d10116/sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java#L41-L43

Should we close directly here too? 
What happens if we close it twice?

bruno-garcia (commented):
One questions but other than that. lgtm

src/Sentry.Unity.Android/SentryNativeAndroid.cs
bruno-garcia, line 53:
```suggestion
                    // Sentry Native is initialized and closed by the Java SDK, no need to call into it directly
                    options.DiagnosticLogger?.LogDebug("Closing the sentry-java SDK");
                    SentryJava.Close();
```

vaind: shouldn't be an issue from what I can tell but removing the native close

bruno-garcia (approved):
The only risk I see is that an error/crash happens after the Unity's AppShutdown thing ran, and we'd be all shut down and not capture. For that reason, recently we've changed the .NET SDK to have the `Dispose` handle returned by the `Init` call behave like a Flush only since that's what the semantics where in the code snippets (block to capture and flush). So the SDK would continue to capture even after that block.
That said, .NET SDK's `Close` continues to completely close the SDK since it swaps to a NoOp hub instance (the old one would be GC'ed):

https://github.com/getsentry/sentry-dotnet/blob/2d6c847268c9422c0a02bffe87dd656736cf197c/src/Sentry/SentrySdk.cs#L123-L125

PR #611, opened 2022-03-03
CI: wait for artifacts to speed up builds

vaind:


.github/workflows/ci.yml
```diff
@@ -82,28 +76,32 @@ jobs:
...
 
-      - uses: actions/download-artifact@v2
+      - uses: vaind/download-artifact@8bd9f2d7956f47bf3969323f3b2bc1b80987ef79
         with:
           name: iOS-sdk
           path: package-dev/Plugins/iOS
+          waitTimeout: 3600
 
-      - uses: actions/download-artifact@v2
+      - uses: vaind/download-artifact@8bd9f2d7956f47bf3969323f3b2bc1b80987ef79
```
bruno-garcia, line 100: Do you plan to send these changes upstream to get back to using the official download action?
If not, we're likely going to need this fork under `getsentry` org

bruno-garcia (approved):

.github/workflows/ci.yml
vaind: yes, [I've already created one](https://github.com/actions/download-artifact/pull/139). As for if/when it's going to be merged... I can't tell

vaind: On the other hand, how is using a fork in my repo and a specific git tag any different than doing that for other third-party actions, like this one? https://github.com/getsentry/sentry-unity/blob/701ca89b729b613ace2597498b1f6d420577ac54/.github/workflows/ci.yml#L27

.github/workflows/ci.yml
```diff
@@ -82,28 +76,32 @@ jobs:
             temp/unity-packages/Library/ScriptAssemblies/*.TestRunner.*
           key: samples/unity-of-bugs|${{ steps.env.outputs.unityVersion }}-${{ hashFiles('samples/unity-of-bugs/Packages/packages-lock.json') }}
 
-      - uses: actions/download-artifact@v2
+      - name: Start the Unity the docker container
```
bruno-garcia, line 79:
```suggestion
      - name: Start the Unity docker container
```

.github/workflows/ci.yml
```diff
@@ -195,6 +181,18 @@ jobs:
             package-release.zip
             modules/sentry-java/sentry-android-ndk/build/intermediates/merged_native_libs/release/out/lib/*
 
+      # TODO: can we make unity only prepare the project, without compiling it, to unlock the unity license faster? iOS is already built like that.
```
bruno-garcia, line 184: We can. We'll need to "export gradle project" on Android instead of building it. It makes sense to go that route

bruno-garcia (approved):

.github/workflows/ci.yml
bruno-garcia: It's not. But given the fact this fork is being done specifically for this project, if we plan on maintaining, it's best it becomes part of the getsentry org as it's possible it'll serve this project longer than you and I

.github/workflows/ci.yml
vaind: unrelated change but OK, applying

.github/workflows/ci.yml
vaind: not a new todo, just moved a couple of lines. 

And I'm aware it should work but didn't get around to do it

PR #614, opened 2022-03-07
feat: Capture Native Instruction Addrs for Exceptions

Swatinem:
This is very much WIP for now and does need more work on:

* Add the `debug_meta.images` to the Protocol definitions (we have a single `GameAssembly`).
* Add the `addr_mode` to Protocol definitions (il2cpp instruction addrs are all relative to the main `GameAssembly`).
* Make sure the native functions are safe to call and backwards/forwards compatible. (We are calling into `libil2cpp` internals which might change any day, and might not be available on older versions anyways)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Adding override for Sentry CLI URL ([#666](https://github.com/getsentry/sentry-unity/pull/666))
+
 ### Fixes
 
 - Whitespaces no longer cause issues when uploading symbols for Windows native  ([#655](https://github.com/getsentry/sentry-unity/pull/655))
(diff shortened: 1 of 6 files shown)


src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -273,6 +275,14 @@ private void PopulateTags(SentryEvent @event)
         }
     }
 
+
+    internal class NativeStackTrace
+    {
+        public IntPtr[] Frames = null!;
```
bruno-garcia, line 281: Doesn't look safe to `null!` here. Start off with an instance, or make this nullable. 
```suggestion
        public IntPtr[] Frames = new IntPtr[0];
```

```suggestion
        public IntPtr[]? Frames { get; set; }
```

PR #622, opened 2022-03-09
feat: Building playbackengine

bitsandfoxes:
To note:

- the whole functionality is behind an opt-in feature flag `AddSentryToWindowsPlayer` on the editor options for now
- renaming SentryCliOptions to EditorOptions is a breaking change
- the actual project manipulation for the player is still missing

#skip-changelog


src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+using Sentry.Extensibility;
+using UnityEditor;
+using UnityEngine;
+using Debug = UnityEngine.Debug;
+
+namespace Sentry.Unity.Editor.Native
+{
+    internal static class SentryWindowsPlayer
+    {
+        internal static readonly ProcessStartInfo StartInfo = new()
```
bruno-garcia, line 13: It's best to allocate this when needed and let GC collect it. Instantiating this is cheap: https://github.com/dotnet/runtime/blob/78e840e1072127d68714b55cffdcfa96a7e2f8af/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/ProcessStartInfo.cs#L20-L36

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Diagnostics;
+using System.IO;
+using Sentry.Extensibility;
+using UnityEditor;
+using UnityEngine;
+using Debug = UnityEngine.Debug;
+
+namespace Sentry.Unity.Editor.Native
+{
+    internal static class SentryWindowsPlayer
```
bruno-garcia, line 11: This class has a lot of methods taking `IDiagnosticLogger`. It seems to me this should be an instance class instead of static.

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+            using var props = File.CreateText(Path.Combine(windowsPlayerTarget, "UnityCommon.props"));
+            props.Write(@"<?xml version=""1.0"" encoding=""utf-8""?>
+<Project xmlns=""http://schemas.microsoft.com/developer/msbuild/2003"">
+</Project>");
+        }
+
+        internal static string LocateMSBuild(string vsWherePath, IDiagnosticLogger? logger)
+        {
+            StartInfo.FileName = vsWherePath;
+            StartInfo.Arguments = "-latest -requires Microsoft.Component.MSBuild -find MSBuild\\**\\Bin\\MSBuild.exe";
```
bruno-garcia, line 56: How do we control what version we'll be working with here? If the user has multiple installed, some could be compatible and others not.

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+            var process = new Process { StartInfo = StartInfo };
+            process.OutputDataReceived += (sender, args) => vsWhereOutput += args.Data;
+            process.Start();
+            process.BeginOutputReadLine();
+            process.BeginErrorReadLine();
+            process.WaitForExit();
+
+            logger?.LogDebug("VSWhere returned with: {0}", vsWhereOutput);
+
+            if (!File.Exists(vsWhereOutput))
```
bruno-garcia, line 68: Is this a sanity check "just in case"? Or can vswhere really return a path that doesn't exist?

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+                throw new FileNotFoundException($"Failed to locate 'msbuild'. VSWhere returned {vsWhereOutput}");
+            }
+
+            return vsWhereOutput;
+        }
+
+        internal static string LocateVSWhere(IDiagnosticLogger? logger)
+        {
+            var projectPath = Path.GetDirectoryName(Application.dataPath);
+            var directories = Directory.GetDirectories(Path.Combine(projectPath, "Library", "PackageCache"), "com.unity.ide.visualstudio@*");
```
bruno-garcia, line 79: if I use Rider on Windows, can I uninstall this package? We're throwing below so I wonder if we're locking ourselves into a place where we'll start breaking ppl with a slightly different setup than we built for

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+            if (directories is null || directories.Length < 1)
+            {
+                throw new Exception("Failed lo locate the 'com.unity.ide.visualstudio' package.");
```
bruno-garcia, line 82: Some other type other than the base is preferred:

```suggestion
                throw new InvalidOperationException("Failed lo locate the 'com.unity.ide.visualstudio' package.");
```

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            // TODO: Not sure if there can be more than one version of a package in the Library/PackageCache and if it even matters
+            var vsPackagePath = directories[0];
```
bruno-garcia, line 86: Sounds like this could be a reason why things don't work and proper logging would help. So iterating over the list and debug logging could be a good idea.

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+            var vsWherePath = Path.Combine(vsPackagePath, "Editor", "VSWhere", "vswhere.exe");
+            if (!File.Exists(vsWherePath))
+            {
+                throw new FileNotFoundException($"Failed to find 'vswhere.exe' at '{vsWherePath}'");
+            }
+
+            return vsWherePath;
+        }
+
+        public static void Build(SentryUnityOptions options, string executablePath)
```
bruno-garcia, line 98: I suggest having this method before any of the internal ones above. It's easier to dig into such code when you get straight into the entry point instead of reading multiple small/disconnected internal (for testing I assume?) methods

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            var playerSource = LocateWindowsPlayerSource();
+            var playerTarget = FileUtil.GetUniqueTempPathInProject();
+
+            CreateWindowsPlayerProject(playerSource, playerTarget, options.DiagnosticLogger);
+
+            AddNativeOptions(options);
+            AddSentryToMain(options);
+
+            StartInfo.FileName = msBuildPath;
```
bruno-garcia, line 111: We're not checking if the value was set before and skipping the vswhere call or any other optimization so modifying this static property here isn't giving us much and has side effects beyond this method invocation. Lets avoid it

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+            var errorData = "";
+            var process = new Process { StartInfo = StartInfo };
+            process.OutputDataReceived += (sender, args) => outputData += args.Data;
+            process.ErrorDataReceived += (sender, args) => errorData += args.Data;
+            process.Start();
+            process.BeginOutputReadLine();
+            process.BeginErrorReadLine();
+            process.WaitForExit();
+
+            if (outputData.Contains("Build succeeded"))
```
bruno-garcia, line 124: If the machine is set to another language, is this going to break?

src/Sentry.Unity.Editor/Native/SentryWindowsPlayer.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                options.DiagnosticLogger?.LogDebug("Succeeded building the PlaybackEngine");
+            }
+            else if (outputData.Contains("Build failed"))
```
bruno-garcia, line 128: Can we rely on the exit code of the program instead of these strings?

bruno-garcia (commented):
I'm very uneasy about the approach of finding msbuild and recompiling that project. I know I got convinced on our last call but I believe we need the trade offs (and risks) written down somewhere and involve other folks before we go ahead. I worry about side effects beyond building this one project (e.g: What if I have 2 unity projects open?)

src/Sentry.Unity.Editor/Native/MSBuildLocator.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            if (!File.Exists(editorOptions.VSWherePath))
+            {
+                logger?.LogDebug("Failed to find 'VSWhere' at '{0}'. Trying to locate.", editorOptions.VSWherePath);
+                SetVSWherePath(editorOptions, logger);
+            }
+
+            logger?.LogDebug("Using 'VSWhere' at '{0}' to locate MSBuild.", editorOptions.VSWherePath);
+
+            var vsWhereOutput = "";
```
bruno-garcia, line 22:
```suggestion
            var vsWhereOutput = new StringBuilder();
```

src/Sentry.Unity.Editor/Native/MSBuildLocator.cs
```diff
@@ -0,0 +1,93 @@
...
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true,
+                    RedirectStandardError = true,
+                    CreateNoWindow = true
+                }
+            };
+
+            process.OutputDataReceived += (sender, args) => vsWhereOutput += args.Data;
+            process.Start();
+            process.BeginOutputReadLine();
```
bruno-garcia, line 38: I haven't used this in years but I wonder: Can we start listening before we start the program? I wonder if we could miss out on its output due to a race here

bruno-garcia (commented):
🚀

bruno-garcia:
We're taking the inverse route where we init native from C# on mobile now. So it's unlikely we'll get to this anytime soon. So closing.

PR #624, opened 2022-03-10
fix: Move SentryUnityInfo into SentryInitialization

bitsandfoxes:
As discussed in the call: Sentry should try to keep its internals to itself.

#skip-changelog


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -45,4 +45,17 @@ public static void Init()
...
+       public bool IL2CPP
+       {
+           get =>
+   #if ENABLE_IL2CPP
+               true;
+   #else
+               false;
+   #endif
+       }
+   }
```
bruno-garcia, line 60:
```suggestion
    {
        public bool IL2CPP
        {
            get =>
#if ENABLE_IL2CPP
               true;
#else
               false;
#endif
        }
    }
```

bruno-garcia (approved):

PR #640, opened 2022-03-16
Addressing Infinite Loop issue 

Hoodad:
It appears to be unncessary to call the AssetDatabase.Import on the XML file as its only used when building the end binary. By removing this call to Import the asset, it hopefully address the infinite loop issue that occurs on a clean project when first loading SentryWindow.cs. However, I say hopefully as I cannot verify that this fix address the infinite loop import as I don't know how to build the managed dlls for the SentryWindow.cs.


bitsandfoxes (approved):
Hey @Hoodad! Sorry for the very delayed response. And thanks for contributing! LGTM!

PR #630, opened 2022-03-10
Test/mobile crash

vaind:
closes #625 

TODOs:
* [x] add to iOS tests
* [x] add to Android tests
* [x] fix android manifest (connecting using HTTP): https://stackoverflow.com/questions/45940861/android-8-cleartext-http-traffic-not-permitted
* ~rework CI so that both iOS & Android use the same process to create a project as windows does (i.e. create an empty project, build, add unity, build again, only test afterwards)~
* [x] or do a workaround - just update the sentry URL in the assets.

#skip-changelog

Issue #264, opened 2021-07-19
Auto Session Tracking should only be enabled by default on Mobile until total process isolation is supported


bruno-garcia:
The session support currently expects exclusive access to the directory it caches data. This works well in sandboxed apps like on mobile but not on Desktop when multiple instances can be executed concurrently. 

For this reason until  https://github.com/getsentry/sentry-dotnet/issues/1067 is fixe, we can only opt-in to auto session tracking safely if we are sure there's only 1 instance of the app (mobile, or a player on desktop that only allows one instance running).

semuserable:
`PlayerSettings.forceSingleInstance` is present in `UnityEditor.dll` and **not** available in `Sentry.Unity` (references `UnityEngine`) where logic for this is executing (`SentryOptionsUtility.SetDefaults`).

There are 2 possible approaches to determine the platform: [Application.platform](http://docs.unity3d.com/ScriptReference/Application-platform.html) or [Platform dependent compilation](https://docs.unity3d.com/Manual/PlatformDependentCompilation.html).

Suggested way is to use `Platform dependent compilation` approach. But the problem is still in `PlayerSettings.forceSingleInstance` not available at the place of calling.

Possible implementation could look something like this

```csharp
scriptableOptions.AutoSessionTracking = IsAutoSessionTracking(options.AutoSessionTracking);

private bool IsAutoSessionTracking(bool autoSessionTracking)
{
#if UNITY_IOS || UNITY_ANDROID || UNITY_PS4 || UNITY_XBOXONE
    return autoSessionTracking;
#elif UNITY_STANDALONE || UNITY_EDITOR
    return PlayerSettings.forceSingleInstance;
#else
    return autoSessionTracking;
#endif
}
```

bruno-garcia:
`SentryWindow` runs in the editor (`Sentry.Unity.Editor`), we could read `PlayerSettings.forceSingleInstance` there and set it to the scriptable object and rely on that value at runtime to disable auto session tracking if the `Application.platform` is standalone and `forceSingleInstance` is not true.

Or we live with this until we address getsentry/sentry-dotnet#1067

bitsandfoxes:
The `SentryWindow` would not notice a change to the player settings. But we could read them during pre-build and set it accordingly?

vaind:
As per a discord discussion:
* this is rather important for the Windows standalone support so bumping to P1
[...]

PR #643, opened 2022-03-17
Don't run sentry-dotnet with multiple app instances at the same time (desktop)

vaind:
Workaround for #264 

* [x] depends on #630 which changes smoke test so that it doesn't call init twice

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -29,7 +29,7 @@ namespace Sentry.Unity
     public static class SentryInitialization
     {
 #if SENTRY_NATIVE_WINDOWS
-        private static FileStream _globalMutex;
+        private static FileStream _lockFile;
 #endif
 
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
@@ -54,14 +54,15 @@ namespace Sentry.Unity
                 {
                     try
                     {
-                        _globalMutex = new FileStream(Path.Combine(options.CacheDirectoryPath, "sentry-unity.lock"), FileMode.OpenOrCreate,
+                        _lockFile = new FileStream(Path.Combine(options.CacheDirectoryPath, "sentry-unity.lock"), FileMode.OpenOrCreate,
                                 FileAccess.ReadWrite, FileShare.None);
 
                         Application.quitting += () =>
                         {
                             try
                             {
-                                _globalMutex.Close();
+                                // We don't really need to close, Windows would do that anyway, but let's be nice.
+                                _lockFile.Close();
                             }
                             catch (Exception ex)
                             {
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -17,6 +17,9 @@ using Sentry.Unity.iOS;
 using Sentry.Unity.Android;
 #elif SENTRY_NATIVE_WINDOWS
 using Sentry.Unity.Native;
+using Sentry.Extensibility;
+using System;
+using System.IO;
 #endif
 
 [assembly: AlwaysLinkAssembly]
@@ -25,6 +28,10 @@ namespace Sentry.Unity
 {
     public static class SentryInitialization
     {
+#if SENTRY_NATIVE_WINDOWS
+        private static FileStream _globalMutex;
+#endif
+
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void Init()
         {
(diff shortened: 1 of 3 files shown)


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -25,6 +28,10 @@ namespace Sentry.Unity
 {
     public static class SentryInitialization
     {
+#if SENTRY_NATIVE_WINDOWS
```
bruno-garcia, line 31: I wonder if it's best to move this into the package and just pass the "windows native" flag in. Like we did il2cpp. Goal is to keep this entry point free of logic and be able to test stuff in the test project (without the need of running this inot

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -39,6 +46,38 @@ public static void Init()
...
+                    try
+                    {
+                        _globalMutex = new FileStream(Path.Combine(options.CacheDirectoryPath, "sentry-unity.lock"), FileMode.OpenOrCreate,
+                                FileAccess.ReadWrite, FileShare.None);
+
+                        Application.quitting += () =>
+                        {
+                            try
+                            {
+                                _globalMutex.Close();
```
bruno-garcia, line 64: In the event of a crash (or any reason this event doesn't fire) will it ever recover?

vaind: added a comment: "We don't really need to close, Windows would release the lock anyway, but let's be nice."

vaind: I've tested that manually

package-dev/Runtime/SentryInitialization.cs
vaind: moved... should I also try to add a test case even though this is temporary until dotnet is fixed?

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -24,19 +28,54 @@ public static void Init(Action<SentryUnityOptions> sentryUnityOptionsConfigure)
...
+                if (ApplicationAdapter.Instance.Platform == RuntimePlatform.WindowsPlayer && options.CacheDirectoryPath != null)
+                {
+                    try
+                    {
+                        _lockFile = new FileStream(Path.Combine(options.CacheDirectoryPath, "sentry-unity.lock"), FileMode.OpenOrCreate,
+                                FileAccess.ReadWrite, FileShare.None);
+                    }
+                    catch (Exception ex)
+                    {
+                        options.DiagnosticLogger?.Log(SentryLevel.Warning, "An exception was thrown while trying to " +
```
bruno-garcia, line 50: Can we do that instead?

```suggestion
                        options.DiagnosticLogger?.LogWarning("An exception was thrown while trying to " +
```

bruno-garcia (approved):
Nice! Thank you @vaind

package-dev/Runtime/SentryInitialization.cs
bruno-garcia: yeah, at least now we know Mutex wouldn't work either if we're trying to add to .NET

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -24,19 +28,54 @@ public static void Init(Action<SentryUnityOptions> sentryUnityOptionsConfigure)
...
+                            "acquire a lockfile on the config directory: .NET event cache will be disabled.", ex);
+                        options.CacheDirectoryPath = null;
+                        options.AutoSessionTracking = false;
+                    }
+                }
+
+                options.DiagnosticLogger?.LogDebug(options.ToString());
```
bruno-garcia, line 57:
```suggestion
```

bruno-garcia: Just noticed this ^ should be removed?

src/Sentry.Unity/SentryUnity.cs
vaind: couldn't in the public code. It should work now.

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -24,19 +28,53 @@ public static void Init(Action<SentryUnityOptions> sentryUnityOptionsConfigure)
...
-            if (sentryUnityOptions.ShouldInitializeSdk())
+            SentryOptionsUtility.TryAttachLogger(options);
+            if (options.ShouldInitializeSdk())
             {
-                sentryUnityOptions.DiagnosticLogger?.LogDebug(sentryUnityOptions.ToString());
-                var sentryDotNet = SentrySdk.Init(sentryUnityOptions);
+                // On Standalone, we disable cache dir in case multiple app instances run over the same path.
+                // Note: we cannot use a named Mutex, because Unit doesn't support it. Instead, we create a file with `FileShare.None`.
+                // https://forum.unity.com/threads/unsupported-internal-call-for-il2cpp-mutex-createmutex_internal-named-mutexes-are-not-supported.387334/
+                if (ApplicationAdapter.Instance.Platform == RuntimePlatform.WindowsPlayer && options.CacheDirectoryPath != null)
```
bruno-garcia, line 41: nit:
```suggestion
                if (ApplicationAdapter.Instance.Platform is RuntimePlatform.WindowsPlayer && options.CacheDirectoryPath is not null)
```

bruno-garcia (approved):
👍

PR #648, opened 2022-03-18
CI: support a manual run and run on all pushes

vaind:
#skip-changelog

turns out the recent change wasn't so good... we still don't get proper CI on 3rd-party PRs and now we also don't get CI on changes by the unity-update action. This should fix that and with the change of the license secret to allow 3rd-party branches, we could even use it to manually trigger a run on those branches (hopefully)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,15 +3,9 @@ name: CI
 on:
   push:
     paths-ignore:
-      - '**/*.md'
-      - '**/*.txt'
-    branches:
-      - main
-      - release/**
-  pull_request:
-    paths-ignore:
-      - '**/*.md'
-      - '**/*.txt'
+      - '**.md'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -3,15 +3,9 @@ name: CI
...
-      - '**/*.txt'
-    branches:
-      - main
-      - release/**
-  pull_request:
-    paths-ignore:
-      - '**/*.md'
-      - '**/*.txt'
+      - '**.md'
+      - '**.txt'
```
vaind, line 7: I've realized '**' is implemented as that it just matches everything including '/' (while '*' only matches any character except '/') - thus no need for '/*' in the pattern

bitsandfoxes (approved):

bruno-garcia (approved):

PR #654, opened 2022-03-21
Added ability to specify additional debug symbol paths

LukasKastern:
This change exposes a list in the DebugSymbolsTab and SentryCliOptionsEditor in which the user can enter custom symbol paths.
My personal use case for this is to specify the **lib_burst_generated.pdb** that burst generates during compilation.

### DIFF
--- a/src/Sentry.Unity.Editor/ConfigurationWindow/DebugSymbolsTab.cs
+++ b/src/Sentry.Unity.Editor/ConfigurationWindow/DebugSymbolsTab.cs
@@ -1,10 +1,24 @@
 using UnityEditor;
+using UnityEditorInternal;
 using UnityEngine;
 
 namespace Sentry.Unity.Editor.ConfigurationWindow
 {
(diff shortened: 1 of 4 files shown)


bitsandfoxes:
Hey @LukasKastern! Welcome and thanks for contributing! I'm curious: Where does `lib_burst_generated.pdb` end up that you have to specify the path yourself? I just want to make sure that that's not a gap we have in general.

LukasKastern:
It's in two separate locations, next to the game executable and in ProjectName_BurstDebugInformation_DoNotShip.

I assumed it wouldn't be picked up because it wasn't directly referenced in the paths lists but after making another build without specifying it as an extra symbol path it looks like it's uploaded just fine. 
Seems to just have been the same issue as the one that #655 fixed.  

On a different note, the pdb is uploaded fine but the .dll is missing. I'm getting **A required debug information file was missing. lib_burst_generated.dll** in the dashboard. Is that actually an issue or can this be safely ignored?

PR #655, opened 2022-03-21
Added parentheses to symbol paths which are passed to sentry-cli

LukasKastern:
Adding parentheses to symbol paths fixes an issue where some symbols wouldn't be picked up when the project had a whitespace in its name.

#skip-changelog


bitsandfoxes (approved):
Nice catch! Thanks!
Do you mind adding an entry to the changelog?

PR #537, opened 2022-01-31
feat: WebGL support for C# layer

bruno-garcia:
This will superseed #329 with the final WebGL support.

This PR in the .NET SDK has the change needed to add a customer BW here: https://github.com/getsentry/sentry-dotnet/pull/1450

PR #657, opened 2022-03-22
feat: WebGL - dotnet support

### LINKED ISSUE #516: Question/Suggestion: when/how to get WebGL working
I understand sentry-unity doesn't support WebGL, because of single-threading (?). But as our app run on many platforms, is there a way I can re-use most of the Sentry logic for building the events, but just do the actual sending myself with UnityWebRequest? Is there a way I could capture the event to be sent and handle it myself?

And are there any plans to make WebGL builds work with this plugin?

Thanks!

### LINKED ISSUE #600: Attachment silently fails to serialize when using BackgroundWorker on WebGL
### Environment

How do you use Sentry? **Sentry SaaS (sentry.io)**

Which version of the SDK? **0.10.1**

Which version of Unity? **2020.3.23f1**

Is this happening in Unity (editor) or on a player like Android, iOS, Windows? **On WebGL player**

### Background

I'm running Sentry on WebGL. Thanks to the new support for IBackgroundWorker to handle the sending of Envelopes, it was easy enough to make Sentry report error events on WebGL. My BackgroundWorker is very simple, it holds a queue and when flushed, it serializes each envelope and sends it using UnityWebRequest. The flushing happens in a coroutine. However, I noticed that user error reports did not work. I found a work around to this, but it may point to some issue in how attachments are handled (or it's just something that would be naturally solved when this SDK gets official WebGL support).

### Steps to Reproduce

[I have made a repro repo](https://bitbucket.org/fictivr/repro-sentry-unity-attach/src/master/). In short, it's an empty project with Sentry installed. Two keys trigger two different functions in `TestCase.cs` for attaching a screenshot that is generated (for the user report).

`A` key triggers code that adds attachment simply with `scope.AddAttachment(filePath)` which causes `Envelope.SerializeAsync` to later silently fail. 

`B` key triggers an otherwise identical function that instead attaches the screenshot like this:
```
using Stream fileStream = File.Open(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
var data = new byte[fileStream.Length];
fileStream.Read(data, 0, data.Length);
fileStream.Seek(0, SeekOrigin.Begin);
scope.AddAttachment(data, filename);
```

Clone the repo, go to `Assets/TestCase.cs` and enter a correct DSN in the code. If you then run this in Editor, it makes no difference whether i press A or B, both works as expected. If I build and run in WebGL, A silently fails but B works.

Here's a sample browser console log that first shows A version, then B version: https://bitbucket.org/fictivr/repro-sentry-unity-attach/src/master/test-case-error-log.txt

No error is seen in the log, two envelopes are queued up in both, but in A, one of the envelopes (the event with attachment) is never sent (SerializeAsync fails such that the web request never gets sent, but I haven't been able to catch any exception - but in WebGL that often doesn't work due to IL2CPP).

What puzzles me is that just opening the file myself and giving the raw data works, while adding the file path fails.

vaind:
* Replaces #537 & #329
* closes #516 
* closes #600 

### Approaches
I've gone through multiple iterations/alternative approaches. You can see them in previous commits:
* Background worker - OK, but no business logic reuse from sentry-dotnet's `HttpTransport` - https://github.com/getsentry/sentry-unity/blob/472cb003bb1cf9c04d380b4909c4ae3faeeff451/src/Sentry.Unity/WebGL/SentryWebGL.cs 
* Replace `HttpMessageHandler` while using the full `HttpTransport` without any changes needed - this looked promising up until the point where I couldn't get `await` that polute the whole `HttpTransport` to ever complete (continue) on WebGL - https://github.com/getsentry/sentry-unity/blob/bb993e70ad4bc1a999b4097e2c86b2aff8b69eb3/src/Sentry.Unity/WebGL/SentryWebGL.cs
* Override `HttpTransport` with a custom implementation for the async stuff -> no `await` - the most complete approach so far - https://github.com/getsentry/sentry-unity/blob/2bbbcd905125398683dc39441409b280c68c4f23/src/Sentry.Unity/WebGL/SentryWebGL.cs

### MVP TODOs
* [x] Unity 2021 fails on accessing [RuntimeInformation.FrameworkDescription](https://github.com/getsentry/sentry-dotnet/blob/38bd5e291496440528f817851ba0ebc133ffa972/src/Sentry/PlatformAbstractions/RuntimeInfo.cs#L109) - [reported to Unity](https://fogbugz.unity3d.com/default.asp?1413035_ensfb816j2jfmq03) but I guess we'll need [this workaround](https://github.com/getsentry/sentry-dotnet/pull/1554) to unblock this PR
* [x] wait until ~https://github.com/getsentry/sentry-dotnet/pull/1553~ https://github.com/getsentry/sentry-dotnet/pull/1560 is merged
* [x] basic CI build
* [x] CI smoke/integration test
* [x] requires current sentry-dotnet main

### Future PRs
* Capture Javascript & native errors - #668
* warn/error out during build when `"com.unity.modules.unitywebrequest": "1.0.0"` package is disabled - otherwise Unity just throws a vague error: `Cannot create web request without initializing the system` that I couldn't get any useful resolution for by googling - requires editor build script so could be also part of #668 which brings that.


src/Sentry.Unity/WebGL/SentryWebGL.cs
```diff
@@ -0,0 +1,227 @@
...
+            options.DiagnosticLogger?.LogDebug("Updating configuration for Unity WebGL.");
+
+            // Caching transport relies on a background thread
+            options.CacheDirectoryPath = null;
+            // Note: we need to use a custom background worker which actually doesn't work in the background
+            // because Unity doesn't support async (multithreading) yet. This may change in the future so let's watch
+            // https://docs.unity3d.com/2019.4/Documentation/ScriptReference/PlayerSettings.WebGL-threadsSupport.html
+            options.BackgroundWorker = new WebBackgroundWorker(options, SentryMonoBehaviour.Instance);
+
+            options.CrashedLastRun = () => false; // no way to recognize crashes in WebGL yet
```
bruno-garcia, line 42: Note: Align the concept of 'crash' sessions with the JS SDK but it's never about 'on-restart' which is what this field is used for

```suggestion
            options.CrashedLastRun = () => false; // Since there's no offline caching, this won't be called anyway
```

src/Sentry.Unity/WebGL/SentryWebGL.cs
```diff
@@ -0,0 +1,227 @@
...
+
+            // Still can't find out what's using Threads so:
```
bruno-garcia, line 44: I recall writing this one :D
I'm not sure which of the properties below are needed

vaind: none of them has to be disabled - at least I didn't find any issues with my manual testing and don't really look like having any problematic code (a.k.a. task await). Though I must say I didn't know how to trigger some of the features, like the`AppDomainUnhandledExceptionCapture`.

src/Sentry.Unity/WebGL/SentryWebGL.cs
vaind: The suggested comment is not true anymore because it turns out we can have offline caching just fine (uses indexeddb automatically as a filesystem)
image

mattjohnsonpint:
@vaind - Please try layering on top of the changes I made with getsentry/sentry-dotnet#1560 and let me know if you have any pain points.  I think it should let your code be much cleaner.  Thanks.

src/Sentry.Unity/WebGL/SentryWebGL.cs
```diff
@@ -0,0 +1,163 @@
...
+            try
+            {
+                envelope.Serialize(stream, _options.DiagnosticLogger);
+                stream.Flush();
+            }
+            catch (Exception e)
+            {
+                _options.DiagnosticLogger?.LogError("Failed to serialize Envelope into the network stream", e);
+                throw;
+            }
```
vaind, line 114: @mattjohnsonpint do you think this is acceptable or do you want to expose sentry-dotnet's `EnvelopeHttpContent` & add a synchronous `SerializeToStream` method?

mattjohnsonpint: Thanks for pointing that out.  I made a couple more changes.  You should now just be able to do the following instead:

```csharp
using var stream = ReadStreamFromHttpContent(message.Content);
```

vaind: thx

src/Sentry.Unity/WebGL/SentryWebGL.cs
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    HandleResponse(response, processedEnvelope);
+                }
+            }
+        }
+
+        private UnityWebRequest CreateWebRequest(HttpRequestMessage message)
+        {
+            // Note: In order to use the synchronous Envelope.Serialize() we ignore the `message.Content`
+            // which is an `EnvelopeHttpContent` instance and use the actual envelope it wraps.
```
mattjohnsonpint, line 103: I think these comments are incorrect now and can be deleted.

vaind: yes, accidental leftover, removing. Thanks

src/Sentry.Unity/WebGL/SentryWebGL.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            return www;
+        }
+
+        private HttpResponseMessage? GetResponse(UnityWebRequest www)
+        {
+            // Let's disable treating "warning:obsolete" as an error here because the alternative of putting a static
+            // function to user code (to be able to use #if UNITY_2019) is just ugly.
+#pragma warning disable 618
+            // if (www.result == UnityWebRequest.Result.ConnectionError) // unity 2021+; `.result` not present on 2019
```
bitsandfoxes, line 127:
```suggestion
            // if (www.result == UnityWebRequest.Result.ConnectionError) // Unity 2020.1+; `.result` not present on 2019
```

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Features
+
+- Bump Sentry Java SDK to v6.0.0-alpha.4 ([#653](https://github.com/getsentry/sentry-unity/pull/653))
+  - [changelog](https://github.com/getsentry/sentry-java/blob/6.0.0-alpha.4/CHANGELOG.md)
+  - [diff](https://github.com/getsentry/sentry-java/compare/5.5.0...6.0.0-alpha.4)
```
bruno-garcia, line 9: Already in the fixes block below

```suggestion
```

src/Sentry.Unity/WebGL/SentryWebGL.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        private UnityWebRequest CreateWebRequest(HttpRequestMessage message)
+        {
+            using var contentStream = ReadStreamFromHttpContent(message.Content);
+            var www = new UnityWebRequest
+            {
+                url = message.RequestUri.ToString(),
+                method = message.Method.Method.ToUpperInvariant(),
+                uploadHandler = new UploadHandlerRaw(((MemoryStream)contentStream).ToArray()),
```
bruno-garcia, line 110: Is this cast going to be safe straight up?

https://github.com/getsentry/sentry-dotnet/blob/f28b1c513547413340c5077189b57b051e15fdf9/src/Sentry/Http/HttpTransportBase.cs#L228

CHANGELOG.md
vaind: I thought I've already removed that...

src/Sentry.Unity/UnityWebRequestTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+            _ = _behaviour.StartCoroutine(_transport.SendEnvelopeAsync(envelope));
+            return true;
+        }
+
+        public Task FlushAsync(TimeSpan timeout) => Task.CompletedTask;
+
+        public int QueuedItems { get; }
+    }
+
+    internal class UnityWebRequestTransport : HttpTransportBase
```
bruno-garcia, line 37: Even though it's not the goal of this PR. this transport has the potential to unblock usages outside WebGL. Mainly Console platforms. We can refactor things out to be reused later, so that perhaps a selected number of platforms. Possibly even: Anything that's not desktop and mobile

src/Sentry.Unity/UnityWebRequestTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        private readonly SentryUnityOptions _options;
+
+        public UnityWebRequestTransport(SentryUnityOptions options)
+            : base(options)
+        {
+            _options = options;
+        }
```
bruno-garcia, line 45: style suggestion:

```suggestion
        => _options = options;
```

src/Sentry.Unity/UnityWebRequestTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+                // Send envelope to ingress
+                var httpRequest = CreateRequest(processedEnvelope);
+                var www = CreateWebRequest(httpRequest);
+                yield return www.SendWebRequest();
+
+                var response = GetResponse(www);
+                if (response != null)
+                {
+                    HandleResponse(response, processedEnvelope);
+                }
```
bruno-garcia, line 62: When can it be null? Should we log?

```suggestion
                if (response is not null)
                {
                    HandleResponse(response, processedEnvelope);
                } else {
                  // log?
                }
```

src/Sentry.Unity/UnityWebRequestTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                _options.DiagnosticLogger?.LogWarning("Failed to send request: {0}", www.error);
+                return null;
+            }
+
+            var response = new HttpResponseMessage((HttpStatusCode)www.responseCode);
+            foreach (var header in www.GetResponseHeaders())
+            {
+                // Unity would throw if we tried to set content-type or content-length
+                if (header.Key != "content-length" && header.Key != "content-type")
```
bruno-garcia, line 110: nit: prefer pattern matching over operations
```suggestion
                if (header.Key is not "content-length" && header.Key is not "content-type")
```

bruno-garcia (approved):
This is awesome! :shipit: 

I left some notes for and nits for reference

samples/unity-of-bugs/Assets/Scripts/NativeSupport/JavaScriptPlugin.jslib
```diff
@@ -0,0 +1,14 @@
+mergeInto(LibraryManager.library, {
+
+  throwJavaScript: function () {
+    var something = undefined;
+    // Note: if we trigger the JS error by calling `something.do();` directly here, Unity get's stuck:
```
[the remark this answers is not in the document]
vaind: yes https://fogbugz.unity3d.com/default.asp?1415583_kc2sm3722bi2ci9s (I don't think you can see attachments but I've made a repro project)

src/Sentry.Unity/UnityWebRequestTransport.cs
vaind: that's why I've moved it to a separate file in the Sentry.Unity csproj - should be a good starting point without major git-diffs when we want to use for another platform.

src/Sentry.Unity/UnityWebRequestTransport.cs
vaind: it's already logged when null (i.e. when there was a network error)

galenmolk:
Hi @vaind ! Thank you for your work on WebGL support for this SDK. We're Sentry customers utilizing the tool on other platforms, so we're excited to be able to use it for WebGL too.

I had a quick question: as far as you know, are there Unity Editor version minimums for Sentry Unity <> WebGL compatibility?

vaind:
> Hi @vaind ! Thank you for your work on WebGL support for this SDK. We're Sentry customers utilizing the tool on other platforms, so we're excited to be able to use it for WebGL too.
> 
> I had a quick question: as far as you know, are there Unity Editor version minimums for Sentry Unity <> WebGL compatibility?

Glad you find it useful. It should work on standard versions as for the rest of the SDK, i.e. 2019, 2020, 2021 at the moment. We're testing WebGL specifically in our integration test CI on those.

galenmolk:
> > Hi @vaind ! Thank you for your work on WebGL support for this SDK. We're Sentry customers utilizing the tool on other platforms, so we're excited to be able to use it for WebGL too.
> > I had a quick question: as far as you know, are there Unity Editor version minimums for Sentry Unity <> WebGL compatibility?
> 
> Glad you find it useful. It should work on standard versions as for the rest of the SDK, i.e. 2019, 2020, 2021 at the moment. We're testing WebGL specifically in our integration test CI on those.

Thanks for getting back to me so fast. Sounds good!

jfbloom22:
Awesome work! So glad to have WebGL dotnet support!

PR #666, opened 2022-03-29
feat: Added override for Sentry CLI URL

Swatinem:
This is useful for self-hosted users, as well as for local testing ;-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Allow configuring the Sentry URL ([#666](https://github.com/getsentry/sentry-unity/pull/666))
+- Adding override for Sentry CLI URL ([#666](https://github.com/getsentry/sentry-unity/pull/666))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


vaind:
Why do we need to configure a different URL than the one in the main options tab?
image

Swatinem:
The new setting is specific to how sentry-cli is invoked, as in: where to upload the debug files.

vaind:
> The new setting is specific to how sentry-cli is invoked, as in: where to upload the debug files.

Got it, but why would we want/need it to be different than the DSN? When would it be necessary to have symbol upload to a different server than the one where the events will be uploaded?

Swatinem:
> to a different server than the one where the events will be uploaded?

Well you wouldn’t. That’s why the setting is needed, otherwise sentry-cli would always try to upload to the SaaS instance.

Or are you rather suggesting to parse the DSN and extract the url from there?

vaind:
> That’s why the setting is needed, otherwise sentry-cli would always try to upload to the SaaS instance.

sorry, my bad. I've assumed it already uses the DSN.

> Or are you rather suggesting to parse the DSN and extract the url from there?

IMO we should do exactly that. Otherwise, it wouldn't work with self-hosted instances, which I guess is a common use case.

bitsandfoxes (approved):

Issue #471, opened 2021-12-15
Error sending when exception thrown in new thread (callbacks also not on unity main thread when this happens)


oddgames-david:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
Sentry Unity 0.8.0

### Steps to Reproduce

Initialize sdk

```
        SentryUnity.Init(o =>
        {
            o.Dsn = "XXXX";
            // Useful when setting things up to see the SDK logs in the Unity console.
            o.Debug = true;
            o.StackTraceMode = StackTraceMode.Enhanced;
            o.EnableLogDebouncing = true;
            o.AndroidNativeSupportEnabled = true;
            o.IosNativeSupportEnabled = true;
            o.TracesSampleRate = UnityEngine.Debug.isDebugBuild ? 1 : 0.01f;
     
        });
```

Throw an exception from a thread that isn't the unity main thread
`
new Thread(() => throw new System.Exception("This is a test exception from another thread.")).Start()
`
### Expected Result

No error, BeforeSend and other callbacks should be running on main thread so I can easily add a screenshot or other data from non thread safe areas.

### Actual Result

Throws an error but also seems to run all callbacks off the main thread so I can't use unity engine code when applying tags or attachments (i've removed all callback code for this test, perhaps this is two bugs).

Sentry: (Error) Failed to send cached envelope: C:/Users/David/AppData/LocalLow/ODD Games/MTD\Sentry\F187FFF4B6258D70B60D1537BB1C9F2AF2C703A1\__processing\1639566925_-1552_cc380117b5b34bd2a94c3cff3b94e2c4_390037632.envelope, retrying after a delay. System.Threading.Tasks.TaskCanceledException: A task was canceled.
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x00026] in <695d1cc93cca45069c528c15c9fdd749>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x00028] in <695d1cc93cca45069c528c15c9fdd749>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) [0x00008] in <695d1cc93cca45069c528c15c9fdd749>:0 
[...]

Issue #231, opened 2021-06-14
Special defaults for Desktop vs Mobile vs Xyz


bruno-garcia:
We're adding defaults that make sense on Mobile. For example #230. This features does make sense on Desktop though so we could have different defaults for Mobile vs Desktop.

- [ ] On desktop environments default `IsEnvironmentUser` to true.

bruno-garcia:
We could just flip this flag in the Sentry native Windows support configure method.
So if a user opts-in to SendDefaultPii, we'll take the username of the user logged in to the machine.

PR #670, opened 2022-03-30
Feat: screenshots

### LINKED ISSUE #284: Add `AttachScreenshot` option
When the option is set to true, events will include a screenshot of the game.

This feature is being added to Xamarin:

https://github.com/getsentry/sentry-xamarin/search?q=screenshot

https://github.com/getsentry/sentry-xamarin/blob/411be60d616789ad81ec13e2e50c7c2a777b4c67/Src/Sentry.Xamarin/Internals/Device/Screenshot/ScreenshotEventProcessor.droid.ios.cs

vaind:
Closes #284 

### TODO:
* [x] update smoke test
* [x] resolution & quality configuration
* [x] update sentry-dotnet to the latest version to get https://github.com/getsentry/sentry-dotnet/pull/1557

### DIFF
--- a/samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
+++ b/samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
@@ -1,27 +1,3 @@
-#if !UNITY_EDITOR
-#if UNITY_IOS
-#define SENTRY_NATIVE_IOS
-#elif UNITY_ANDROID
-#define SENTRY_NATIVE_ANDROID
-#elif UNITY_STANDALONE_WIN && ENABLE_IL2CPP
-#define SENTRY_NATIVE_WINDOWS
-#endif
-#endif
-
-#if SENTRY_NATIVE_IOS
-using Sentry.Unity.iOS;
-#elif UNITY_ANDROID
-using Sentry.Unity.Android;
-#elif SENTRY_NATIVE_WINDOWS
-using Sentry.Unity.Native;
-#endif
-
-#if UNITY_IOS
-using System.IO;
-using System.Diagnostics;
-using Debug = UnityEngine.Debug;
-#endif
-
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -30,38 +6,19 @@ using System.Net.Http;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Web;
 using Sentry;
 using Sentry.Infrastructure;
 using Sentry.Unity;
 using UnityEngine;
+using Debug = UnityEngine.Debug;
 
 public class SmokeTester : MonoBehaviour
 {
     public void Start()
     {
-        string arg = null;
-#if SENTRY_NATIVE_ANDROID
-        using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
-        using (var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity"))
-        using (var intent = currentActivity.Call<AndroidJavaObject>("getIntent"))
-        {
-            arg = intent.Call<String>("getStringExtra", "test");
-        }
-#elif UNITY_IOS
-        // .net `Environment.GetCommandLineArgs()` doens't seem to work on iOS so we get the test arg in Objective-C
-        arg = getTestArgObjectiveC();
-#else
-        var args = Environment.GetCommandLineArgs();
-        if (args.Length > 2 && args[1] == "--test")
-        {
-            arg = args[2];
-        }
-#endif
-        if (arg == null)
-        {
-            Debug.Log($"SmokeTest not executed - no argument given");
-        }
-        else if (arg == "smoke")
+        var arg = GetTestArg();
+        if (arg == "smoke")
         {
             SmokeTest();
         }
@@ -77,7 +34,7 @@ public class SmokeTester : MonoBehaviour
         {
             HasCrashedTest();
         }
-        else
+        else if (arg != null)
         {
             Debug.Log($"Unknown command line argument: {arg}");
             Application.Quit(-1);
@@ -85,8 +42,33 @@ public class SmokeTester : MonoBehaviour
     }
 
 #if UNITY_IOS
-    [DllImport("__Internal")]
-    private static extern string getTestArgObjectiveC();
+    // .NET `Environment.GetCommandLineArgs()` doesn't seem to work on iOS so we get the test arg in Objective-C
+    [DllImport("__Internal", EntryPoint="getTestArgObjectiveC")]
+    private static extern string GetTestArg();
+#else
+    private static string GetTestArg()
+    {
+        string arg = null;
+#if UNITY_EDITOR
+#elif UNITY_ANDROID
+        using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
+        using (var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity"))
+        using (var intent = currentActivity.Call<AndroidJavaObject>("getIntent"))
+        {
+            arg = intent.Call<String>("getStringExtra", "test");
+        }
+#elif UNITY_WEBGL
+        var uri = new Uri(Application.absoluteURL);
+        arg = HttpUtility.ParseQueryString(uri.Query).Get("test");
+#else
+        var args = Environment.GetCommandLineArgs();
+        if (args.Length > 2 && args[1] == "--test")
+        {
+            arg = args[2];
+        }
+#endif
+        return arg;
+    }
 #endif
 
     private static TestHandler t = new TestHandler();
@@ -96,7 +78,13 @@ public class SmokeTester : MonoBehaviour
     // Forwarded from SmokeTestOptions.Configure()
     public static void Configure(SentryUnityOptions options)
     {
-        Debug.Log("SmokeTester.Configure() called");
+        if (GetTestArg() == null)
+        {
+            Debug.Log("SmokeTester.Configure() called but skipped because this is not a SmokeTest (no arg)");
+            return;
+        }
+
+        Debug.Log("SmokeTester.Configure() running");
         options.CreateHttpClientHandler = () => t;
         _crashedLastRun = () =>
         {
@@ -121,7 +109,7 @@ public class SmokeTester : MonoBehaviour
             t.ExpectMessage(currentMessage, "'type':'session'");
 
             var guid = Guid.NewGuid().ToString();
-            Debug.LogError(guid);
+            Debug.LogError($"LogError(GUID)={guid}");
 
             // Skip the session init requests (there may be multiple of othem). We can't skip them by a "positive"
             // because they're also repeated with standard events (in an envelope).
@@ -136,12 +124,12 @@ public class SmokeTester : MonoBehaviour
             Debug.Log($"Done skipping non-event requests. Last one was: #{currentMessage}");
 
             t.ExpectMessage(currentMessage, "'type':'event'");
-            t.ExpectMessage(currentMessage, guid);
+            t.ExpectMessage(currentMessage, $"LogError(GUID)={guid}");
             t.ExpectMessage(currentMessage, "'filename':'screenshot.jpg','attachment_type':'event.attachment'");
 
-            SentrySdk.CaptureMessage(guid);
+            SentrySdk.CaptureMessage($"CaptureMessage(GUID)={guid}");
             t.ExpectMessage(++currentMessage, "'type':'event'");
-            t.ExpectMessage(currentMessage, guid);
+            t.ExpectMessage(currentMessage, $"CaptureMessage(GUID)={guid}");
             t.ExpectMessage(currentMessage, "'filename':'screenshot.jpg','attachment_type':'event.attachment'");
 
             var ex = new Exception("Exception & context test");
@@ -279,7 +267,10 @@ public class SmokeTester : MonoBehaviour
 
                 // Exit Code 200 to avoid false positive from a graceful exit unrelated to this test run
                 exitCode = 200;
+
+#if !UNITY_WEBGL  // We don't quit on WebGL because outgoing HTTP requests (in coroutines) would be cancelled.
                 Application.Quit(exitCode);
+#endif
             }
         }
 
@@ -317,8 +308,14 @@ public class SmokeTester : MonoBehaviour
 
         public bool CheckMessage(int index, String substring)
         {
+#if UNITY_WEBGL
+            // Note: we cannot use the standard checks on WebGL - it would get stuck here because of the lack of multi-threading.
+            // The verification is done in the python script used for WebGL smoke test
+            return true;
+#else
             var message = GetMessage(index);
             return message.Contains(substring) || message.Contains(substring.Replace("'", "\""));
+#endif
         }
 
         public void ExpectMessage(int index, String substring) =>
--- a/samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
+++ b/samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
@@ -137,10 +137,12 @@ public class SmokeTester : MonoBehaviour
 
             t.ExpectMessage(currentMessage, "'type':'event'");
             t.ExpectMessage(currentMessage, guid);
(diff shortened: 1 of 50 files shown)


bruno-garcia:
We can also replace this sample then: https://github.com/getsentry/sentry-unity/blob/dbc72c64c4cb5cee3bd58b458298aa726b7794cb/samples/unity-of-bugs/Assets/Scripts/AdditionalButtons.cs#L48

src/Sentry.Unity.Editor/ConfigurationWindow/EnrichmentTab.cs
```diff
@@ -72,6 +72,10 @@ internal static void Display(ScriptableSentryUnityOptions options)
                 new GUIContent("Report Assemblies Mode", "Whether or not to include referenced assemblies " +
                                                          "Version or InformationalVersion in each event sent to sentry."),
                 options.ReportAssembliesMode);
+
+            options.AttachScreenshot = EditorGUILayout.Toggle(
+                new GUIContent("Attach Screnshot", "Try to attach current screenshot for error events"),
```
bruno-garcia, line 77:
```suggestion
                new GUIContent("Attach Screnshot", "Try to attach current screenshot for error events.\n" +
                  "A screenshot might not be able to be attached, for example when the error happens on a background thread."),
```

src/Sentry.Unity/Integrations/IApplication.cs
```diff
@@ -25,6 +25,7 @@ public sealed class ApplicationAdapter : IApplication
 
         private ApplicationAdapter()
         {
+            // TODO why not logMessageReceivedThreaded? We're currently ignoring errors from other than the main thread.
```
bruno-garcia, line 28: I believe we receive events from other threads but they get queued to be pushed in the UI thread, isn't that the case? We used to have that but it had issues with processing events from the background thread because the SDK used APIs that required the UI thread. It might have changed now (I believe we already have checks "don't retrieve this data if NOT running in the UI thread) so we might be able to move back to `logMessageReceivedThreaded`.

If we confirm that ALL events go through `logMessageReceived` the trade off becomes "cause less overhead by capturing stuff on the background when that's possible but losing additional context that requires the UI thread"

src/Sentry.Unity/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,107 @@
+using System;
+using System.IO;
+using Sentry;
+using Sentry.Extensibility;
+using UnityEngine;
+using UnityEngine.Rendering;
+
```
bruno-garcia, line 7: nit:

```suggestion
```

src/Sentry.Unity/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,107 @@
...
+            _behaviour = behaviour;
+            _options = options;
+        }
+
+        public Stream GetStream()
+        {
+            if (!_behaviour.MainThreadData.IsMainThread())
+            {
+                _options.DiagnosticLogger?.LogDebug("Won't capture screenshot because we're not on the main thread");
+                return new MemoryStream();
```
bruno-garcia, line 33: Today this will render a broken UI preview because of a 0 byte attachment:

image


No tracking issue I don't think but we should figure something out

src/Sentry.Unity/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,107 @@
...
+                catch (Exception ex)
+                {
+                    _options.DiagnosticLogger?.LogError("Couldn't capture screenshot", ex);
+                }
+            }
+            return new MemoryStream();
+        }
+    }
+
+    // The async version, currently unused because we want the screenshot to be as early as possible after the error
```
bruno-garcia, line 57: OK it's quite a bit of code so maybe we store this in the GitHub issue as a comment so we don't lose it. but here it's a lot of deadcode

src/Sentry.Unity/ScreenshotAttachment.cs
vaind: With the current options we have in dotnet I don't think we can avoid sending an empty attachment.

I guess the UI needs to be fixed - The [specification](https://develop.sentry.dev/sdk/features/#screenshots) says: "In this case, a 0 byte attachment will be included. In that case, Sentry will not show a screenshot preview."

src/Sentry.Unity.Editor/ConfigurationWindow/EnrichmentTab.cs
vaind: btw I've removed "error" because with the current implementation, it's attached to messages as well.

src/Sentry.Unity/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                _options.DiagnosticLogger?.LogDebug("Won't capture screenshot because we're not on the main thread");
+                return new MemoryStream();
+            }
+            else
+            {
+                // The async version - uses a custom stream & captures in a coroutine
+                // return new ScreenshotCaptureStream(_options, _behaviour);
+
+                // Captures current screenshot synchronously
```
vaind, line 40:
```suggestion
                // Captures current screenshot synchronously
```

src/Sentry.Unity/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,107 @@
...
+    //     public override bool CanSeek => throw new NotImplementedException();
+    //     public override bool CanWrite => throw new NotImplementedException();
+    //     public override long Length => throw new NotImplementedException();
+    //     public override long Position { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+    //     public override void Flush() => throw new NotImplementedException();
+    //     public override int Read(byte[] buffer, int offset, int count) => throw new NotImplementedException();
+    //     public override long Seek(long offset, SeekOrigin origin) => throw new NotImplementedException();
+    //     public override void SetLength(long value) => throw new NotImplementedException();
+    //     public override void Write(byte[] buffer, int offset, int count) => throw new NotImplementedException();
+    // }
```
vaind, line 106:
```suggestion
```

src/Sentry.Unity/Integrations/IApplication.cs
vaind: I've tested it by running `Task.Run(() => Debug.LogError("Debug.LogError() in a background thread"));` and it's not forwarded, while the plain `Debug.LogError()` is. I'd say we should address this in https://github.com/getsentry/sentry-unity/issues/471 & add buttons for the test cases (or maybe a global checkbox that causes all buttons to work on background). Leaving this conversation unresolved and part of the main issue.

src/Sentry.Unity/ScreenshotAttachment.cs
lucas-zimerman: https://github.com/getsentry/sentry-dotnet/pull/1557
With this PR up you can instead of logging with debug log, throw an exception with the above message that the attachment will not be linked to the current event.

vaind: Maybe if we could change `public Stream GetStream()` to `public Stream? GetStream()` then we wouldn't even need to log an exception in the above case, because we can recognize it's not going to be available. But because `IAttachmentContent` is a public interface, any change would be breaking...

vaind: What about returning a custom `NoStream()` just for this case and recognizing it while adding the attachment?

src/Sentry.Unity/Integrations/IApplication.cs
vaind: resolved by https://github.com/getsentry/sentry-unity/pull/673

lucas-zimerman:
Out of curiosity, does this feature works with UWP? I recall on Xamarin that when I tried to take a screenshot it usually freezes the app on UWP so I just disabled it on that platform 😅

vaind:
> We can also replace this sample then:
> 
> https://github.com/getsentry/sentry-unity/blob/dbc72c64c4cb5cee3bd58b458298aa726b7794cb/samples/unity-of-bugs/Assets/Scripts/AdditionalButtons.cs#L48

Replace by what though? I'd actually remove that.

src/Sentry.Unity/ScreenshotAttachment.cs
vaind: After a finding today that CaptureScreenshotAsTexture() only throws in development build and crashes the app in a release one, I've brought back the manual check for the main thread and thrown an exception explicitly. Starting the text with "Sentry:" seems to cause some part of the SDK (I think I've seen that somewhere but can't remember) to not report that exception.

src/Sentry.Unity.Editor/ConfigurationWindow/EnrichmentTab.cs
```diff
@@ -45,18 +45,19 @@ internal static void Display(ScriptableSentryUnityOptions options)
             EditorGUI.DrawRect(EditorGUILayout.GetControlRect(false, 1), Color.gray);
             EditorGUILayout.Space();
 
-            options.SendDefaultPii = EditorGUILayout.BeginToggleGroup(
-                new GUIContent("Send default Pii", "Whether to include default Personal Identifiable " +
-                                                   "Information."),
-                options.SendDefaultPii);
+            {
```
bitsandfoxes, line 48: Why the scope?

src/Sentry.Unity.Editor/ConfigurationWindow/SentryEditorWindowInstrumentation.cs
```diff
@@ -44,6 +44,26 @@ private static void ConfigureOptions(Dictionary<string, string> args, [CallerMem
                 OptionsConfigurationDotNet.SetScript(value);
             }
 
+            if (args.TryGetValue("attachScreenshot", out value))
+            {
+                bool boolValue;
+                switch (value.ToLowerInvariant())
```
bitsandfoxes, line 50: That's a pretty nice touch!

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -38,6 +38,10 @@ public static string GetConfigPath(string? notDefaultConfigName = null)
         [field: SerializeField] public string ReleaseOverride { get; set; } = string.Empty;
         [field: SerializeField] public string EnvironmentOverride { get; set; } = string.Empty;
         [field: SerializeField] public bool AttachStacktrace { get; set; }
+        [field: SerializeField] public bool AttachScreenshot { get; set; }
+        [field: SerializeField] public int ScreenshotMaxWidth { get; set; }
+        [field: SerializeField] public int ScreenshotMaxHeight { get; set; }
```
bitsandfoxes, line 43: I'm not sure about dealing with absolute pixel sizes here. It does not take into account devices with different aspect ratios or switching from portrait to landscape.
Maybe we should start with a ratio instead?

src/Sentry.Unity/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            RenderTexture.ReleaseTemporary(rtFull);
+
+            // Remember the previous render target and change it to our target texture.
+            var previousRT = RenderTexture.active;
+            RenderTexture.active = rtResized;
+
+            try
+            {
+                // actually copy from the current render target a texture & read data from the active RenderTexture
```
bitsandfoxes, line 76:
```suggestion
                // Actually reads the pixels from the current render target and writes them to the texture.
```

src/Sentry.Unity.Editor/ConfigurationWindow/EnrichmentTab.cs
```diff
@@ -72,6 +73,25 @@ internal static void Display(ScriptableSentryUnityOptions options)
                 new GUIContent("Report Assemblies Mode", "Whether or not to include referenced assemblies " +
                                                          "Version or InformationalVersion in each event sent to sentry."),
                 options.ReportAssembliesMode);
+
```
bitsandfoxes, line 76:
```suggestion
            
        EditorGUILayout.Space();
        EditorGUI.DrawRect(EditorGUILayout.GetControlRect(false, 1), Color.gray);
        EditorGUILayout.Space();
            
```

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
vaind: We've discussed in the call to align across other SDKs. @bitsandfoxes is going to gather inputs

src/Sentry.Unity.Editor/ConfigurationWindow/EnrichmentTab.cs
vaind: just find it easier to read

src/Sentry.Unity.Editor/ConfigurationWindow/SentryEditorWindowInstrumentation.cs
```diff
@@ -44,6 +44,26 @@ private static void ConfigureOptions(Dictionary<string, string> args, [CallerMem
                 OptionsConfigurationDotNet.SetScript(value);
             }
 
+            if (args.TryGetValue("attachScreenshot", out value))
+            {
+                bool boolValue;
```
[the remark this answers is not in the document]
vaind: didn't know about that and while it's not exactly the same (doesn't handle 1 & 0), it's good enough for me.

src/Sentry.Unity.Editor/ConfigurationWindow/SentryEditorWindowInstrumentation.cs
```diff
@@ -44,6 +44,26 @@ private static void ConfigureOptions(Dictionary<string, string> args, [CallerMem
...
+                    case "1":
+                        boolValue = true;
+                        break;
+                    case "false":
+                    case "0":
+                        boolValue = false;
+                        break;
+                    default:
+                        throw new ArgumentException("Unknown boolean argument value: " + value, "attachScreenshot");
+                }
```
vaind, line 62:
```suggestion
                if (!Boolean.TryParse(value, out boolValue))
                {
                    throw new ArgumentException("Unknown boolean argument value: " + value, "attachScreenshot");
                }
```

samples/unity-of-bugs/Assets/Scripts/SmokeTester.cs
```diff
@@ -137,10 +137,12 @@ public static void SmokeTest()
 
             t.ExpectMessage(currentMessage, "'type':'event'");
             t.ExpectMessage(currentMessage, guid);
+            t.ExpectMessage(currentMessage, "'filename':'screenshot.jpg','attachment_type':'event.attachment'");
```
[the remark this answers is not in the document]
vaind: I've added a check for `length:0`

vaind:
I've disabled for WebGL where the screenshots consistently **don't** work (they're black), and updated the tooltip with more information. If you think that's OK for now, please merge @bitsandfoxes  or @bruno-garcia

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -51,10 +51,27 @@ public static void Init(SentryUnityOptions options)
...
                         options.AutoSessionTracking = false;
+
                     }
                 }
 
                 var sentryDotNet = SentrySdk.Init(options);
+
+                if (options.AttachScreenshot)
+                {
+                    if (ApplicationAdapter.Instance.Platform is RuntimePlatform.WebGLPlayer)
```
bitsandfoxes, line 62: Just to make sure: I thought we wanted to flat out disable screenshots inside Configure with an appropriate log?

vaind: That's what happens here... Am I missing something?

bitsandfoxes: I was thinking in here: https://github.com/getsentry/sentry-unity/blob/242dd3acac6248a1cc27047ef71041f992893d94/src/Sentry.Unity/WebGL/SentryWebGL.cs#L14

bitsandfoxes (approved):

src/Sentry.Unity/SentryUnity.cs
bitsandfoxes: I think it'd fit nicely with future things like: https://github.com/getsentry/sentry-unity/issues/231
I want to get rid of the SentryOptionsUtility and move that stuff into its respective platform-specific Configure.
Maybe that's just me nitpicking.
LGTM! Thanks!

Issue #251, opened 2021-07-05
SentryInitialization.Init is always called in tests


semuserable:
`SentryInitialization.Init` is called even for the simplest test because of `RuntimeInitializeOnLoadMethod` attribute.

```csharp
public void AssertPass()
  => Assert.Pass();
```

I encountered some not expected side effects while I was writing tests for `UnityEventProcessor`.

bruno-garcia:
Not sure how we can skip it just for tests. [Doesn't look like there's compilation directive ](https://docs.unity3d.com/Manual/PlatformDependentCompilation.html)set when compiling for tests which makes sense since you might just run tests on a normal build.

bitsandfoxes:
We could disable the SDK selectively for those tests where we Initialize the SDK as part of the test by loading and disabling the ScriptableOptions within the editor?

vaind:
Hmm, why is it `[RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]`? Looks like something called multiple times even during normal runtime? Maybe we should instead use another event, for example this one sounds about right: [AfterAssembliesLoaded](https://docs.unity3d.com/ScriptReference/RuntimeInitializeLoadType.AfterAssembliesLoaded.html)

bitsandfoxes:
I think think this was due to `AfterAssembliesLoaded` and `BeforeSplashScreen` not being called in the Editor.

bitsandfoxes:
Disabling self-initialization during tests with [this](https://github.com/getsentry/sentry-unity/blob/main/test/Sentry.Unity.Tests/DisabledSelfInitializationTests.cs)
If individual test classes are supposed to be executed with self-initialization disabled they should inherit from this.

PR #673, opened 2022-04-04
Feat: background LogError & samples

vaind:
* resolves https://github.com/getsentry/sentry-unity/pull/670#discussion_r839493001

### DIFF
--- a/test/Sentry.Unity.Tests/IntegrationTests.cs
+++ b/test/Sentry.Unity.Tests/IntegrationTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using NUnit.Framework;
 using Sentry.Unity.Integrations;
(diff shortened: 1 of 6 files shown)


bitsandfoxes:
I modified the samples and moved the thread-samples into the `Additional Samples` scene. I think we should keep the samples as simple and copy&paste-able as possible.
image

bitsandfoxes:
While I do think that the `Action<Action> _executor` solution is pretty cool and nifty I think it's a bit of an overkill for the initial sample that ends up in the hands of users of all levels of experience.
Maybe we just move it into its own `Threaded Samples` scene.

test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -239,6 +240,65 @@ public IEnumerator BugFarmScene_DebugLogException_IsMarkedUnhandled()
...
+                while (testEventCapture.Events.Count == 0 && stopWatch.ElapsedMilliseconds < 5000)
+                {
+                    yield return null;
+                }
+            }
+
+            // assert
+            Assert.AreEqual(1, testEventCapture.Events.Count);
+            var isMainThread = testEventCapture.Events.First().Tags.SingleOrDefault(t => t.Key == "unity.is_main_thread");
+            Assert.AreEqual((!inTask).ToString(), isMainThread.Value);
```
bitsandfoxes, line 299: It seems flakey? 
https://github.com/getsentry/sentry-unity/actions/runs/2115833801/attempts/1
https://github.com/getsentry/sentry-unity/runs/5968904127?check_suite_focus=true

bitsandfoxes: Has this: https://github.com/getsentry/sentry-unity/issues/251 finally caught up with us?

vaind: fixed by making sure SDK Dispose() is being called before the test finishes, so that there are no outstanding envelopes

vaind: OK, not completely fixed then... https://github.com/getsentry/sentry-unity/runs/5990231610?check_suite_focus=true

vaind: Can't reproduce that locally, but I'm getting other errors similar to this:

```
Unhandled log message: '[Error] Sentry: (Error) Exception in background worker of CachingTransport. System.IO.IOException: Sharing violation on path C:/Users/dlugo/AppData/LocalLow/DefaultCompany/unity-of-bugs\Sentry\618762E37120B6E01219C43B9076FFEC1ACE8406\__processing\1649773679_3939__1948942592.envelope
  at System.IO.File.Delete (System.String path) [0x00073] in <eae584ce26bc40229c1b1aa476bfa589>:0 
  at Sentry.Internal.Http.CachingTransport+<InnerProcessCacheAsync>d__17.MoveNext () [0x002a9] in <e46e07d242f048179999156e39b482f9>:0 
```

Must be something related to initializing the .net SDK multiple times during the same runtime. I know the .net SDK isn't really supposed to be used like that, because we had to specifically implement workarounds to NOT initialize it multiple times on Windows for example. 

I'm thinking we should split out these unit tests out of the example app, so they're not subject to the same automatic initialization via `SentryInitialization.cs`. Alternatively, we can just configure Sentry to be disabled before running these tests.

vaind: I've removed the DSN from options in CI just before the PlayerMode test is run, so that SentryInitialization.init() doesn't run and we only run the manual init in the tests. Tests seem to be fixed now.

Issue #312, opened 2021-09-13
Use GHA `concurrency` to lock a single run per license used


bruno-garcia:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency

Currently two pushes will trigger two concurrent builds where the second fails due to license issues

vaind:
I've had a look at this after the painful start of the week with a lot of license failures... Currently `concurrency` doesn't provide an option to limit to a specific number, it only enforces "one" for a given group name. I was also thinking of using a modulus operator to construct the group name based on a job run ID - not ideal but kinda what we need, but there's no `%` in expressions either, so it would need to be a manually composed disgustingly-long expression composed of ternary operations to emulate the "modulus 6" (the current number of available licences)

PR #686, opened 2022-04-13
Chore: parallel smoketest build

### LINKED ISSUE #617: CI export Gradle project
See: https://github.com/getsentry/sentry-unity/pull/611#discussion_r820814841

The goal is to have the build with Unity only export the gradle project to be built on later step an parallelize things better

vaind:
Adding webgl added a lot of build time to the main `build` job. I'm thinking of splitting it up to separate parallel builds for android and WebGL (those two take most of the time). I originally wanted to do it in the WebGL PR, that's why this is built on top of that and has all the commits, but then I found out I need to make a few more changes to make the sample app use the extracted UPM package. In the end I decided to use the existing set of scripts already used to build integration-test app for Windows.

The changes are:
1. add a new `smoke-test-build` matrix job that builds for Android and WebGL
2. update scripts to accommodate those two changes
3. add an auto-retry wrapper to unity - `scripts/unity.ps1` - when a license activation fails, it will retry after a minute of wait. This change was previously discussed and becomes more important here when we have many jobs running at the same time. Would be solved by #312 but that one is blocked (unavailable on GHA).

closes #617 - it doesn't make much sense to have a Gradle export anymore, with the CI looking as it does now in this PR - we're already building on a separate platform. The best we could achieve with pre-building into a grade project first is a little more concurrency of Unity instances (due to the limited number of licences on the server), because after the gradle project is prepared, the app would be built outside of Unity, so the licence wouldn't be held. The benefit looks too small to even bother with this.

#skip-changelog


.github/workflows/ci.yml
```diff
@@ -145,68 +165,79 @@ jobs:
...
+      - name: Build with Sentry SDK
+        run: sudo pwsh ./test/Scripts.Integration.Test/integration-build-project.ps1 -UnityPath "${{ env.UNITY_PATH }}" -Platform "${{ matrix.platform }}"
+
+      - name: Prepare for upload
         run: |
-          docker exec unity dotnet msbuild /t:UnityConfigureSentryOptions /p:TestDsn=http://publickey@127.0.0.1:8000/12345 /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
-          docker exec unity dotnet msbuild /t:UnityBuildPlayerWebGL /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
+          cd samples/IntegrationTest/Build
+          sudo chown -R $(id -u):$(id -g) .
+          # sudo rm -rf *_BackUpThisFolder_ButDontShipItWithYourGame
```
bruno-garcia, line 234: This should be uncomented?

.github/workflows/ci.yml
```diff
@@ -145,68 +165,79 @@ jobs:
...
+    name: Build ${{ matrix.platform }} Smoke Test - ${{ matrix.unity-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        unity-version: ['2019', '2020', '2021']
+        platform: ['Android', 'WebGL']
+    env:
+      # Run the `unity-editor` command inside the `unity` docker container under the current user & group.
+      # UNITY_PATH: docker exec --user 1001:121 unity unity-editor
```
bruno-garcia, line 185: This serves as an example to debug ?

scripts/unity-utils.ps1
```diff
@@ -0,0 +1,149 @@
+$RunUnityLicenseRetryTimeoutSeconds = 3600
```
bruno-garcia, line 1: This file is a dupe of scripts/unity-utils.ps1
Do we need a copy? Can we use a symlink?

bruno-garcia (approved):

scripts/unity-utils.ps1
vaind: neither, one was unused, don't really know how that happened... removed

.github/workflows/ci.yml
vaind: cleaned up

.github/workflows/ci.yml
vaind: yes

Issue #287, opened 2021-08-23
Updated link.xml would not override existing one


bitsandfoxes:
We should make sure to override outdated data.

bitsandfoxes:
When taking a look at this please also note: https://github.com/getsentry/sentry-unity/issues/521

bitsandfoxes:
And this: https://github.com/getsentry/sentry-unity/issues/81

bruno-garcia:
I believe this was resolved, no?

bitsandfoxes:
Still doesn't override. We have not yet updated the link.xml once.

bruno-garcia:
Instead of opening a new issue this seems like the right spot to discuss:

Should we revisit if we really need link.xml now? Some of the rules added there were from 2018 when the Sentry SDK for .NET used `Newtonsoft.Json` with reflection. Now the serialization is mostly done manually. Only bits with reflection are what's set in `Event.Contexts` or `Event.Extra`.

bruno-garcia:
If the issue is overriding things added manually by the user, is this something we should even do?
They should have their own link.xml file, right?

bitsandfoxes:
Yes, users have their own. It was just a general observation that there is nothing in place that would override an already existing link.xml if we were to update.
Another thing: We never updated it after introducing aliasing. Why and how does it still work - I do not know.

PR #707, opened 2022-04-26
feat: Removed link.xml

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/287

Hypothesis: We required the `link.xml` back when the SDK was self-initializing from within the DLL. Since we now compile `SentryInitialization` with the actual project we no longer require it.

We added aliasing but we never updated the contents of the `link.xml` and it worked regardless.

To discuss: How does this change affect users updating the SDK? Should we look for and delete the file in their project?


bruno-garcia (approved):
Be brave

PR #423, opened 2021-11-18
Fix long path error on Windows build.

lucas-zimerman:
Fix reduces the path on windows so long path errors will no longer happen during Github Actions tests for now.

#skip-changelog

PR #710, opened 2022-04-27
feat: build macOS SDK as a dylib from sentry-cocoa

### LINKED ISSUE #693: Native crash support for macOS
* Support native crashes on macOS
* Update session status as `crashed`

Relates to #527, #694

vaind:
closes #693 - finally found a working solution that's reasonably integrable into Unity build - using a `dylib`, letting Unity add it to the project, and accessing the content dynamically,  using `dlopen()/dsym()`.

Replaces #703 which was deemed too complicated as well as limited (required users to export to xcode).

### TODOs:
* [x] PoC - calling an exposed c-api init() from SentryInitialization.cs, hard-coded DSN, got an issue to sentry.io - https://sentry.io/organizations/sentry-sdks/issues/3083605658/?project=5439417&query=is%3Aunresolved
* [x] bridge code (expose as the necessary objective-c code as C-APIs so we can call from C#)
* [x] initialize (at runtime), after Unity is loaded
* [x] Integration test in CI - after #686 is merged
* [x] Symbol upload
* [ ] anyone some ideas how to improve the [App section](https://sentry.io/organizations/sentry-sdks/issues/3235150078/?project=5439417&query=is%3Aunresolved#context-app) - e.g. "App Build", "Build Type" - it's likely set in [SentryCrashIntegration.m in sentry-cocoa](https://github.com/getsentry/sentry-cocoa/blob/aa03532ec9a5cd1e6816e14b3e6d10aebb1358cf/Sources/Sentry/SentryCrashIntegration.m#L243-L251) but not sure how to influence the content...

### Issue samples:
* https://sentry.io/organizations/sentry-sdks/issues/2677827744/?project=5439417&query=is%3Aunresolved
* https://sentry.io/organizations/sentry-sdks/issues/3235146941/?project=5439417&query=is%3Aunresolved
* https://sentry.io/organizations/sentry-sdks/issues/3235150078/?project=5439417&query=is%3Aunresolved

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,6 +3,5 @@ BasedOnStyle: WebKit
 IndentWidth: 4
 IndentPPDirectives: AfterHash
 ColumnLimit: 100
-AlwaysBreakAfterDefinitionReturnType: All
 PointerAlignment: Right
 ForEachMacros: ['SENTRY_WITH_SCOPE', 'SENTRY_WITH_SCOPE_MUT']
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,12 @@ jobs:
       target: iOS
       runsOn: macos-latest
 
+  macos-sdk:
+    uses: ./.github/workflows/sdk.yml
+    with:
+      target: macOS
+      runsOn: macos-latest
+
   windows-sdk:
     uses: ./.github/workflows/sdk.yml
     with:
@@ -88,6 +94,12 @@ jobs:
         with:
           name: iOS-sdk
           path: package-dev/Plugins/iOS
+          wait-timeout: 3600
+
+      - uses: vaind/download-artifact@989a39a417730897d098ab11c34e49ac4e13ed70
+        with:
+          name: macOS-sdk
+          path: package-dev/Plugins/macOS
           wait-timeout: 3600
 
       - uses: vaind/download-artifact@989a39a417730897d098ab11c34e49ac4e13ed70
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -311,11 +311,6 @@ jobs:
           New-Item -Path '${{ matrix.unity-config-path }}' -ItemType Directory
           Set-Content -Path '${{ matrix.unity-config-path }}services-config.json' -Value '${{ secrets.UNITY_LICENSE_SERVER_CONFIG }}'
 
-      # Step used to reduce the Unity path to avoid long path errors on Windows.
-      - name: Make symbolic link for Unity (Windows)
-        if: matrix.os == 'windows'
-        run: New-Item -ItemType SymbolicLink -Path "C:\${{ steps.env.outputs.unityVersion }}" -Target "C:\Program Files\Unity\Hub\Editor\${{ steps.env.outputs.unityVersion }}"
-
       - name: Create new Project
         run: ./test/Scripts.Integration.Test/integration-create-project.ps1 -UnityPath "${{ env.UNITY_PATH }}"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,7 +273,7 @@ jobs:
         run: ./test/Scripts.Tests/test-pack-contents.ps1
 
   desktop-smoke-test:
-    name: Run  ${{ matrix.os }} Unity ${{ matrix.unity-version }} Smoke Test
+    name: Run  ${{ matrix.os }} Smoke Test - ${{ matrix.unity-version }}
     runs-on: ${{ matrix.os }}-latest
     strategy:
       fail-fast: false
@@ -317,10 +317,10 @@ jobs:
         run: New-Item -ItemType SymbolicLink -Path "C:\${{ steps.env.outputs.unityVersion }}" -Target "C:\Program Files\Unity\Hub\Editor\${{ steps.env.outputs.unityVersion }}"
 
       - name: Create new Project
-        run: ./test/Scripts.Integration.Test/integration-create-project.ps1 "${{ env.UNITY_PATH }}"
+        run: ./test/Scripts.Integration.Test/integration-create-project.ps1 -UnityPath "${{ env.UNITY_PATH }}"
 
       - name: Build Standalone Player without Sentry SDK
-        run: ./test/Scripts.Integration.Test/integration-build-project.ps1 "${{ env.UNITY_PATH }}"
+        run: ./test/Scripts.Integration.Test/integration-build-project.ps1 -UnityPath "${{ env.UNITY_PATH }}"
 
       - name: Download UPM package
         uses: vaind/download-artifact@989a39a417730897d098ab11c34e49ac4e13ed70
@@ -333,10 +333,10 @@ jobs:
         run: ./test/Scripts.Integration.Test/integration-extract-package.ps1
 
       - name: Add Sentry to test project
-        run: ./test/Scripts.Integration.Test/integration-update-sentry.ps1 "${{ env.UNITY_PATH }}"
+        run: ./test/Scripts.Integration.Test/integration-update-sentry.ps1 -UnityPath "${{ env.UNITY_PATH }}"
 
       - name: Build Standalone Player Sentry SDK
-        run: ./test/Scripts.Integration.Test/integration-build-project.ps1 "${{ env.UNITY_PATH }}"
+        run: ./test/Scripts.Integration.Test/integration-build-project.ps1 -UnityPath "${{ env.UNITY_PATH }}"
 
       - name: Run Player - Smoke Test
         run: ./test/Scripts.Integration.Test/integration-run-smoke-test.ps1 -Smoke
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -272,16 +272,21 @@ jobs:
         # pwsh ./test/Scripts.Tests/test-pack-contents.ps1 accept
         run: ./test/Scripts.Tests/test-pack-contents.ps1
 
-  windows-smoke-test:
-    name: Run Windows Unity ${{ matrix.unity-version }} Smoke Test
-    runs-on: windows-latest
+  desktop-smoke-test:
+    name: Run  ${{ matrix.os }} Unity ${{ matrix.unity-version }} Smoke Test
+    runs-on: ${{ matrix.os }}-latest
     strategy:
       fail-fast: false
       matrix:
         unity-version: ['2019', '2020', '2021']
+        os: ['windows', 'macos']
         include:
-          - os: windows-latest
+          - os: windows
+            unity-modules: windows-il2cpp
             unity-config-path: C:/ProgramData/Unity/config/
+          - os: macos
+            unity-modules: mac-il2cpp
+            unity-config-path: /Library/Application Support/Unity/config/
     defaults:
       run:
         shell: pwsh
@@ -293,20 +298,11 @@ jobs:
         id: env
         run: echo "::set-output name=unityVersion::$(./scripts/ci-env.ps1 "unity${{ matrix.unity-version }}")"
 
-      # Caching the whole Unity installation would overflow the 10 GiB limit repositories have but we can cache the Hub
-      # to achieve some speedup here at a low size cost.
-      # See https://github.com/kuler90/setup-unity/blob/797ebd0cc9d0e0ca5bcd6cc280ec1a6e685fbf83/src/setup.js#L36
-      - name: Cache Unity Hub
-        uses: actions/cache@v2
-        with:
-          path: 'C:/Program Files/Unity Hub'
-          key: unity-hub|windows
-
       - name: Setup Unity
         uses: getsentry/setup-unity@46c2e082d98cc3a825a5b59038cb31705fe9ff56
         with:
           unity-version: ${{ steps.env.outputs.unityVersion }}
-          unity-modules: windows-il2cpp
+          unity-modules: ${{ matrix.unity-modules }}
 
       - run: echo "::add-mask::${{ secrets.LICENSE_SERVER_URL }}"
 
@@ -317,6 +313,7 @@ jobs:
 
       # Step used to reduce the Unity path to avoid long path errors on Windows.
       - name: Make symbolic link for Unity (Windows)
+        if: matrix.os == 'windows'
         run: New-Item -ItemType SymbolicLink -Path "C:\${{ steps.env.outputs.unityVersion }}" -Target "C:\Program Files\Unity\Hub\Editor\${{ steps.env.outputs.unityVersion }}"
 
       - name: Create new Project
--- a/package-dev/Plugins/iOS/SentryNativeBridge.m
+++ b/package-dev/Plugins/iOS/SentryNativeBridge.m
@@ -2,31 +2,51 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// must be here to match the native bridge API for macOS
-int SentryNativeBridgeInit() {
-    return 0; // this shouldn't be used so return "false"
+int SentryNativeBridgeLoadLibrary()
+{
+    // macOS only
 }
 
-int SentryNativeBridgeCrashedLastRun() {
-    return [SentrySDK crashedLastRun] ? 1 : 0;
+void *SentryNativeBridgeOptionsNew()
+{
+    // macOS only
 }
 
-void SentryNativeBridgeClose() {
-    [SentrySDK close];
+void SentryNativeBridgeOptionsSetString(void *options, const char *name, const char *value)
+{
+    // macOS only
 }
 
-void SentryNativeBridgeAddBreadcrumb(const char* timestamp, const char* message, const char* type, const char* category, int level) {
+void SentryNativeBridgeOptionsSetInt(void *options, const char *name, int32_t value)
+{
+    // macOS only
+}
+
+void SentryNativeBridgeStartWithOptions(void *options)
+{
+    // macOS only
+}
+
+int SentryNativeBridgeCrashedLastRun() { return [SentrySDK crashedLastRun] ? 1 : 0; }
+
+void SentryNativeBridgeClose() { [SentrySDK close]; }
+
+void SentryNativeBridgeAddBreadcrumb(
+    const char *timestamp, const char *message, const char *type, const char *category, int level)
+{
     if (timestamp == NULL && message == NULL && type == NULL && category == NULL) {
         return;
     }
 
-    [SentrySDK configureScope:^(SentryScope * scope) {
+    [SentrySDK configureScope:^(SentryScope *scope) {
         SentryBreadcrumb *breadcrumb = [[SentryBreadcrumb alloc] init];
 
         if (timestamp != NULL) {
             NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
             [dateFormatter setDateFormat:NSCalendarIdentifierISO8601];
-            breadcrumb.timestamp = [dateFormatter dateFromString:[NSString stringWithCString:timestamp encoding:NSUTF8StringEncoding]];
+            breadcrumb.timestamp =
+                [dateFormatter dateFromString:[NSString stringWithCString:timestamp
+                                                                 encoding:NSUTF8StringEncoding]];
         }
 
         if (message != NULL) {
@@ -38,7 +58,8 @@ void SentryNativeBridgeAddBreadcrumb(const char* timestamp, const char* message,
         }
 
         if (category != NULL) {
-            breadcrumb.category = [NSString stringWithCString:category encoding:NSUTF8StringEncoding];
+            breadcrumb.category = [NSString stringWithCString:category
+                                                     encoding:NSUTF8StringEncoding];
         }
 
         breadcrumb.level = level;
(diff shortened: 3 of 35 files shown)


.clang-format
```diff
@@ -3,6 +3,5 @@ BasedOnStyle: WebKit
 IndentWidth: 4
 IndentPPDirectives: AfterHash
 ColumnLimit: 100
-AlwaysBreakAfterDefinitionReturnType: All
```
vaind, line 6: wasn't respected so actually applying it would cause more changes in files (.m)

package-dev/Plugins/iOS/SentryNativeBridge.m
```diff
@@ -2,116 +2,181 @@
...
+            SentryBreadcrumb *breadcrumb = [[SentryBreadcrumb alloc]
+                initWithLevel:level
+                     category:(category ? [NSString stringWithUTF8String:category] : nil)];
+
+            if (timestamp != NULL) {
+                NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
+                [dateFormatter setDateFormat:NSCalendarIdentifierISO8601];
+                breadcrumb.timestamp =
+                    [dateFormatter dateFromString:[NSString stringWithUTF8String:timestamp]];
+                [dateFormatter release];
```
vaind, line 67: off-topic, fixing a memory leak here

bruno-garcia:
@philipphofmann or @brustolin could either of you take a look at the Objective-C code in here please? Just in case you have some suggestions on something that can be improved or could be an issue.

package-dev/Plugins/macOS/SentryNativeBridge.m
```diff
@@ -0,0 +1,246 @@
...
+            // everything above passed - mark as successfully loaded
+            loadStatus = 1;
+        } while (false);
+    }
+    return loadStatus;
+}
+
+const void *SentryNativeBridgeOptionsNew()
+{
+    return CFBridgingRetain([[NSMutableDictionary alloc] init]);
```
bruno-garcia, line 45: There's no ARC available? I haven't seen `CFBridgingRetain` used with ARC so wondering why we need this here

package-dev/Plugins/macOS/SentryNativeBridge.m
```diff
@@ -0,0 +1,246 @@
...
+            [breadcrumb setValue:[NSString stringWithUTF8String:type] forKey:@"type"];
+        }
+
+        if (category != NULL) {
+            [breadcrumb setValue:[NSString stringWithUTF8String:category] forKey:@"category"];
+        }
+
+        [breadcrumb setValue:[NSNumber numberWithInt:level] forKey:@"level"];
+
+        [scope performSelector:@selector(addBreadcrumb:) withObject:breadcrumb];
```
bruno-garcia, line 127: So no way we can add the header and use the actual types? I assume not since you went with this approach.

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -13,7 +13,7 @@
 using UnityEngine;
 using UnityEngine.Scripting;
 
-#if SENTRY_NATIVE_IOS
+#if SENTRY_NATIVE_COCOA
 using Sentry.Unity.iOS;
```
bruno-garcia, line 17: Should we change the namespace?

bruno-garcia (commented):
30% through, gotta get back to it

package-dev/Runtime/SentryInitialization.cs
vaind: Yes, but didn't want to do that in the same PR to avoid polluting the changelist, if it didn't show up just as "renames". I'll do that in a followup PR

package-dev/Plugins/macOS/SentryNativeBridge.m
vaind: CFBridgingRetain is used specifically with ARC - I had to change that because of some compilation issues. I'm glad if someone with good objective-c understanding has a look but to my limited knowledge, this should be OK - we want to take ownership of the pointer here, keeping the value in memory, later in `SentryNativeBridgeStartWithOptions` it's transferred back and left to ARC to free as needed.

package-dev/Plugins/macOS/SentryNativeBridge.m
vaind: I couldn't get it to work. That being said, maybe there's a way, I just don't know it :)

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -13,7 +13,7 @@ public static class BuildPostProcess
         [PostProcessBuild(1)]
         public static void OnPostProcessBuild(BuildTarget target, string executablePath)
         {
-            if (target is not (BuildTarget.StandaloneWindows or BuildTarget.StandaloneWindows64))
+            if (EditorUserBuildSettings.selectedBuildTargetGroup is not BuildTargetGroup.Standalone)
```
bitsandfoxes, line 16: That's a neat way to deal with the different Standalone plaforms!

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -53,16 +54,38 @@ public static void OnPostProcessBuild(BuildTarget target, string executablePath)
...
         }
 
-        private static void AddCrashHandler(IDiagnosticLogger logger, string projectDir)
+        private static bool IsEnabledForPlatform(BuildTarget target, SentryUnityOptions options) => target switch
         {
-            var crashpadPath = Path.GetFullPath(Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins",
-                "Windows", "Sentry", "crashpad_handler.exe"));
-            var targetPath = Path.Combine(projectDir, Path.GetFileName(crashpadPath));
-            logger.LogInfo("Copying the native crash handler '{0}' to the output directory", Path.GetFileName(crashpadPath));
+            BuildTarget.StandaloneWindows64 => options.WindowsNativeSupportEnabled,
```
bitsandfoxes, line 59:
```suggestion
            BuildTarget.StandaloneWindows => options.WindowsNativeSupportEnabled,
            BuildTarget.StandaloneWindows64 => options.WindowsNativeSupportEnabled,
```
Is this missing?

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -53,16 +54,38 @@ public static void OnPostProcessBuild(BuildTarget target, string executablePath)
...
+            BuildTarget.StandaloneOSX => options.MacosNativeSupportEnabled,
+            _ => false,
+        };
+
+        private static void AddCrashHandler(IDiagnosticLogger logger, BuildTarget target, string projectDir, string executableName)
+        {
+            string crashpadPath;
+            string targetPath;
+            if (target is BuildTarget.StandaloneWindows64)
```
bitsandfoxes, line 68:
```suggestion
            if (target is BuildTarget.StandaloneWindows or BuildTarget.StandaloneWindows64)
```

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -90,13 +113,18 @@ private static void UploadDebugSymbols(IDiagnosticLogger logger, string projectD
...
 
             addPath(executableName);
-            addPath("GameAssembly.dll");
-            addPath("UnityPlayer.dll");
             addPath(Path.GetFileNameWithoutExtension(executableName) + "_BackUpThisFolder_ButDontShipItWithYourGame");
-            addPath(Path.GetFileNameWithoutExtension(executableName) + "_Data/Plugins/x86_64/sentry.dll");
-
-            // Note: using Path.GetFullPath as suggested by https://docs.unity3d.com/Manual/upm-assets.html
-            addPath(Path.GetFullPath($"Packages/{SentryPackageInfo.GetName()}/Plugins/Windows/Sentry/sentry.pdb"));
+            if (target is BuildTarget.StandaloneWindows64)
```
bitsandfoxes, line 117:
```suggestion
            if (target is BuildTarget.StandaloneWindows or BuildTarget.StandaloneWindows64)
```

src/Sentry.Unity.iOS/SentryNative.cs
```diff
@@ -0,0 +1,63 @@
+using Sentry.Extensibility;
+using Sentry.Unity.Integrations;
+using UnityEngine;
+
+namespace Sentry.Unity.iOS
+{
+    /// <summary>
+    /// Access to the Sentry native support on iOS/macOS.
+    /// </summary>
+    public static class SentryNativeCocoa
```
bitsandfoxes, line 10: When fixing the namespace we should align file name with class name.

test/Sentry.Unity.iOS.Tests/SentryNativeIosTests.cs
```diff
@@ -1,46 +1,48 @@
+using System;
 using NUnit.Framework;
+using UnityEngine;
 
 namespace Sentry.Unity.iOS.Tests
 {
-    public class SentryNativeIosTests
+    public class SentryNativeCocoaTests
```
bitsandfoxes, line 7: Consider renaming the file in the follow-up namespace-PR.

.github/workflows/ci.yml
```diff
@@ -303,15 +311,11 @@ jobs:
           New-Item -Path '${{ matrix.unity-config-path }}' -ItemType Directory
           Set-Content -Path '${{ matrix.unity-config-path }}services-config.json' -Value '${{ secrets.UNITY_LICENSE_SERVER_CONFIG }}'
 
-      # Step used to reduce the Unity path to avoid long path errors on Windows.
-      - name: Make symbolic link for Unity (Windows)
-        run: New-Item -ItemType SymbolicLink -Path "C:\${{ steps.env.outputs.unityVersion }}" -Target "C:\Program Files\Unity\Hub\Editor\${{ steps.env.outputs.unityVersion }}"
```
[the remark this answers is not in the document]
vaind: 1. it was a temporary fix AFAICT - https://github.com/getsentry/sentry-unity/pull/423
2. it wasn't even used anymore - originally it was in the `build` job and was supposed to stay there - here, it was completely ignored because we passed the full unity path anyway.

src/Sentry.Unity.Editor/ConfigurationWindow/AdvancedTab.cs
```diff
@@ -59,9 +59,12 @@ internal static void Display(ScriptableSentryUnityOptions options)
                 options.AndroidNativeSupportEnabled);
 
             options.WindowsNativeSupportEnabled = EditorGUILayout.Toggle(
-                new GUIContent("Windows Native Support", "Whether to enable Native Windows support to " +
-                                                         "capture errors written in languages such as C and C++."),
+                new GUIContent("Windows Native Support", "Whether to enable native crashes support on Windows."),
```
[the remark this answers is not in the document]
vaind: it's not really a nitpick - you're suggesting we put back the part of the sentence I've dropped, about errors from C and C++ - the thing is, we have no idea what language the error is coming from, it's just a native crash...

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: no, we don't support 32-bit windows. It was there originally by mistake

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: no 32-bit support

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
vaind: no 32-bit support

brustolin:
The Objective-c code looks good to me. I would only suggest, instead of calling performSelector multiple times with the same selector, create a helper function to do it. Then instead of using @try-catch, you could check with the object can perform the given selector like so: 

```objc
if ([SentrySDK respondsToSelector:@selector(configureScope:)] )
```

src/Sentry.Unity.Editor/ConfigurationWindow/AdvancedTab.cs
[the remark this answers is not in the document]
vaind: what about 

```suggestion
                new GUIContent("Windows Native Support", "Whether to enable native (C, C++, ...) crash support on Windows."),
```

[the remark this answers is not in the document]
bruno-garcia: I think it's fine to leave the C/C++ off in this case to simplify things

philipphofmann (commented):
I'm not sure if we chose the right approach here. Maybe @vaind it would be best to pair up on this via a call and look into the code.

package-dev/Plugins/macOS/SentryNativeBridge.m
```diff
@@ -0,0 +1,220 @@
...
+            }
+
+            LOAD_CLASS_OR_BREAK(SentrySDK)
+            LOAD_CLASS_OR_BREAK(SentryScope)
+            LOAD_CLASS_OR_BREAK(SentryBreadcrumb)
+            LOAD_CLASS_OR_BREAK(SentryUser)
+
+            // everything above passed - mark as successfully loaded
+            loadStatus = 1;
+        } while (false);
```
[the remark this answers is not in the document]
vaind: Just a pattern I sometimes use if I have multiple guards (checks) - these `break` on failure. There are multiple ways to write this of course...

src/Sentry.Unity.iOS/SentryCocoaBridgeProxy.cs
```diff
@@ -11,10 +13,74 @@ namespace Sentry.Unity.iOS
...
-        [DllImport("__Internal")]
+        public static bool Init(SentryUnityOptions options)
+        {
+            // Note: used on macOS only
+            if (LoadLibrary() != 1)
+            {
+                return false;
+            }
+
+            var cOptions = OptionsNew();
```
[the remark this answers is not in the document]
vaind: I couldn't find a reasonable way to do that. There's no automatic marshalling afaict and I  really don't want to mess with the underlying structures trying to marshall it manually. The implemented solution is a simple c-api wrapper that should hopefully be straightforward to reason about. But I'm more than happy to accept suggestions if you know a better way to implement this - c# and objective-c side.

package-dev/Plugins/macOS/SentryNativeBridge.m
```diff
@@ -0,0 +1,220 @@
...
+
+static Class SentrySDK;
+static Class SentryScope;
+static Class SentryBreadcrumb;
+static Class SentryUser;
+
+#define LOAD_CLASS_OR_BREAK(name)                                                                  \
+    name = (__bridge Class)dlsym(dylib, "OBJC_CLASS_$_" #name);                                    \
+    if (!name) {                                                                                   \
+        NSLog(@"Sentry (native bridge): Couldn't load %@ class from the dynamic library", name);   \
```
Swatinem, line 14: what does this print when `name` happens to be null?

Swatinem (approved):
lgtm in general, though I have some open questions

package-dev/Plugins/macOS/SentryNativeBridge.m
vaind: well spotted, it was supposed to be `#name` (stringified). 

I've changed this code because xcode didn't like the standard constant string concatenation - `"...couldn't load " #name " class..."` - saying it should be a param... When I made it a param I messed it up and used `name` instead of `#name`... thanks

src/Sentry.Unity.iOS/SentryCocoaBridgeProxy.cs
```diff
@@ -11,10 +13,74 @@ namespace Sentry.Unity.iOS
...
+            options.DiagnosticLogger?.LogDebug("Setting SendDefaultPii: {0}", options.SendDefaultPii);
+            OptionsSetInt(cOptions, "sendDefaultPii", options.SendDefaultPii ? 1 : 0);
+
+            options.DiagnosticLogger?.LogDebug("Setting MaxBreadcrumbs: {0}", options.MaxBreadcrumbs);
+            OptionsSetInt(cOptions, "maxBreadcrumbs", options.MaxBreadcrumbs);
+
+            options.DiagnosticLogger?.LogDebug("Setting MaxCacheItems: {0}", options.MaxCacheItems);
+            OptionsSetInt(cOptions, "maxCacheItems", options.MaxCacheItems);
+
+            StartWithOptions(cOptions);
```
philipphofmann, line 61: `m`: I think we could add a method wrapping a call to `CFRelease` for the `cOptions`, as we don't need the dictionary anymore after starting the SDK.

package-dev/Plugins/macOS/SentryNativeBridge.m
```diff
@@ -0,0 +1,220 @@
...
+
+        if (ipAddress != NULL) {
+            [user setValue:[NSString stringWithUTF8String:ipAddress] forKey:@"ipAddress"];
+        }
+
+        if (username != NULL) {
+            [user setValue:[NSString stringWithUTF8String:username] forKey:@"username"];
+        }
+
+        [scope setUser:user];
```
philipphofmann, line 213: `m`: I'm a bit confused. Above we use `performSelector` and here we can call the method directly. Why is that?

package-dev/Plugins/macOS/SentryNativeBridge.m
```diff
@@ -0,0 +1,220 @@
...
+    [SentrySDK performSelector:@selector(startWithOptions:) withObject:dictOptions];
+}
+
+void SentryConfigureScope(void (^callback)(id))
+{
+    @try {
+        [SentrySDK performSelector:@selector(configureScope:) withObject:callback];
+    } @catch (NSException *exception) {
+        NSLog(@"Sentry (bridge): failed to configure scope: %@", exception.reason);
+    }
```
philipphofmann, line 72: `m`: Why do we wrap these calls with try catch?

philipphofmann (commented):
I added a few comments.

package-dev/Plugins/macOS/SentryNativeBridge.m
```diff
@@ -0,0 +1,220 @@
...
+{
+    if (email == NULL && userId == NULL && ipAddress == NULL && username == NULL) {
+        return;
+    }
+
+    SentryConfigureScope(^(id scope) {
+        id user = [[SentryUser alloc] init];
+
+        if (email != NULL) {
+            [user setValue:[NSString stringWithUTF8String:email] forKey:@"email"];
```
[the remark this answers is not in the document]
vaind: described here https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/KeyValueCoding/SearchImplementation.html#//apple_ref/doc/uid/20000955

### Search Pattern for the Basic Setter
The default implementation of setValue:forKey:, given key and value parameters as input, attempts to set a property named key to value (or, for non-object properties, the unwrapped version of value, as described in [Representing Non-Object Values](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/KeyValueCoding/DataTypes.html#//apple_ref/doc/uid/20002171-BAJEAIEE)) inside the object receiving the call, using the following procedure:

* Look for the first accessor named set<Key>: or _set<Key>, in that order. If found, invoke it with the input value (or unwrapped value, as needed) and finish.

* If no simple accessor is found, and if the class method accessInstanceVariablesDirectly returns YES, **look for an instance variable with a name like _<key>, _is<Key>, <key>, or is<Key>, in that order. If found, set the variable directly with the input value (or unwrapped value) and finish.**

* Upon finding no accessor or instance variable, invoke setValue:forUndefinedKey:. This raises an exception by default, but a subclass of NSObject may provide key-specific behavior.

package-dev/Plugins/macOS/SentryNativeBridge.m
vaind: it gives a lot of warnings unless the argument is nil or `id`:

`instance method '-setExtraValue:forKey:' not found (return type defaults to 'id') [-Wobjc-method-access]`

with some Unity version, these warnings actually caused the build to fail.

I'll also change setUser to performSelector and add this comment to the code.

package-dev/Plugins/macOS/SentryNativeBridge.m
vaind: setValue forKey: may throw if the property is not found; same for performSelector I think - even though this shouldn't happen, better not take the chance of letting an unhandled exception while setting error info - it would just crash the app immediately.

src/Sentry.Unity.iOS/SentryCocoaBridgeProxy.cs
vaind: the ownership is returned by `__bridge_transfer` here:

```objective-c
void SentryNativeBridgeStartWithOptions(const void *options)
{
    NSMutableDictionary *dictOptions = (__bridge_transfer NSMutableDictionary *)options;
    [SentrySDK performSelector:@selector(startWithOptions:) withObject:dictOptions];
}
```

and AFAIU the object will be freed after startWithOptions: is done with it. Therefore we can't do it again manually with CFRelease() or it would be a double-free. But that's just my limited understanding of memory management in objective-c...

vaind:
Thanks everyone for the reviews. 

@bruno-garcia please merge if you're done reviewing

bruno-garcia:
This will get merged today unless we get some tips from an Objective-C wizard

bruno-garcia:
This will get merged today unless we get some tips from an Objective-C wizard

philipphofmann:
> This will get merged today unless we get some tips from an Objective-C wizard

I wouldn't call myself a wizard. Just a normal dev 😁

package-dev/Plugins/macOS/SentryNativeBridge.m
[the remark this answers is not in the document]
philipphofmann: Ups forgot about that. Usually, I don't have to use this, as I'm not so limited.

src/Sentry.Unity.iOS/SentryCocoaBridgeProxy.cs
philipphofmann: Thanks for the detailed explanation. Never had to do manual ARC yet.

philipphofmann (approved):
Thanks, @vaind, for your patience and replying to all my questions. LGTM 🚀

bruno-garcia (approved):

PR #712, opened 2022-04-27
ref: Scriptable Object loading

bitsandfoxes:
The goal is to unify the way we load our ScriptableObjects within the editor and refactor things out of the Config Window that don't belong there in the first place.

#skip-changelog


test/Sentry.Unity.Editor.Tests/SentryScriptableObjectTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        [SetUp]
+        public void Setup() => _tempPath = Path.Combine("Assets", Guid.NewGuid().ToString(), "TestOptions.asset");
+
+        [TearDown]
+        public void TearDown() => AssetDatabase.DeleteAsset(Path.GetDirectoryName(_tempPath));
+
+        [Test]
+        public void Load_ScriptableSentryUnityOptionsDoesNotExist_ReturnsNewObject()
+        {
+            SentryScriptableObject.Load<ScriptableSentryUnityOptions>(_tempPath);
```
bruno-garcia, line 21: Best to take this temp path here, so there's no chance of another test (that ran before on concurrently) to share that resource. Maybe the file already existed?

test/Sentry.Unity.Editor.Tests/SentryScriptableObjectTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            Assert.IsTrue(File.Exists(_tempPath));
+        }
+
+        [Test]
+        public void Load_ScriptableSentryUnityOptionsAlreadyExist_ReturnsObject()
```
bruno-garcia, line 27: Name doesn't match the assert

test/Sentry.Unity.Editor.Tests/SentryScriptableObjectTests.cs
bitsandfoxes: The `[Setup]` creates a new path containing a Guid. Does that suffice?

bruno-garcia (approved):

PR #723, opened 2022-05-03
ref: SetupLogging as extension 

bitsandfoxes:
Mirroring the .NET SDK by moving `TryAttachLogger` to options extension `SetupLogging`.

#skip-changelog

### DIFF
--- a/test/Sentry.Unity.Tests/SentryUnityOptionsExtensionsTests.cs
+++ b/test/Sentry.Unity.Tests/SentryUnityOptionsExtensionsTests.cs
@@ -92,9 +92,11 @@ namespace Sentry.Unity.Tests
         }
 
         [Test]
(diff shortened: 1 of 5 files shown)


test/Sentry.Unity.Tests/SentryUnityOptionsExtensionsTests.cs
```diff
@@ -88,5 +90,27 @@ public void ShouldInitializeSdk_NotCaptureInEditorAndApplicationIsEditor_Returns
 
             Assert.IsFalse(shouldInitialize);
         }
+
+        [Test]
+        public void SetupLogging_DebugTrue_SetsUnityLogger()
```
bruno-garcia, line 95: Could have a test that checks the code path of having already a loggers set. With Debug true and false.

bruno-garcia (approved):

PR #728, opened 2022-05-04
Feat: Automatic user IDs on native crashes & .NET events

### LINKED ISSUE #606: Align the default `user-id` between the C# and native layers
SDKs usually have some default 'installation-id' as a User. When capturing from C# and Native those don't match.

Relates to: https://github.com/getsentry/sentry-cocoa/issues/1554

vaind:
closes #606 

* [x] android -> [dotnet](https://sentry.io/organizations/sentry-sdks/issues/3013122412/events/24772b6605694aeaa42378755ae9f31b/?project=5439417)
* [x] iOS-> [dotnet](https://sentry.io/organizations/sentry-sdks/issues/2771507099/events/c74dfdc14c5444fda5a37287f59e62cb/?project=5439417)
* [x] macOS -> [dotnet](https://sentry.io/organizations/sentry-sdks/issues/2771507099/events/23e47e291dc34cfeb2b62984693b4085/?project=5439417)
* [x] [Unity](https://docs.unity3d.com/ScriptReference/Analytics.AnalyticsSessionInfo-userId.html) -> [dotnet](https://sentry.io/organizations/sentry-sdks/issues/3013122412/events/2d5c0c8bdf0a46059dc01a69bfcf4799/?project=5439417) -> [Windows](https://sentry.io/organizations/sentry-sdks/issues/3250798745/events/3a5d39fd411f4a1e90ec3963349b8b86/?project=5439417)
* [x]  Unity -> [dotnet](https://sentry.io/organizations/sentry-sdks/issues/2771507099/events/59843f407c1048cbaa8d95efe0be85b1/?project=5439417) -> [Linux](https://sentry.io/organizations/sentry-sdks/issues/3262191687/events/26b52e0cc279401d4f1ee39d28707288/?project=5439417)
* [x]  Unity -> [dotnet](https://sentry.io/organizations/sentry-sdks/issues/3013122412/events/061c6bca5932424288ccdae07ca8bab3/?project=5439417) on WebGL. Notes:
  -  the ID seems to persist across tabs & browser sessions, likely a cookie or local storage based, so pretty useful and in line with other platforms
  - opening in a different browser or a private tab gives a new ID, which is also OK

### DIFF
--- a/src/Sentry.Unity/Sentry.Unity.csproj
+++ b/src/Sentry.Unity/Sentry.Unity.csproj
@@ -27,4 +27,9 @@
       Overwrite="true"/>
   </Target>
 
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
+      <_Parameter1>Sentry.Unity.Native</_Parameter1>
+    </AssemblyAttribute>
+  </ItemGroup>
 </Project>
--- a/src/Sentry.Unity/SentryUnityOptions.cs
+++ b/src/Sentry.Unity/SentryUnityOptions.cs
@@ -110,8 +110,9 @@ namespace Sentry.Unity
         /// Whether the SDK should add native support for Linux
         /// </summary>
         public bool LinuxNativeSupportEnabled { get; set; } = true;
-        // Initialized by native SDK binding code and to set the User.ID in sentry-dotnet.
 
+
+        // Initialized by native SDK binding code to set the User.ID in .NET (UnityEventProcessor).
         internal string? DefaultUserId;
 
         public SentryUnityOptions() : this(ApplicationAdapter.Instance, false)
--- a/src/Sentry.Unity/SentryUnityOptions.cs
+++ b/src/Sentry.Unity/SentryUnityOptions.cs
@@ -1,6 +1,7 @@
 using System.Linq;
 using System.Text.RegularExpressions;
 using Sentry.Unity.Integrations;
+using Sentry.Extensibility;
 using CompressionLevel = System.IO.Compression.CompressionLevel;
 
 namespace Sentry.Unity
@@ -113,7 +114,23 @@ namespace Sentry.Unity
 
 
         // Initialized by native SDK binding code to set the User.ID in .NET (UnityEventProcessor).
-        internal string? DefaultUserId;
+        internal string? _defaultUserId;
+        internal string? DefaultUserId
+        {
+            get => _defaultUserId;
+            set
+            {
+                _defaultUserId = value;
+                if (_defaultUserId is null)
+                {
+                    DiagnosticLogger?.LogWarning("Couldn't set the default user ID - the value is NULL.");
+                }
+                else
+                {
+                    DiagnosticLogger?.LogDebug("Setting '{0}' as the default user ID.", _defaultUserId);
+                }
+            }
+        }
 
         public SentryUnityOptions() : this(ApplicationAdapter.Instance, false)
         {
--- a/src/Sentry.Unity/SentryUnityOptions.cs
+++ b/src/Sentry.Unity/SentryUnityOptions.cs
@@ -110,6 +110,9 @@ namespace Sentry.Unity
         /// Whether the SDK should add native support for Linux
         /// </summary>
         public bool LinuxNativeSupportEnabled { get; set; } = true;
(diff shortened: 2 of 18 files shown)


src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -111,6 +112,26 @@ public new CompressionLevelWithAuto RequestBodyCompressionLevel
...
+            {
+                _defaultUserId = value;
+                if (_defaultUserId is null)
+                {
+                    DiagnosticLogger?.LogWarning("Couldn't set the default user ID - the value is NULL.");
+                }
+                else
+                {
+                    DiagnosticLogger?.LogDebug("Setting '{0}' as the default user ID.", _defaultUserId);
+                }
```
bitsandfoxes, line 131:
```suggestion
                if (value is null)
                {
                    DiagnosticLogger?.LogWarning("Couldn't set the default user ID - the value is NULL.");
                }
                else
                {
                    _defaultUserId = value;
                    DiagnosticLogger?.LogDebug("Setting '{0}' as the default user ID.", _defaultUserId);
                }
```
Should this not set the id if it logs that it can't?

bitsandfoxes (approved):
So in summary: 
- For Android, Cocoa (iOS and macOS) we fetch the ID from the native SDKs (which is great because they initialize first)
- For Native (Windows & Linux) we set the ID from the C# layer.

This is great!

src/Sentry.Unity/Sentry.Unity.csproj
```diff
@@ -27,4 +27,9 @@
       Overwrite="true"/>
   </Target>
 
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
```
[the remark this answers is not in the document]
bruno-garcia: No, this is the same as adding `[assembly: InternalsVisibleToAttribute(..)` in C#. Just another way.

src/Sentry.Unity/SentryUnityOptions.cs
bruno-garcia: Is the goal to reset to null as it'll be needed later to "unset"?

src/Sentry.Unity/WebGL/SentryWebGL.cs
```diff
@@ -36,6 +37,9 @@ public static void Configure(SentryUnityOptions options)
                 options.DiagnosticLogger?.LogWarning("Attaching screenshots on WebGL is disabled - " +
                     "it currently produces blank screenshots mid-frame.");
             }
+
+            // Use AnalyticsSessionInfo.userId as the default UserID in native & dotnet
+            options.DefaultUserId = AnalyticsSessionInfo.userId;
```
bruno-garcia, line 42: If this is an ID that can be used to link the user across apps/services, we shouldn't be using it here.
We purposely on some GUIDs we create so that we can have some consistency in detecting "number ofusers impacted" but the goal is never to do actual user tracking. For that reason we've dropped "Android ID" also, for example

bruno-garcia: JS will default to the connection IP Address, and there's no native so we might be better off without anything here

src/Sentry.Unity/SentryUnityOptions.cs
vaind: > Should this not set the id if it logs that it can't?

this is a setter - doesn't feel right to ignore the value being passed (i.e. not set)

src/Sentry.Unity/WebGL/SentryWebGL.cs
vaind: 1. we don't have sentry-js (yet)
2. therefore setting this here only affects c# events
3. I'm setting it _here_ because that's what feeled right - each "native" integration defines the source for the DefaultUserId value
4. > If this is an ID that can be used to link the user across apps/services, we shouldn't be using it here.

    I really don't think so. It's an installation ID.
    
   > https://docs.unity3d.com/ScriptReference/Analytics.AnalyticsSessionInfo-userId.html 
   > A random GUID uniquely identifying sessions played on the same instance of your game or app.
   > Unity generates the userId GUID immediately after installation. The userId value is only regenerated if the user re-installs the game.

vaind:
since this is already approved I'm merging to avoid conflicts in #747 - if we need to make changes based on those unresolved conversations, let's do them in the other PR.

PR #730, opened 2022-05-05
fix: Android debug symbols 

bitsandfoxes:
Starting with 2021.2 the caching directory changed from `Temp` to `Library`.

Additionally, the gradle project persists across builds, so the "append symbol upload" functionality must be reentrant and removed in case the user opts out at a later point.

### DIFF
--- a/src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
+++ b/src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
@@ -11,17 +11,15 @@ namespace Sentry.Unity.Editor.Android
     {
         private readonly IDiagnosticLogger _logger;
 
-        private readonly bool _isNewBuildingBacked;
-
-        private const string RelativeBuildOutputPathOld = "Temp/StagingArea/symbols";
-        private const string RelativeGradlePathOld = "Temp/gradleOut";
-        private const string RelativeBuildOutputPathNew = "Library/Bee/artifacts/Android";
-        private const string RelativeAndroidPathNew = "Library/Android";
+        internal const string RelativeBuildOutputPathOld = "Temp/StagingArea/symbols";
+        internal const string RelativeGradlePathOld = "Temp/gradleOut";
+        internal const string RelativeBuildOutputPathNew = "Library/Bee/artifacts/Android";
+        internal const string RelativeAndroidPathNew = "Library/Android";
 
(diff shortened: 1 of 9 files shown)


src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
```diff
@@ -1,59 +1,109 @@
...
+
+            _logger = logger;
+
+            _unityProjectPath = unityProjectPath;
+            _gradleProjectPath = gradleProjectPath;
 
-                return copiedSymbolsPath;
+            // Starting from 2021.2 Unity caches the build output inside 'Library' instead of 'Temp'
+            if (application.UnityVersion.StartsWith("2022")
+                || application.UnityVersion.StartsWith("2021.3")
```
bruno-garcia, line 39: This will break when 2023 comes out?
Best to try to have this inverted like: if it's not 2018, 2019, 2020 or 2021.1, then it's the new stuff.

.NET Has a class that handled Version (int he format \d\.\d\.\d\.\d\ though but might work here. https://docs.microsoft.com/en-us/dotnet/api/system.version?view=net-6.0

So you can do something like `var version = new Version(application.UnityVersion)` and then compare:
`version >= new Version("2021.1")`

Otherwise you'd need to split by `.` and check the individual parts, perhaps easier.

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
```diff
@@ -63,28 +113,58 @@ public static void AppendUploadToGradleFile(string sentryCliPath, string gradleP
...
-            executable ""{sentryCliPath}""
-            args = [""upload-dif"", ""{symbolsDirectoryPath}""]
+            args = [""upload-dif"", {symbolPathArgument}]
         }}
     }}
 }}");
         }
 
-        internal static void CopySymbols(string sourcePath, string targetPath)
+        public void RemoveUploadFromGradleFile()
```
bruno-garcia, line 122: This seems pretty fragile

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
bitsandfoxes: TIL. Thanks!

vaind (approved):

Issue #60, opened 2021-03-01
Unity Log level


bruno-garcia:
This issued describes this todo:

https://github.com/getsentry/sentry-unity/blob/d30c7765172972162db86bac1ab4414a6aad75bf/src/Sentry.Unity/SentryInitialization.cs#L158-L159

We should have a configurable minimum level for what becomes a breadcrumb and what becomes an event.
Similar to what other logging integrations do.

Example:

bruno-garcia:
With such a feature, we could configure the SDK to send an invite for what the warning seeing here as a crumb:

![image](https://user-images.githubusercontent.com/1633368/131217989-703fda79-fbab-453b-a036-877e5e9343d9.png)

oddgames-david:
+1 would like to limit reporting only to exceptions, errors still to be in breadcrumbs though.

mattjohnsonpint:
This should mirror the behavior we have with `MinimumBreadcrumbLevel` and `MinimumEventLevel` in Sentry.Extensions.Logging.

oddgames-david:
Still want to see this. Unity developers rely on many 3rd party sdk's and generally error log is used as a "serious" log when something has gone wrong but hasn't actually broken the app and sometimes we cant change the code (or at least shouldn't have to since we'll need to update it later and make those changes again).

bitsandfoxes:
You basically just want to be able to automatically separate exceptions from logged errors. I think everything should be in place within the `UnityLogHandlerIntegration` to make that work. The SDK won't be able to distinguish between LogException and an actual exception tho.

bitsandfoxes:
The "`loglevel` to breadcrumb" is now individually selectable even through the config window.
![Image](https://github.com/user-attachments/assets/8d5db59c-a12c-4a72-b4bb-59069ccffeae)

PR #731, opened 2022-05-06
ref: Use the global LogHandler to capture unhandled Exceptions

### LINKED ISSUE #579: LogException is not splitting Inner exceptions
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)

Which version of the SDK? Unity (Mono - Android)

Which version of Unity? 2019.4.34f1

Tested on Android

### Steps to Reproduce

Run sample code below and you'll notice something like 
image

### Expected Result

A StackTrace similar to the one captured by CaptureException.

### Actual Result

Event captured by LogException
https://sentry.io/organizations/sentry-sdks/issues/3037507922/?query=is%3Aunresolved&statsPeriod=1h

Event captured by CaptureException
https://sentry.io/organizations/sentry-sdks/issues/3037473558/?_allp=1&query=is%3Aunresolved&statsPeriod=1h

### Any logs or screenshots
Sample code
```csharp
        try
        {
            Task.Run(async () =>
            {
                var  client = new HttpClient();
                (await client.GetAsync("https://expired.badssl.com/")).EnsureSuccessStatusCode();
            }).Wait();
            
        }
        catch (Exception e)
        {
            Debug.LogException(e, this);
//          VS            
            SentrySdk.CaptureException(e);
        }
```

### LINKED ISSUE #599: Document the lack of Exception on errors captured through the Unity logger
Parts of the docs point users to instances of `Exception` like `SentryEvent.Exception` accessed on BeforeSend snippets:

https://docs.sentry.io/platforms/dotnet/configuration/filtering/#using-platformidentifier-namebefore-send-

And 
* `AddExceptionFilter`: https://github.com/getsentry/sentry-dotnet/blob/35b72f296ea000bfc5ee31a7264701cffff39437/src/Sentry/SentryOptionsExtensions.cs#L106-L109
* `ISentryEventExceptionProcessor`: https://github.com/getsentry/sentry-dotnet/blob/35b72f296ea000bfc5ee31a7264701cffff39437/src/Sentry/Extensibility/ISentryEventExceptionProcessor.cs#L8

The docs need to clarify that exceptions captured through the Unity logger API don't give us an Exception instance and therefore those hooks can't be used to filter out those.

We need to replace the code snippets to use equivalent APIs that work on Unity. e.g:

```csharp
options.BeforeSend = sentryEvent =>
{
    if (sentryEvent.SentryExceptions.Any(e =>
      e.Message.Contains("Noisy Exception"))
    {
        return null; // Don't send this event to Sentry
    }

    sentryEvent.SetTag("myTag, "myValue");
    return sentryEvent;
};
```

Alternatively, document that exceptions coming from Unity logger are in fact `UnityLogException` and how to inspect values from those to make decisions.

[As suggested on Discord](https://discord.com/channels/621778831602221064/621783582628642822/947881309714399252), we can expose the original stacktrace:

```csharp
public override string StackTrace => LogStackTrace;
```

Swatinem:
Turns out the Unity Runtime actually "handles" uncaught Exceptions from scripts and hands them off to its global LogHandler. This is presumably the reason why AppDomain.UnhandledException is never being invoked for them.

We can capture these Exceptions (and in the future enhance them with native instruction addrs) by shimming the global LogHandler.

This completely replaces and removes the Application.OnLogMessageReceived integration and the need to parse exceptions from string.

Resolves #579
Resolves #599


src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -1,54 +1,102 @@
...
+
+            // NOTE: This might not be entirely true, as a user could as well call `Debug.LogException`
+            // and expect a handled exception but it is not possible for us to differentiate
+            // https://docs.sentry.io/platforms/unity/troubleshooting/#unhandled-exceptions---debuglogexception
+            exception.Data[Mechanism.HandledKey] = false;
+            exception.Data[Mechanism.MechanismKey] = "Unity.LogException";
+            _ = _hub.CaptureException(exception);
+
+            // So the next event includes this error as a breadcrumb
+            _hub.AddBreadcrumb(message: exception.GetType() + ": " + exception.Message, category: "unity.logger", level: BreadcrumbLevel.Error);
```
bruno-garcia, line 61: Prefer interpolation:

```suggestion
            _hub.AddBreadcrumb(message: $"{exception.GetType()}: {exception.Message}", category: "unity.logger", level: BreadcrumbLevel.Error);
```

Here are some examples how we do this elsewhere:

https://github.com/getsentry/sentry-dotnet/blob/48a78fba69683cc34125d28e20238b27fd9f199d/src/Sentry.Extensions.Logging/SentryLogger.cs#L106-L126

https://github.com/getsentry/sentry-dotnet/blob/48a78fba69683cc34125d28e20238b27fd9f199d/src/Sentry.Serilog/SentrySink.cs#L107-L129

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -1,54 +1,102 @@
...
 
-        // Internal for testability
-        internal void OnLogMessageReceived(string? condition, string? stackTrace, LogType type)
+        public void LogException(Exception exception, UnityEngine.Object context)
+        {
+            CaptureException(exception, context);
+            _unityLogHandler.LogException(exception, context);
+        }
+
+        internal void CaptureException(Exception exception, UnityEngine.Object? context)
```
bruno-garcia, line 44: what's in `context`? Anything useful to add to the event or crumb?

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -1,54 +1,102 @@
...
+            }
+
+            // We're not capturing SDK internal logs
+            if (logMessage.StartsWith(UnityLogger.LogPrefix, StringComparison.Ordinal))
+            {
+                // TODO: Maybe color Sentry internal logs (highlight 'Sentry'?)
+                return;
+            }
+
+            if (_hub?.IsEnabled is not true)
```
bruno-garcia, line 92: What about this as the first line in the method?

src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -287,14 +287,6 @@ internal class UnityEventExceptionProcessor : ISentryEventExceptionProcessor
     {
         public void Process(Exception exception, SentryEvent sentryEvent)
         {
-            if (exception is UnityLogException ule)
```
bruno-garcia, line 290: I guess the whole class can go now?

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
bitsandfoxes: That's the crazy awesome part: When you're calling `Debug.Log("My log message", this);` you can actually pass "whatever" in there as context (the MonoBehaviour in this case). Inside the editor this lets you click the log in the console and highlights the GameObject in the hierarchy. What that means for us is that we could grab all kinds of useful stuff from there.

- The name of the GameObject 
- Other components on that GameObject
- Child Count
- e.t.c.

I'm just not sure what to grab. Probably the name.

src/Sentry.Unity/UnityEventProcessor.cs
Swatinem: `UnityLogException` is gone.
If you mean the `UnityEventExceptionProcessor`, I would rather keep it here to avoid too much churn with #683 which will put its code into this class.

vaind (approved):
LGTM with some questions. Also maybe add links to issues on sentry.io for before and after the change? If there's any change in the context

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -1,54 +1,102 @@
...
+            _ = _hub.CaptureException(exception);
+
+            // So the next event includes this error as a breadcrumb
+            _hub.AddBreadcrumb(message: $"{exception.GetType()}: {exception.Message}", category: "unity.logger", level: BreadcrumbLevel.Error);
+        }
+
+        public void LogFormat(LogType logType, UnityEngine.Object? context, string format, params object[] args)
+        {
+            CaptureLogFormat(logType, context, format, args);
+            _unityLogHandler.LogFormat(logType, context, format, args);
```
mattjohnsonpint, line 67:
```suggestion
            try
            {
                CaptureLogFormat(logType, context, format, args);
            }
            finally
            {
                // Always pass the exception on to Unity
                _unityLogHandler.LogFormat(logType, context, format, args);
            }
```

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -1,54 +1,102 @@
...
+            if (_hub?.IsEnabled is not true)
             {
                 return;
             }
 
-            if (_sentryOptions?.EnableLogDebouncing == true)
+            // TODO: Figure out if format {0} and args.length == 1 is guaranteed?
+            // TODO: Capture the context (i.e. grab the name if != null)
+
+            if (!format.Equals("{0}") || args.Length is > 1 or <= 0)
```
mattjohnsonpint, line 80: This will protect us from the uncertainty indicated by the TODO comment on line 77.

```suggestion
            if (args.Length == 0 || !format.Contains("{0}"))
```

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -63,41 +111,23 @@ internal void OnLogMessageReceived(string? condition, string? stackTrace, LogTyp
             }
 
             // TODO: to check against 'MinBreadcrumbLevel'
```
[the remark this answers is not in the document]
mattjohnsonpint: This is related to issue #60

test/Sentry.Unity.Tests/TestHttpClientHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+                if (!string.IsNullOrEmpty(eventRequest))
+                {
+                    Debug.Log(UnityLogger.LogPrefix + "TestHttpClientHandler returns event: \n" + eventRequest);
+                    return eventRequest;
+                }
+            }
+
+            // While within timeout: check every newly received request
+            var startTime = DateTime.Now;
+            while (DateTime.Now < startTime + timeout)
```
mattjohnsonpint, line 50: `DateTime` is not reliable for this sort of thing.  Use `System.Diagnostics.Stopwatch` instead.

```suggestion
            var stopwatch = Stopwatch.StartNew();
            while (stopwatch.Elapsed < timeout)
```

test/Sentry.Unity.Tests/TestHttpClientHandler.cs
```diff
@@ -0,0 +1,71 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
```
mattjohnsonpint, line 3: Add namespace to go with suggestion below

```suggestion
using System.Diagnostics;
using System.Net;
```

mattjohnsonpint (changes requested):

mattjohnsonpint:
Looks mostly good, except the changes I suggested.  But also, why is the Build - 2020 task failing?

test/Directory.Build.props
```diff
@@ -30,6 +30,18 @@
...
+<!--      referenced here and we just add the CoreModule it leads to 'ambiguous references' and 'Type exists in both -->
+<!--      'UnityEngine.CoreModule and UnityEngine' -->
+      <Reference Include="UnityEngine">
+        <HintPath>$(UnityManagedPath)/UnityEngine/UnityEngine.dll</HintPath>
+        <Private>false</Private>
+      </Reference>
+      <Reference Include="UnityEngine.CoreModule">
+        <HintPath>$(UnityManagedPath)/UnityEngine/UnityEngine.CoreModule.dll</HintPath>
+        <Private>false</Private>
+      </Reference>
```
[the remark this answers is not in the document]
bitsandfoxes: After double-checking, this no longer seems to be required.

mattjohnsonpint (approved):

PR #733, opened 2022-05-08
Linux desktop support - crashpad

### LINKED ISSUE #694: Native crash support for Linux
* Support native crashes on Linux
* Scope sync to the native layer
* Update session status as crashed

Relates to: #693 and #527

vaind:
closes #694, replaces #699

## TODOs
* [x] symbol upload
* [ ] integration test


vaind:
So, it turns out Crashpad cannot be used because it forks the process on Linux and that's not working well with Unity. 
I've tried by investigating the sentry-native source and commenting out/adding `return false at some appropriate places. The good thing is I can recompile the library and swap it in the compiled game folder at `app_Data/Plugins/libsentry.so`

1. `CrashpadClient::StartHandler()` calls `DoubleForkAndExec()`. If I `return false` before the `fork()` call, everything works fine (of course no native crash handling). If I do it right after `fork()`, the UI stops rendering, turns out as a black screen, while you can still click around and in the log I can see the button actions (send dotnet errors) are still triggered) - but no UI, see the image above: https://github.com/getsentry/sentry-unity/pull/733#issuecomment-1120772223

2. Additionally, there seems to be another error, related to log forwarding from native to c# in `SentryNativeBridge.cs`, I assume it has something to do with the `fork()` as well because it only has issues after the call.

Therefore, I'm dropping this approach and we're going to `breakpad`, see #734

PR #734, opened 2022-05-09
Linux desktop support - breakpad

### LINKED ISSUE #694: Native crash support for Linux

vaind:
closes #694, replaces #699 and #733

This implements the linux crash support using the breakpad handler instead of crashpad.

## TODOs
* [x] symbol upload
* [ ] integration test


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -37,17 +38,30 @@ public static void Init()
             {
                 var sentryUnityInfo = new SentryUnityInfo();
 
+                Exception nativeInitException = null;
```
vaind, line 41: I wonder why we didn't have anything like that, as is, if any of the `SentryNative*.Configure()` throws for whatever reason, that exception is not captured, **nor are any others because `SentryUnity.Init()` is never called**
https://sentry.io/organizations/sentry-sdks/issues/3259732515/events/5f72ce01d1074bfb97c59c6fa0a1b7df/?project=5439417

src/Sentry.Unity.Native/SentryNativeBridge.cs
```diff
@@ -168,27 +180,93 @@ private static void nativeLog(int cLevel, string message, IntPtr args, IntPtr us
...
+                    WithAllocatedPtr(formattedLength, buffer =>
+                    {
+                        vsnprintf_windows(buffer, (UIntPtr)formattedLength, format, args);
+                        message = Marshal.PtrToStringAnsi(buffer);
+                    });
+                }
+            }
+            catch (Exception err)
+            {
+                logger.Log(level, $"Exception in native log forwarder: {err}");
```
bruno-garcia, line 216: Exception can be null, but prefer the formatting overload so there's no allocations:

```suggestion
                logger.LogError("Exception in native log forwarder.", err);
```

bruno-garcia (approved):

.gitignore
```diff
@@ -24,6 +24,8 @@ package-dev/**/*.pdb
 package-dev/**/*.xml
 package-dev/**/*.dylib
 package-dev/**/*.dSYM
+package-dev/**/*.so
+package-dev/**/*.full
```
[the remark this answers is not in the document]
vaind: rather ".so.debug" - https://answers.unity.com/questions/1634142/libil2cppsym-libil2cppsodebug-libunitysymso-what-s-1.html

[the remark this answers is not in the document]
vaind: or maybe not, this says something else... https://docs.unity3d.com/2020.3/Documentation/Manual/android-symbols.html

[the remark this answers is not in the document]
vaind: I've made an android build and seems like ".dbg.so" is the corresponding extension (i.e. full unstripped lib) Unity uses in Android builds:

image

package-dev/Runtime/SentryInitialization.cs
vaind: I've improved the exception text here: https://sentry.io/organizations/sentry-sdks/issues/3259732515/?project=5439417&query=is%3Aunresolved

src/Sentry.Unity.Native/SentryNativeBridge.cs
vaind: committed locally

PR #743, opened 2022-05-12
fix: align boolean tag values

### LINKED ISSUE #680: Align boolean tag values
There seem to be multiple boolean value representations, used by different tags. I believe it would be better to only use a single representation, so that filtering is easier for users. What I've seen so far: yes/no, true/false, True/False

image

Also, why is there "handled: no" and "handled: false"?

If we decide to change this, can/should we also update existing data?

vaind:
closes #680 

#skip-changelog

### DIFF
--- a/src/Sentry.Unity/UnityEventProcessor.cs
+++ b/src/Sentry.Unity/UnityEventProcessor.cs
@@ -117,9 +117,7 @@ namespace Sentry.Unity
                 device.BatteryStatus = SystemInfo.batteryStatus.ToString(); // don't cache
 
                 var batteryLevel = SystemInfo.batteryLevel;
-#pragma warning disable RECS0018 // Value is exact when expressing no battery level
-                if (batteryLevel != -1.0)
-#pragma warning restore RECS0018
+                if (batteryLevel > 0.0)
                 {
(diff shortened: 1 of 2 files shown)


src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -117,9 +117,7 @@ private void PopulateDevice(Device device)
                 device.BatteryStatus = SystemInfo.batteryStatus.ToString(); // don't cache
 
                 var batteryLevel = SystemInfo.batteryLevel;
-#pragma warning disable RECS0018 // Value is exact when expressing no battery level
-                if (batteryLevel != -1.0)
-#pragma warning restore RECS0018
+                if (batteryLevel > 0.0)
```
bitsandfoxes, line 120: IIRC this was due to some iOS device reporting the battery status as `-1`.

bitsandfoxes (approved):

src/Sentry.Unity/UnityEventProcessor.cs
vaind: which is lower than 0.0 so the new code would also catch that. It's just that comparing exact floating values doesn't work properly, as the disabled warning was saying.

IMO, no need to try and report negative battery level - it's not a valid value.

PR #744, opened 2022-05-12
fix: collect main thread data synchronously during init

vaind:
After adding the exception handling to SentryInitialization.cs, I've noticed it doesn't collect any of the usual context. This fixes that (and maybe a few other very early user-space error captures). The original behaviour was to run the main thread info collection asynchronously in a coroutine. Now, we collect it right away during init.

## Before 
https://sentry.io/organizations/sentry-sdks/issues/3267829133/events/2651b13a859b4859860b3f44859ffacd/?project=5439417
 
image
 
## After 
https://sentry.io/organizations/sentry-sdks/issues/3267829133/events/e12e6634035b4eb0bdc3c4349a5d03e1/?project=5439417

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Linux native crash support ([#734](https://github.com/getsentry/sentry-unity/pull/734))
+- Collect context information synchronously during init to capture it for very early events ([#744](https://github.com/getsentry/sentry-unity/pull/744))
 
 ## 0.16.0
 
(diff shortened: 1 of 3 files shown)


vaind:
> Can we measure the actual impact once?

I've done a rough measurement with the stopwatch. In the editor, the runtime of `CollectData()`, after removing the yields, ~0.002 sec, i.e. 2 milliseconds. In a desktop build, it was immeasurable (0.000). I could run it on mobile, though from the actual APIs we use, I doubt there would be much of a difference.

bitsandfoxes (approved):

PR #746, opened 2022-05-12
draft: Compilation hooks of IL Weaving

bitsandfoxes:
Goal: Explore how to hook up into the build pipeline before the assemblies are passed on to the IL2CPP compiler.

There seems to be a difference between `ScriptAssemblies` and `PlayerScriptAssemblies` which leads to the events being invoked twice for `Assembly-CSharp.dll` with different sizes.

image

The one in the `PlayerScriptAssemblies` then seems to be cached in `PlayerDataCache` and the `BackupThisFolder_DontShip`:

```
➜  Unity git:(poc/il-weaving-hook) ✗ find . -name "Assembly-CSharp.dll" -exec ls -lh {} \;
-rw-r--r--  1 bitfox  staff    23K May 12 16:11 ./samples/unity-of-bugs/Library/PlayerDataCache/OSXUniversal/Data/Managed/Assembly-CSharp.dll
-rw-r--r--  1 bitfox  staff    24K May 12 16:11 ./samples/unity-of-bugs/Library/ScriptAssemblies/Assembly-CSharp.dll
-rw-r--r--  1 bitfox  staff    23K May 12 16:11 ./samples/unity-of-bugs/Builds/test_BackUpThisFolder_ButDontShipItWithYourGame/Managed/Assembly-CSharp.dll
```

### DIFF
--- a/samples/unity-of-bugs/Assets/Editor/CompileInterceptor.cs
+++ b/samples/unity-of-bugs/Assets/Editor/CompileInterceptor.cs
@@ -1,0 +1,39 @@
+﻿using UnityEditor;
+using UnityEditor.Compilation;
+using UnityEngine;
+
+[InitializeOnLoad]
+public class CompileInterceptor
+{
+    static CompileInterceptor()
+    {
+        CompilationPipeline.compilationStarted -= OnCompilationStarted;
+        CompilationPipeline.compilationStarted += OnCompilationStarted;
+
+        CompilationPipeline.assemblyCompilationStarted -= OnAssemblyCompilationStarted;
+        CompilationPipeline.assemblyCompilationStarted += OnAssemblyCompilationStarted;
+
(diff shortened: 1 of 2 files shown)


bitsandfoxes:
Alternative possibility to look at: https://docs.unity3d.com/2019.4/Documentation/ScriptReference/Build.IPostBuildPlayerScriptDLLs.html

bitsandfoxes:
> Alternative possibility to look at: https://docs.unity3d.com/2019.4/Documentation/ScriptReference/Build.IPostBuildPlayerScriptDLLs.html

Sounds good - doesn't work. It seems to be called at the very end. The previous approach seems to be the way to go with modifying the assembly in the `PlayerScriptAssemblies` directory. We just need to make sure this behaves the same in all Unity versions.

PR #756, opened 2022-05-21
feat: use single-threaded web worker on unknown platforms

### LINKED ISSUE #290: Game Crashed on Nintendo Switch with Sentry SDK
### Environment

How do you use Sentry?
on-premise

Which SDK and version?
0.4.3

### Steps to Reproduce

1. Run game with sdk

### Expected Result

Not crashing.

### Actual Result

Game crashed.
I guess it's because of trying to write to disk incorrect way.

vaind:
Uses UnityWebRequestTransport by default on other than the officially supported platforms.

(hopefully) fixes #290

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -7,12 +7,15 @@
 #define SENTRY_NATIVE
 #elif UNITY_WEBGL
 #define SENTRY_WEBGL
+#else
+#define SENTRY_DEFAULT
 #endif
 #endif
 
 using System;
 using UnityEngine;
 using UnityEngine.Scripting;
+using UnityEngine.Analytics;
 
 #if SENTRY_NATIVE_COCOA
 using Sentry.Unity.iOS;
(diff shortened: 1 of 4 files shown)


src/Sentry.Unity/Default/UnknownPlatform.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Configure Sentry on an officially unsupported platform.
+    /// This works on a best-effort basis, to try to improve compatibility.
+    /// </summary>
+    public static class SentryUnknownPlatform
+    {
+        public static void Configure(SentryUnityOptions options)
+        {
+            // This is only provided on a best-effort basis for other then the explicitly supported platforms.
```
bitsandfoxes, line 14:
```suggestion
            // This is only provided on a best-effort basis for other than the explicitly supported platforms.
```

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -7,12 +7,15 @@
...
 #define SENTRY_WEBGL
+#else
+#define SENTRY_DEFAULT
 #endif
 #endif
 
 using System;
 using UnityEngine;
 using UnityEngine.Scripting;
+using UnityEngine.Analytics;
```
bitsandfoxes, line 18:
```suggestion
```

bitsandfoxes (approved):
LGTM! Thanks!

PR #763, opened 2022-05-24
Test: symbol upload

vaind:


.github/workflows/ci.yml
```diff
@@ -266,6 +265,13 @@ jobs:
           path: samples/IntegrationTest/Build
           if-no-files-found: error
 
+      - name: Upload IntegrationTest project on failure
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: failed-project-${{ matrix.platform }}-${{ matrix.unity-version }}
+          path: samples/IntegrationTest
+
```
bitsandfoxes, line 274: Great idea!

scripts/smoke-test-ios.ps1
```diff
@@ -43,26 +58,32 @@ Class AppleDevice
...
-    $xCodeBuild.WaitForExit()
-    If ($xCodeBuild.ExitCode -ne 0)
-    {
-        Throw "Xcode build exited with code $($xCodeBuild.ExitCode)."
+    $symbolServerOutput = RunWithSymbolServer -Callback {
+        Write-Host "Building iOS project"
+
+        # We need to manually switch the CLI executable, because the artifact that came through GH actions'
+        # upload-artifact has it's permissions stripped. See https://github.com/actions/upload-artifact/issues/38
+        # Side note: the permissions are set by the SentryCli.cs so end-users aren't n affected if we ship the CLI
```
bitsandfoxes, line 74:
```suggestion
        # Side note: The permissions are set by the SentryCli.cs so end-users aren't affected if we ship the CLI
```

src/Sentry.Unity.Editor/SentryCli.cs
```diff
@@ -102,5 +102,21 @@ internal static void AddExecutableToXcodeProject(string projectPath, IDiagnostic
             File.Copy(executableSource, executableDestination);
             SetExecutePermission(executableDestination);
         }
+
+        internal static string? UrlOverride(string? dsnOption, string? urlOverrideOption)
```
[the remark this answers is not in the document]
vaind: Because it may not be needed to specify the override in the first place. If you've configured the DSN to be an internal server that's available during build, it should just work without specifying the same server as upload override. As is, it would try to upload to sentry.io instead, if you didn't specify any override.

PR #765, opened 2022-05-25
dont alias System.Runtime.CompilerServices.Unsafe

### LINKED ISSUE #691: Builds on Unity 2022.1 break with aliased package
Building an IL2CPP build with the aliased Sentry package installed results in: 
The build works fine when using the un-aliased Sentry package.

```
Building .../MacStandalonePlayerBuildProgram/4g2s2/96vv_icMethods2.o failed with output:
.../MacStandalonePlayerBuildProgram/il2cppOutput/cpp/GenericMethods2.cpp:16744:10:
error: cannot initialize return object of type 'Il2CppFullySharedGenericAny *' (aka 'void **') with an lvalue of type 'void *'
                return (L_2);
                       ^~~~~
1 error generated.
```

This is the generated method in question:
```
// T& System.Runtime.CompilerServices.Unsafe::Unbox<Unity.IL2CPP.Metadata.__Il2CppFullySharedGenericType>(System.Object)
IL2CPP_EXTERN_C IL2CPP_METHOD_ATTR Il2CppFullySharedGenericAny* Unsafe_Unbox_TisIl2CppFullySharedGenericAny_mF6CD1F81A0CA0414C3EA88A19A4BB220865E8182_gshared (RuntimeObject* ___box0, const RuntimeMethod* method) 
{
  il2cpp_rgctx_method_init(method);
  // T
  // sizeof(T)
  const uint32_t SizeOf_T_t6EEA42083AA5049EFD4B883DDFBF7D3F1BF8696C = il2cpp_codegen_sizeof(il2cpp_rgctx_data(method->rgctx_data, 0));
  const Il2CppFullySharedGenericAny L_1 = alloca(SizeOf_T_t6EEA42083AA5049EFD4B883DDFBF7D3F1BF8696C);
  {
    RuntimeObject* L_0 = ___box0;
    void* L_2 = UnBox_Any(L_0, il2cpp_rgctx_data(method->rgctx_data, 0), L_1);
    return (L_2);
  }
}
```

The issue has been raised with Unity and can be tracked here: https://fogbugz.unity3d.com/default.asp?1419859_qmt0m8hc0mpc1mc5
Since we're aliasing the DLLs but the IL remains valid and it seems to be related to Unity's new `Il2CppFullySharedGenericAny` feature we hope this can be resolved by the IL2CPP team.

I've appended a minimal repro that contains a link.xml (to keep the Unsafe.dll from being stripped) and the aliased Unsafe.dll
[2022_1.zip](https://github.com/getsentry/sentry-unity/files/8513192/2022_1.zip)

SimonCropp:
fixes #691

### DIFF
--- a/test/Scripts.Tests/package-release.zip.snapshot
+++ b/test/Scripts.Tests/package-release.zip.snapshot
@@ -301,8 +301,6 @@ Runtime/Sentry.System.Numerics.Vectors.dll
 Runtime/Sentry.System.Numerics.Vectors.dll.meta
 Runtime/Sentry.System.Reflection.Metadata.dll
 Runtime/Sentry.System.Reflection.Metadata.dll.meta
-Runtime/Sentry.System.Runtime.CompilerServices.Unsafe.dll
-Runtime/Sentry.System.Runtime.CompilerServices.Unsafe.dll.meta
 Runtime/Sentry.System.Text.Encodings.Web.dll
 Runtime/Sentry.System.Text.Encodings.Web.dll.meta
 Runtime/Sentry.System.Text.Json.dll
@@ -329,6 +327,8 @@ Runtime/Sentry.xml
 Runtime/Sentry.xml.meta
 Runtime/SentryInitialization.cs
 Runtime/SentryInitialization.cs.meta
+Runtime/System.Runtime.CompilerServices.Unsafe.dll
+Runtime/System.Runtime.CompilerServices.Unsafe.dll.meta
 Samples~/unity-of-bugs/Scenes.meta
 Samples~/unity-of-bugs/Scripts.meta
 Samples~/unity-of-bugs/Scenes/1_Bugfarm.unity
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,7 +134,7 @@ jobs:
       - name: Assembly alias
         run: |
           docker exec unity bash -c 'dotnet tool install --global Alias --version 0.4.3 && \
-          ~/.dotnet/tools/assemblyalias --target-directory "package-dev/Runtime" --internalize --prefix "Sentry." --assemblies-to-alias "Microsoft*;System*"'
+          ~/.dotnet/tools/assemblyalias --target-directory "package-dev/Runtime" --internalize --prefix "Sentry." --assemblies-to-alias "System.Collections.Immutable;System.Numerics.Vectors;System.Buffers;System.Threading.Tasks.Extensions;Microsoft.Bcl.AsyncInterfaces;System.Text.Encodings.Web;System.Text.Json;System.Memory;System.Reflection.Metadata"'
 
       - name: Update .meta files before packaging
         # We need to open & close Unity on the sample project to update .meta files in package-dev.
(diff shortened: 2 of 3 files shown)


vaind:
https://github.com/getsentry/sentry-unity/runs/6590500329?check_suite_focus=true

SimonCropp:
how does that look?

test/Scripts.Tests/package-release.zip.snapshot
```diff
@@ -329,6 +327,8 @@ Runtime/Sentry.xml
 Runtime/Sentry.xml.meta
 Runtime/SentryInitialization.cs
 Runtime/SentryInitialization.cs.meta
+Runtime/System.Runtime.CompilerServices.Unsafe.dll
+Runtime/System.Runtime.CompilerServices.Unsafe.dll.meta
 Samples~/unity-of-bugs/Scenes.meta
```
vaind, line 332: Hmm, actually... Should we even ship these libs when they're not aliased?

SimonCropp: i dont know?

SimonCropp: perhaps @bitsandfoxes knows

bitsandfoxes: We should definitely ship with these since they are our dependencies. 

But we've got a bit of a cat chasing its own tail situation here:
1. We alias our dependencies so we don't conflict with other packages
2. We don't alias that specific DLL to make it work with 2022.1
3. If there is any other package that has this DLL as a dependency they will conflict
4. Our solution to resolving that conflict would be to alias the DLL but we can't because 1.

SimonCropp: happy to abandon this and wait for Unity to fix the issue. but in the short term i think potentially having a warning for one dependency is better than (removing alias and) having warnings for all.

vaind: so to summarise, we're going with this solution for now. Anything against merging this @bitsandfoxes ?

bruno-garcia: It's not just a warning, since we depend on version 5.0.0, and some packages that conflicted in the past brought in version 4.0.0.

We should test this change with some known packages and validate it actually works. Ideally add an integration test: Add unity's collection package to the smoke test empty project?

bitsandfoxes:
We've got a response from Unity: They are looking into it and are keen to fix this. 
I'm closing this for now since excluding `System.Runtime.CompilerServices.Unsafe` from aliasing would potentially break users with conflicting dependencies on all other Unity versions.

bruno-garcia:
Leaving the branch alive for a bit until we hear back from Unity

SimonCropp:
@bitsandfoxes thanks for chasing this

bruno-garcia:
This issue is tracked here and Unity is working on a fix: https://issuetracker.unity3d.com/issues/il2cpp-build-fails-when-using-an-assembly-renamed-via-sentrys-assembly-alias-tool

More details here: https://docs.sentry.io/platforms/unity/troubleshooting/#incompatibility-between-sentry-sdk-and-unity-20221
And here: https://github.com/getsentry/sentry-unity/issues/691

PR #770, opened 2022-05-30
feat: disable offline caching on unknown platforms

vaind:


bruno-garcia:
Downside is that it could work on xbox/playstation and we'd be turning it off. 
I wonder if it's best to just ifdef Switch and turn it off only then

src/Sentry.Unity/Default/UnknownPlatform.cs
```diff
@@ -11,14 +11,23 @@ public static class SentryUnknownPlatform
     {
         public static void Configure(SentryUnityOptions options)
         {
+            options.DiagnosticLogger?.Log(SentryLevel.Debug,
+                "Configuring on an unknown platform - some options may be overridden to improve compatibility.");
```
bitsandfoxes, line 15:
```suggestion
                "Configuring on an unsupported platform - some options may be overridden to improve compatibility.");
```
What do you think about unsupported instead of unknown? Because we actually do know the current platform, it's just not supported (yet).

src/Sentry.Unity/Default/UnknownPlatform.cs
```diff
@@ -11,14 +11,23 @@ public static class SentryUnknownPlatform
...
+
             // This is only provided on a best-effort basis for other than the explicitly supported platforms.
             if (options.BackgroundWorker is null)
             {
-                options.DiagnosticLogger?.Log(SentryLevel.Debug,
-                        "Configuring on an unknown platform. Using WebBackgroundWorker to improve compatibility.");
+                options.DiagnosticLogger?.Log(SentryLevel.Debug, "Unknown platform: using WebBackgroundWorker.");
```
bitsandfoxes, line 20:
```suggestion
                options.DiagnosticLogger?.Log(SentryLevel.Debug, "Unsupported platform: using WebBackgroundWorker.");
```

src/Sentry.Unity/Default/UnknownPlatform.cs
```diff
@@ -11,14 +11,23 @@ public static class SentryUnknownPlatform
...
                 options.BackgroundWorker = new WebBackgroundWorker(options, SentryMonoBehaviour.Instance);
             }
+
             options.DefaultUserId = AnalyticsSessionInfo.userId;
+
+            if (options.CacheDirectoryPath is not null)
+            {
+                options.DiagnosticLogger?.Log(SentryLevel.Debug, "Unknown platform: disabling offline caching.");
```
bitsandfoxes, line 28:
```suggestion
                options.DiagnosticLogger?.Log(SentryLevel.Debug, "Unsupported platform: disabling offline caching.");
```

src/Sentry.Unity/Default/UnknownPlatform.cs
```diff
@@ -11,14 +11,23 @@ public static class SentryUnknownPlatform
     {
         public static void Configure(SentryUnityOptions options)
         {
+            options.DiagnosticLogger?.Log(SentryLevel.Debug,
```
bitsandfoxes, line 14:
```suggestion
            options.DiagnosticLogger?.Log(SentryLevel.Info,
```
What do you think about raising this to `Info`?

bitsandfoxes (approved):

src/Sentry.Unity/Default/UnknownPlatform.cs
vaind: I was considering this. Why I chose not to is because it's not really "unsupported", just not "officially supported"

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Collect context information synchronously during init to capture it for very early events ([#744](https://github.com/getsentry/sentry-unity/pull/744))
 - Automatic user IDs on native crashes & .NET events ([#728](https://github.com/getsentry/sentry-unity/pull/728))
 - Use single-threaded HTTP transport on unknown platforms ([#756](https://github.com/getsentry/sentry-unity/pull/756))
+- Disable offline caching on unknown platforms ([#770](https://github.com/getsentry/sentry-unity/pull/770))
```
bitsandfoxes, line 11:
```suggestion
- Use single-threaded HTTP transport on currently unsupported platforms ([#756](https://github.com/getsentry/sentry-unity/pull/756))
- Disable offline caching on currently unsupported platforms ([#770](https://github.com/getsentry/sentry-unity/pull/770))
```

bruno-garcia (commented):
The word "Unsupported" like that might give the wrong message? Maybe reword the whole thing to be more clear:

"Platform support for Offline Caching is unknown. Disabling it"

PR #771, opened 2022-05-30
Application-not-responding detection

### LINKED ISSUE #708: ANR detection of the Unity event loop
The SDK should detect "application-not-responding" based on the Android SDK (need to confirm they are being sent).

Sample can be done by blocking for more than 60 seconds in C#

The Android approach is here: https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java

vaind:
* closes #708 
* https://sentry.io/organizations/sentry-sdks/issues/3314229346/events/e83f4535c2294197b393000c2a09452d/?project=5439417

### DIFF
--- a/src/Sentry.Unity/Integrations/ANRIntegration.cs
+++ b/src/Sentry.Unity/Integrations/ANRIntegration.cs
@@ -27,23 +27,23 @@ namespace Sentry.Unity
 
     internal class ANRWatchDog
     {
-        private int _checkIntervalMilliseconds;
+        private int _detectionTimeoutMs;
 
-        // Note: we don't sleep for the check interval duration to prevent the thread from blocking shutdown.
-        // Note 2: if this is higher than _checkIntervalMilliseconds, the lower value is used.
-        private const int sleepInterval = 1000;
+        // Note: we don't sleep for the whole detection timeout or we wouldn't capture if the ANR started later.
+        private int _sleepIntervalMs;
 
         private IDiagnosticLogger? _logger;
         private SentryMonoBehaviour _monoBehaviour = null!;
-        private bool _uiIsResponsive = false;
+        private int _ticksSinceUiUpdate = 0; // how many _sleepIntervalMs have elapsed since the UI updated last time
         private bool _reported = false; // don't report the same ANR instance multiple times
         private bool _stop = false;
         private Thread _thread = null!;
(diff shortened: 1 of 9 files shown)


bruno-garcia:
Added @marandaneto in case he wants to take a look. He's got a lot of experience reducing noise from the Android ANR so this PR could benefit from that

src/Sentry.Unity/Integrations/ANRIntegration.cs
```diff
@@ -0,0 +1,134 @@
...
+                    _thread = new Thread(Run)
+                    {
+                        Name = "Sentry-ANR-WatchDog"
+                    };
+                    _thread.Start();
+
+                    // Stop the thread when the app is being shut down.
+                    _monoBehaviour.Application.Quitting += () =>
+                    {
+                        _cancel.Cancel();
```
bruno-garcia, line 67: Cancel can throw. Should we have `Join` in a `finally` block and log out on `catch`?

src/Sentry.Unity/Integrations/ANRIntegration.cs
```diff
@@ -0,0 +1,134 @@
...
+                        }
+
+                        _uiIsResponsive = false;
+                        watch.Restart();
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                _logger?.Log(SentryLevel.Warning, "Exception in the ANR watchdog.", e);
```
bruno-garcia, line 123: Perhaps best to use Level Error?

vaind: added a catch clause for ThreadAbortException with a warning and changed this to error

src/Sentry.Unity/Integrations/ANRIntegration.cs
vaind: removed the cancellation token in favour of a plain bool, which in the meantime I've found out is atomic in c#

vaind:
Note: I've refactored the code to make the thread a [background thread](https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread.isbackground?view=netstandard-2.1), to not block during app shutdown

bruno-garcia (approved):

brustolin (approved):
LGTM

PR #773, opened 2022-06-01
feat: Allow uploading sources for debug files

Swatinem:


Swatinem:
I think the source bundles are the same as other debug files in that sense. I can bug @bitsandfoxes to help me set up the integration tests so I can get something up and running.

vaind:
Hmm, the sources don't seem to be easily distinguishable from binary files. Here's the diff of uploading the same directory (windows desktop build), with upload-sources enabled in one case. I deduce `GameAssembly.pdb` are being uploaded...

image

Swatinem:
In sentry, all these files have the same name, with the same debug_id. But they have different contents, like here:

Bildschirmfoto 2022-06-02 um 11 59 47

So I think this is perfectly fine. Although I agree its confusing. We do use `.src.zip` in some places, although its not a valid zip file either because it has its own header prepended before the actual zip contents.

Swatinem:
One thing to watch out for. Some time ago we had the problem that source bundles were being re-uploaded all the time, because they were different all the time because of unstable sorting order internally. Now we should have a stable sorting order. But it seems that at least my local sentry-cli still re-uploads these files over and over again, so that bug is not fixed yet :-(

The reason I say this is because you would probably end up with different chunk hashes, and thus wouldn’t be able to use snapshot testing for this.

vaind:
> One thing to watch out for. Some time ago we had the problem that source bundles were being re-uploaded all the time, because they were different all the time because of unstable sorting order internally. Now we should have a stable sorting order. But it seems that at least my local sentry-cli still re-uploads these files over and over again, so that bug is not fixed yet :-(
> 
> The reason I say this is because you would probably end up with different chunk hashes, and thus wouldn’t be able to use snapshot testing for this.

So far the symbol upload tests didn't use the hash, just the content... not sure what to do about the source upload yet, I may need to change the test-server response so that the CLI actually uploads the content so that I know what came in. Right now the server always responds with a message indicating all chunks are already uploaded.

Swatinem:
Isn’t that a bit contradictory? Using the content, but then the contents are not actually uploaded because the server says chunks already exist?

Btw, this was the issue with source bundles: https://github.com/getsentry/sentry-cli/issues/831 The problem is rather timestamps -_- So maybe there is a fairly easy workaround for this in symbolic.

vaind:
sorry, brain lag there - I was going to say "file name" instead of "content", I've fixed the comment.

vaind:
I've added the tests (took way more effort than anticipated... as usual), though only based on the number of uploaded files as I couldn't (in a self-imposed time box) figure out unpacking & analyzing the content of the uploaded files. Should be still good to catch issues in the future: https://github.com/getsentry/sentry-unity/pull/773/commits/21cb45b368645b02fb56e2b1fa4e636a7bf6d336

Swatinem:
❤️ thanks so much for taking this over! @vaind

test/Sentry.Unity.Editor.iOS.Tests/SentryXcodeProjectTests.cs
```diff
@@ -129,7 +129,7 @@ public void AddBuildPhaseSymbolUpload_CleanXcodeProject_BuildPhaseSymbolUploadAd
             xcodeProject.ReadFromProjectFile();
 
             var didContainUploadPhase = xcodeProject.MainTargetContainsSymbolUploadBuildPhase();
-            xcodeProject.AddBuildPhaseSymbolUpload(_fixture.Options.DiagnosticLogger);
+            xcodeProject.AddBuildPhaseSymbolUpload(_fixture.Options.DiagnosticLogger, new SentryCliOptions());
```
bitsandfoxes, line 132: Since `uploadSources` is `false` by default, should we add unit tests for the path that adds `--include-sources`?

src/Sentry.Unity.Editor/ConfigurationWindow/DebugSymbolsTab.cs
```diff
@@ -17,6 +17,10 @@ internal static void Display(SentryCliOptions cliOptions)
                                                      "on development builds."),
                 cliOptions.UploadDevelopmentSymbols);
 
+            cliOptions.UploadSources = EditorGUILayout.Toggle(
+                new GUIContent("Upload Sources", "Whether source code should be uploaded."),
```
bruno-garcia, line 21:
```suggestion
                new GUIContent("Upload Sources", "Whether your source code should be uploaded to Sentry, so the stack trace in Sentry has the relevant code next to it."),
```

bruno-garcia (approved):
Other than missing tests LGTM

bitsandfoxes (approved):

vaind:
merging as is to get the changes for #785 , I'll add unit tests & update CI to run alternatively with/without sources in a follow up PR

PR #774, opened 2022-06-02
Refactor droid steps

lucas-zimerman:
The goal is to move the additional changes from the latest android test into a separate PR and then move back to the latest version of Android.

Changes:

- Updated action to setup the Simulators (Actions Logs are more organized by groups)
- Fix issue where some simulators may fail due to out of storage errors.
- Retry to install the APK if ADB  failed to communicate with the Device.
- Add new logic to close Alerts (the old one does not work with the latest devices plus when an alert happens it's more readable than having to visualize an xml with the UI information)
- Added additional step to concatenate the variables used by the Simulator steps (removing the if conditions from the smoke test + smoke test retry)
- Added logic to retrieve the latest Android version.

#skip-changelog

### DIFF
--- a/scripts/smoke-test-droid.ps1
+++ b/scripts/smoke-test-droid.ps1
@@ -79,17 +79,17 @@ function CloseSystemAlert([string] $deviceId, [string] $deviceApi, [string] $ale
 {
     if ("$alert" -ne "")
     {
-        $splittedXml = $alert -split "<node"
+        $splitXml = $alert -split "<node"
         $alertTitle = ""
         $alertOption1Label = $null
         $alertOption1Coord = $null
         $alertOption2Label = $null
         $alertOption2Coord = $null
 
-        if ($splittedXml.Count -ne 1) {
+        if ($splitXml.Count -ne 1) {
             # We have a "valid" XML
             # Use Regex to get the message and the options labels + coordinates.
-            foreach ($iterator in $splittedXml) {
+            foreach ($iterator in $splitXml) {
                 if ($iterator.Contains("alertTitle")) {
                     $titleRegex = [regex]::Match($iterator, "text=\`"(?<text>.+)\`" resource-id")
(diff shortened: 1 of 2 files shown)


scripts/smoke-test-droid.ps1
```diff
@@ -79,19 +79,70 @@ function CloseSystemAlert([string] $deviceId, [string] $deviceApi, [string] $ale
...
         {
-            # sends "back" action
-            Write-Warning "Issuing BACK command to close the current window."
-            adb -s $deviceId shell input keyevent 4
+            # Fallback to the old method of closing Alerts.
+            Write-Warning "Active system alert found on $deviceId (API $deviceApi). Closing it. The alert was: '$alert'."
+            if ($deviceApi -eq "21")
+            {
+                Write-Warning "Issuing ENTER command twice to close the current window."
+                # sends "enter" - the first one focues the OK button, the second one taps it
```
lucas-zimerman, line 136:
```suggestion
                # sends "enter" - the first one focus the OK button, the second one taps it
```

scripts/smoke-test-droid.ps1
```diff
@@ -79,19 +79,70 @@ function CloseSystemAlert([string] $deviceId, [string] $deviceApi, [string] $ale
...
+                Write-Host "Tapping on $tapLabel at [$tapX, $tapY]"
+                adb -s $deviceId shell input tap $tapX $tapY
+            }
         }
         else
         {
-            # sends "back" action
-            Write-Warning "Issuing BACK command to close the current window."
-            adb -s $deviceId shell input keyevent 4
+            # Fallback to the old method of closing Alerts.
```
vaind, line 131: Maybe comment which versions still use this.

vaind (approved):
LGTM

vaind (approved):
Awesome

PR #783, opened 2022-06-07
fix: trim CacheDirectoryPath

vaind:
sometimes in developemnt the `application.PersistentDataPath` had a trailing space on Windows... causing test failures

#skip-changelog

### DIFF
--- a/src/Sentry.Unity/SentryUnityOptions.cs
+++ b/src/Sentry.Unity/SentryUnityOptions.cs
@@ -180,7 +180,7 @@ namespace Sentry.Unity
                 ? "editor"
                 : "production";
 
-            CacheDirectoryPath = application.PersistentDataPath;
+            CacheDirectoryPath = application.PersistentDataPath.Trim();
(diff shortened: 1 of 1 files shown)


bruno-garcia (approved):

vaind:
> Assuming it can't be null ever, even on WebGL

The variable is non-nullable, so not an issue here: https://github.com/getsentry/sentry-unity/blob/9e68ee9e9c7c2f78fdbad3c90191701037d52246/src/Sentry.Unity/Integrations/IApplication.cs#L46

bruno-garcia:
I assume `Application.persistentDataPath` doesn't have nullability attributes, but if I'm correct, shouldn't the IDE warn that's possibly null?

Not much help here:
https://github.com/Unity-Technologies/UnityCsReference/blob/3fcad4bbbea0455588843799c67861f7b9eb3825/Runtime/Export/Application/Application.bindings.cs#L185-L189

PR #784, opened 2022-06-07
chore: Replace screenshot options with internal defaults

bitsandfoxes:
The goal is to simplify the screenshot options by moving the size and quality control into the SDK.
The `MaxSize = 1920` is picked from FullHD 1920x1080 and is open for discussion. It's supposed to guard against overly large screenshots being sent.


bruno-garcia (approved):
Nice! Simpler is better

oddgames-david:
I don't understand this change. We lose the no fuss ability to send smaller screenshots and save on storage costs?

bitsandfoxes:
We agree and we're trying to get something like that back in.

PR #785, opened 2022-06-07
Integration-test building without sentry SDK only on `main` branch

vaind:
I've originally made the change to ci.yml in #773, but to my surprise, it actually broke builds - suddenly some symbols weren't uploaded. Turns out this is also reproducible locally - the first build doesn't include `*.dbg.so`/`*.sym.so` symbols. 

Update: the cause was that we relied on `PlayerSettings.GetScriptingBackend(EditorUserBuildSettings.selectedBuildTargetGroup)` but our `Builder.cs` didn't actually set the `EditorUserBuildSettings.selectedBuildTargetGroup`. So this wasn't really an issue for end-users, just the issue of our tests. 

Before finally finding this out, I've thought it was the issue of Unity not putting the il2cpp symbols in the output folder, so I've originally collected these from Unity editor Data folder. I've reverted this change in the final version of the PR, but kept some refactorings that were originally needed to make it (and tests) work.

#skip-changelog


.github/workflows/ci.yml
```diff
@@ -229,6 +229,8 @@ jobs:
         run: sudo pwsh ./test/Scripts.Integration.Test/integration-create-project.ps1 -UnityPath "${{ env.UNITY_PATH }}"
 
       - name: Build without Sentry SDK
+        # This hasn't broken for many months, so disabling on PRs to speed up CI. And also to test a clean build with Sentry SDK included.
```
bitsandfoxes, line 232: A surprise but a welcome one!

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
```diff
@@ -20,19 +20,26 @@ internal class DebugSymbolUpload
...
 
+        private const string SymbolUploadTaskStartComment = "// Autogenerated Sentry symbol upload task [start]";
+        private const string SymbolUploadTaskEndComment = "// Autogenerated Sentry symbol upload task [end]";
+
         private string _symbolUploadTask = @"
 // Credentials and project settings information are stored in the sentry.properties file
 gradle.taskGraph.whenReady {{
     gradle.taskGraph.allTasks[-1].doLast {{
-        println 'Uploading symbols to Sentry'
+        println 'Uploading symbols to Sentry. You can find the full log in ./Logs/sentry-symbols-upload.log (the file content may not be strictly sequential because it\'s a merge of two streams).'
```
bitsandfoxes, line 35: ❤️

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
```diff
@@ -114,12 +125,8 @@ public void RemoveUploadFromGradleFile()
                 return;
             }
 
-            // Replacing the paths with '.*' and escaping the task
-            var uploadTaskFilter = string.Format(_symbolUploadTask, ".*", ".*");
-            uploadTaskFilter = Regex.Replace(uploadTaskFilter, "\"", "\\\"");
-            uploadTaskFilter = Regex.Replace(uploadTaskFilter, @"\[", "\\[");
-
-            gradleBuildFile = Regex.Replace(gradleBuildFile, uploadTaskFilter, "");
+            var regex = new Regex(Regex.Escape(SymbolUploadTaskStartComment) + ".*" + Regex.Escape(SymbolUploadTaskEndComment), RegexOptions.Singleline);
```
bitsandfoxes, line 128: Thanks for fixing that!

bitsandfoxes (approved):

PR #786, opened 2022-06-07
chore: Anr code formatting

bitsandfoxes:
Couldn't help myself. #skip-changelog


src/Sentry.Unity/Integrations/AnrIntegration.cs
```diff
@@ -8,76 +8,76 @@
...
 namespace Sentry.Unity
 {
-    internal class ANRIntegration : ISdkIntegration
+    internal class AnrIntegration : ISdkIntegration
     {
-        private static readonly object _lock = new object();
-        private static ANRWatchDog? _watchdog;
-        private SentryMonoBehaviour _monoBehaviour;
+        private static readonly object Lock = new();
+        private static AnrWatchDog? Watchdog;
```
vaind, line 14: am I missing something? We're using `_` prefix for private members AFAICS

bitsandfoxes: Good catch! Yes, but not for `static`. I "fixed" it where I found it. (except for the SentryMonoBehaviour)

src/Sentry.Unity.Native/NativeScopeObserver.cs
```diff
@@ -57,7 +57,7 @@ public override void SetUserImpl(User user)
         [DllImport("sentry")]
         private static extern int sentry_value_set_by_key(SentryValueU value, string k, SentryValueU v);
 
-        private static void nativeSetValueIfNotNull(SentryValueU obj, string key, string? value)
+        private static void NativeSetValueIfNotNull(SentryValueU obj, string key, string? value)
```
bitsandfoxes, line 60: Not sure about that one since it's the one calling the `extern`.

vaind (approved):

bruno-garcia (approved):

PR #791, opened 2022-06-08
fix: LogException marked unhandled test

bitsandfoxes:
#skip-changelog


bitsandfoxes:
Looks like https://github.com/getsentry/sentry-unity/pull/792 took care of it.

Issue #754, opened 2022-05-18
App.version missing in .NET & desktop native


vaind:
Native mobile SDKs expose app version (at least Android does, I haven't checked iOS but I assume it has it as well) - we don't have it in .NET events or on desktop crashes (Windows, Linux). They're very useful to track which app releases have the issue.
I was thinking of adding the following [contexts](https://develop.sentry.dev/sdk/event-payloads/contexts/#app-context):
* `App.Version` - https://docs.unity3d.com/ScriptReference/Application-version.html
* `App.Build` - https://docs.unity3d.com/ScriptReference/Application-buildGUID.html

bruno-garcia:
We currently use `Application.version` to create the `Release`:

https://github.com/getsentry/sentry-unity/blob/c8960266ef2973d3e4aa804e967c97bb4323c757/src/Sentry.Unity/SentryUnityOptions.cs#L167-L168

These numbers should match. Perhaps Build is missing? The format `package@version+build` is defined here:

https://docs.sentry.io/platforms/unity/configuration/releases/#bind-the-version

The parts of the release could be added to the context["app"] indeed.

bitsandfoxes:
This has been fixed in https://github.com/getsentry/sentry-unity/pull/795

PR #795, opened 2022-06-14
feat: include build ID in release

vaind:
Addresses https://github.com/getsentry/sentry-unity/issues/754#issuecomment-1141913562


bitsandfoxes (approved):

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -164,17 +165,20 @@ internal SentryUnityOptions(IApplication application, bool isBuilding)
             StackTraceMode = StackTraceMode.Original;
             IsEnvironmentUser = false;
 
+            var releaseSB = new StringBuilder();
```
bitsandfoxes, line 168: I'm curious: Why did you go with a StringBuilder here?

vaind: habit, but it's an overkill here, reverting to string concatenation

Issue #637, opened 2022-03-14
Exclude SmokeTester from samples


bitsandfoxes:
The SmokeTester was part of the sample project out of necessity. But with the vast improvements CI experienced we should exclude it in the samples package.

Additionally, we can rework it at the same time to no longer require to `Init` the SDK twice using the OptionsConfiguration.

bitsandfoxes:
Also avoids undesired side-effects like https://github.com/getsentry/sentry-unity/issues/714

vaind:
Let's create a new IntegrationTest Unity project in `./test` and move over the stuff needed for tests. Updating CI shouldn't be a big deal because scripts already use a path we just need to update in a few files.

PR #798, opened 2022-06-15
fix: Removed SmokeTester from package

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/637

#skip-changelog


test/Scripts.Integration.Test/Editor/Builder.cs
```diff
@@ -17,11 +17,23 @@ public static void BuildIl2CPPPlayer(BuildTarget target, BuildTargetGroup group)
...
+        if(args.ContainsKey("uploadSymbols"))
+        {
+            Debug.Log("Enabling automated debug symbol upload.");
+            cliOptions.UploadSymbols = true;
+        }
+        else
         {
-            SetupSentryOptions(args);
+            Debug.Log("Disabling automated debug symbol upload.");
+            cliOptions.UploadSymbols = false;
```
vaind, line 31: What does this solve? UploadSymbols defaults to `true`

test/Scripts.Integration.Test/build-project.ps1
```diff
@@ -15,14 +15,22 @@ $outputPath = "$NewProjectBuildPath/$(GetTestAppName $buildMethod)"
...
-if ($CheckSymbols)
+# The iOS symbol upload gets validated as part of the actual Xcode build
+if($Platform -eq "iOS" -And $CheckSymbols)
 {
+    $unityArgs += "-uploadSymbols"
+    RunUnityCustom $unityPath $unityArgs
+}
+elseif ($CheckSymbols)
+{
+    $unityArgs += "-uploadSymbols"
```
vaind, line 26: I guess this is related to my previous question. Why do we need this, though? It worked before without explicitly handling iOS here.

test/Scripts.Integration.Test/common.ps1
```diff
@@ -1,5 +1,5 @@
-# Note: this is currently used by "integration-*.ps1" scripts as well as "smoke-test-*.ps1" scripts.
-# If/when those are merged to some extent, maybe this file could be merged into `IntegrationGlobals.ps1`.
+# Note: this is currently used by "*.ps1" scripts as well as "smoke-test-*.ps1" scripts.
```
vaind, line 1:
```suggestion
# Note: this is currently used by integration test scripts as well as "smoke-test-*.ps1" scripts.
```

test/Scripts.Integration.Test/integration-test.ps1
```diff
@@ -1,7 +1,7 @@
...
-.\test/Scripts.Integration.Test/integration-create-project.ps1 "$path"
-.\test/Scripts.Integration.Test/integration-build-project.ps1 "$path"
-.\test/Scripts.Integration.Test/integration-update-sentry.ps1 "$path"
-.\test/Scripts.Integration.Test/integration-build-project.ps1 "$path"
-.\test/Scripts.Integration.Test/integration-run-smoke-test.ps1
+.\test/Scripts.Integration.Test/create-project.ps1 "$path"
+.\test/Scripts.Integration.Test/build-project.ps1 "$path"
+.\test/Scripts.Integration.Test/update-sentry.ps1 "$path"
+.\test/Scripts.Integration.Test/build-project.ps1 "$path"
+.\test/Scripts.Integration.Test/run-smoke-test.ps1
```
vaind, line 7: BTW this file is (likely) broken - it hasn't been updated with all the changes for the last few months. Do we want to keep & fix it?

vaind (commented):
nice!

test/Scripts.Integration.Test/Editor/Builder.cs
bitsandfoxes: Since it is `true` by default there is no way of running it locally. The build would always fail with sentry-cli trying to upload to a non-running server.

vaind: Can you maybe write down the scenario step-by-step when this doesn't work, so maybe we can figure out a fix without more exceptional cases in scripts?

test/Scripts.Integration.Test/update-sentry.ps1
```diff
@@ -21,20 +21,23 @@ function RunUnityAndExpect([string] $name, [string] $successMessage, [string] $f
...
-Copy-Item "$PackageReleaseAssetsPath/Scripts/NativeSupport/CppPlugin.*" -Destination "$NewProjectAssetsPath/Scripts/"
-Copy-Item "$PackageReleaseAssetsPath/Scripts/NativeSupport/ObjectiveCPlugin.*" -Destination "$NewProjectAssetsPath/Scripts/"
+Write-Host -NoNewline "Copying Test Files"
+# TODO: Replace copying from sample project with actually importing the package samples
+New-Item -Path "$NewProjectAssetsPath" -Name "Scripts" -ItemType "directory"
+New-Item -Path "$NewProjectAssetsPath" -Name "Scenes" -ItemType "directory"
+Copy-Item -Recurse "$IntegrationScriptsPath/Editor/*" -Destination "$NewProjectAssetsPath/Editor/"
+Copy-Item -Recurse "$UnityOfBugsPath/Assets/Scripts/*" -Destination "$NewProjectAssetsPath/Scripts/"
+Copy-Item -Recurse "$IntegrationScriptsPath/Scripts/*" -Destination "$NewProjectAssetsPath/Scripts/"
+Copy-Item -Recurse "$UnityOfBugsPath/Assets/Scenes/*" -Destination "$NewProjectAssetsPath/Scenes/"
```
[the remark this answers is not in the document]
bitsandfoxes: That's the plan for the next step. I want the SentrySetup to import the samples from the package instead of the script copying it manually.

test/Scripts.Integration.Test/Editor/Builder.cs
bitsandfoxes: So the way I had it running:
- `pwsh create-project.ps1 'unityPath'`
- `pwsh update-sentry.ps1 'unityPath`
- `pwsh build-project.ps1 'unityPath'`

And `build-project` would fail the build. I was running it without `-CheckSymbols` but the upload is enabled by default. `sentry-cli` would fail to upload due to dummy values.

vaind: Why not just call `pwsh build-project.ps1 'unityPath' -CheckSymbols` instead? That's how it's supposed to be run :)

But yeah, I get your point, let me check the other places you had to change to make this work.

vaind: Coming back to this after a bit of time to think it through: I'm still a little torn so I'll leave it to you.

The problematic point for me is that this change will silence an issue in CI if `-CheckSymbols` won't be passed (it's conditionally done so right now). Also, it wipes the differentiation we had between configuring the project (in `update-sentry.ps` and building it).

bitsandfoxes: I see your point.
I've updated the PR to, instead of passing the -CheckSymbol to the `Builder` it instead gets passed to the `sentry_update`. So we only set up the cli options to try to upload symbols if we're actually checking for them.

vaind (approved):

vaind (approved):

test/Scripts.Integration.Test/Editor/Builder.cs
```diff
@@ -16,21 +16,21 @@ public static void BuildIl2CPPPlayer(BuildTarget target, BuildTargetGroup group)
...
             locationPathName = args["buildPath"],
             target = target,
             targetGroup = group,
             options = BuildOptions.StrictMode,
         };
 
+        if(File.Exists("Assets/Scenes/SmokeTest.unity"))
+        {
+            buildPlayerOptions.scenes = new[] { "Assets/Scenes/SmokeTest.unity" };
+        }
```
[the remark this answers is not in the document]
bitsandfoxes: It's not there when we build the clean project on main.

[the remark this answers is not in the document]
bitsandfoxes: We use the `Builder.cs` to build a project without Sentry. In that case, we have not yet copied the SmokeTest-related scripts/scenes and the build would fail.
https://github.com/getsentry/sentry-unity/blob/1fb22e5f6b1517d3e9f8eef4758bb6d1cd207c0e/.github/workflows/ci.yml#L240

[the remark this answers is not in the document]
vaind: Is there actually anything related to smoke tester in that scene? AFAIK it's only the SmokeTester.cs that's needed. Can we thus have the scene there all the time? I assume the game scene doesn't influence that? 

My POV of what the integration test with those two builds should do is test that if we add Sentry to an existing game, then it works fine and doesn't break. In that case it makes sense for the project only to change as necessary between the two builds.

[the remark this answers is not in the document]
bitsandfoxes: The way it's set up right now is that the `SmokeTest.scene` has a GameObject with the `SmokeTester.cs` component. And the SmokeTester has Sentry as a dependency, which is why it gets added during the Sentry SDK setup and why it gets added only after we install the package.

But this is a good point and will be a great optimization and cut down on Unity runs (which cuts down on time. A lot). I'll put that into its own issue.

PR #813, opened 2022-06-16
chore: remove version from the android .jar

vaind:
to avoid having to do this: https://github.com/getsentry/sentry-unity/pull/811/commits/afd1681ba2c64e820b58a26241e8153b7ee416e8

#skip-changelog


bruno-garcia (approved):
Downside is that we can't spot the version by looking at the files, but worth for reducing the maintenance cost

PR #817, opened 2022-06-17
chore: Comments

bitsandfoxes:
Rider keeps complaining about 'em. #skip-changelog


vaind (approved):
How about we include some proper linter in the CI? would clang-format work well with c#?

Issue #480, opened 2021-12-16
TestRunner not found on a new cloned branch.


lucas-zimerman:
using the command `dotnet clean` on a new cloned branch gives the following error:
```
Build FAILED.

D:\Unity\sentry-unity\test\Directory.Build.props(25,5): error : TestRunner not found. Expected: D:\Unity\sentry-unity/temp/unity-packages/Library/ScriptAssemblies/UnityEngine.TestRunner.dll [D:\Unity\sentry-unity\test\Sentry.Unity.Android.Tests\Sentry.Unity.Android.Tests.csproj]
D:\Unity\sentry-unity\test\Directory.Build.props(25,5): error : TestRunner not found. Expected: D:\Unity\sentry-unity/temp/unity-packages/Library/ScriptAssemblies/UnityEngine.TestRunner.dll [D:\Unity\sentry-unity\test\Sentry.Unity.Tests\Sentry.Unity.Tests.csproj]
D:\Unity\sentry-unity\test\Directory.Build.props(25,5): error : TestRunner not found. Expected: D:\Unity\sentry-unity/temp/unity-packages/Library/ScriptAssemblies/UnityEngine.TestRunner.dll [D:\Unity\sentry-unity\test\Sentry.Unity.iOS.Tests\Sentry.Unity.iOS.Tests.csproj]
D:\Unity\sentry-unity\test\Directory.Build.props(25,5): error : TestRunner not found. Expected: D:\Unity\sentry-unity/temp/unity-packages/Library/ScriptAssemblies/UnityEngine.TestRunner.dll [D:\Unity\sentry-unity\test\Sentry.Unity.Editor.iOS.Tests\Sentry.Unity.Editor.iOS.Tests.csproj]
D:\Unity\sentry-unity\test\Directory.Build.props(25,5): error : TestRunner not found. Expected: D:\Unity\sentry-unity/temp/unity-packages/Library/ScriptAssemblies/UnityEngine.TestRunner.dll [D:\Unity\sentry-unity\test\Sentry.Unity.Editor.Tests\Sentry.Unity.Editor.Tests.csproj]
    0 Warning(s)
    5 Error(s)
```
Running `dotnet build` afterwards may also give the same error but on the second try the build will complete without errors.

vaind:
This is again the issue.

@bitsandfoxes any reason why we have to build the project instead of grabbing it from one of the Unity editor templates?

[...]

PR #818, opened 2022-06-17
feat: Remove temp project creation from build process

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/480
The idea is to fetch the `TestRunner.dlls` from the template cache that comes with the installation. That way we no longer need to create a whole new project just to fetch it from the `Library` and we save some time and licenses in CI.

#skip-changelog


vaind (approved):
Hooray

Directory.Build.targets
```diff
@@ -402,6 +383,32 @@ if (version == null)
...
+        <![CDATA[
+var directories = Directory.GetDirectories(UnityLibcache, "com.unity.template.3d-*");
+if (directories.Length == 1)
+{
+    TestRunnerPath = Path.Combine(directories[0], "ScriptAssemblies");
+    Log.LogMessage("Found TestRunner path at: " + TestRunnerPath);
+    return true;
+}
+
+Log.LogError("Failed to resolve 'com.unity.template.3d-*' for TestRunner at: " + UnityLibcache);
```
bruno-garcia, line 411: Is the package name hard-coded here a possible pitfall?

bruno-garcia (approved):

Directory.Build.targets
vaind: fine by me, if it breaks in a future Unity version, it's an easy fix.

vaind (approved):

PR #831, opened 2022-06-20
use MSbuild Copy

SimonCropp:
#skip-changelog

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -200,7 +200,9 @@ Expected to exist:
     </ItemGroup>
 
     <Copy SourceFiles="@(AndroidSdkArtifacts)" DestinationFiles="@(AndroidSdkArtifacts->'$(SentryAndroidArtifactsDestination)%(RecursiveDir)%(Filename)%(Extension)')" />
-    <Exec WorkingDirectory="$(SentryAndroidRoot)" Command="cp sentry/build/libs/sentry*.jar $(SentryAndroidArtifactsDestination)sentry.jar" />
+
+    <Copy SourceFiles="$(SentryAndroidRoot)/sentry/build/libs/sentry*.jar"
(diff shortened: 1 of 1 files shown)


bruno-garcia (approved):

Directory.Build.targets
```diff
@@ -200,7 +200,9 @@ Expected to exist:
     </ItemGroup>
 
     <Copy SourceFiles="@(AndroidSdkArtifacts)" DestinationFiles="@(AndroidSdkArtifacts->'$(SentryAndroidArtifactsDestination)%(RecursiveDir)%(Filename)%(Extension)')" />
-    <Exec WorkingDirectory="$(SentryAndroidRoot)" Command="cp sentry/build/libs/sentry*.jar $(SentryAndroidArtifactsDestination)sentry.jar" />
+
+    <Copy SourceFiles="$(SentryAndroidRoot)/sentry/build/libs/sentry*.jar"
```
bruno-garcia, line 204: Wildcards don't work with that task do it?

SimonCropp: do we need the wildcard? is the file name not known?

vaind: The source name contains a version number while we want the target without that.

vaind (changes requested):
The changes are not just cosmetic but change the behaviour. I'm also missing the reason why this needs to be changed. It currently runs fine on all 3 desktop platforms in either powershell or bash/zsh. If we know whats the issue maybe we can find a solution that doesn't break the rest/changes the expected behaviour

SimonCropp:
@vaind  `cp` wont run on windows

vaind:
> 

I'm on windows and yes it does, you just have to use powershell instead of batch

SimonCropp:
forme, i am in Rider. and the build fails

PR #833, opened 2022-06-21
fix: IL2CPP opt-in by default

bitsandfoxes:
IL2CPP line number support should be opt-in by default for now.
#skip-changelog


vaind (approved):
OK, but why? Would be nice to know the reason without digging further through other PRs. Can you please update the description?

PR #834, opened 2022-06-21
fix: Adding the IL2CPP processor outside the constructor

bitsandfoxes:
Checking the opt-in flag in the constructor does not work since it would have to be passed as an argument.

#skip-changelog

### DIFF
--- a/src/Sentry.Unity/ScriptableSentryUnityOptions.cs
+++ b/src/Sentry.Unity/ScriptableSentryUnityOptions.cs
@@ -93,6 +93,9 @@ namespace Sentry.Unity
         {
             application ??= ApplicationAdapter.Instance;
 
+            UnityEngine.Debug.Log("creating new sentry unity options");
+            UnityEngine.Debug.Log("il2cpp enabled: " + Il2CppLineNumberSupportEnabled);
+
             var options = new SentryUnityOptions(isBuilding, unityInfo, application)
             {
                 Enabled = Enabled,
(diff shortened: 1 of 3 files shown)


src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -133,22 +142,16 @@ internal SentryUnityOptions ToSentryUnityOptions(bool isBuilding, ISentryUnityIn
...
-            // Because SentryOptions.Debug is used inside the .NET SDK to setup the ConsoleLogger we
-            // need to set it here directly.
-            options.Debug = ShouldDebug(application.IsEditor && !isBuilding);
-            options.DiagnosticLevel = DiagnosticLevel;
-
             options.SetupLogging();
 
             OptionsConfiguration?.Configure(options);
 
+            // Doing this after the configure callback to allow users to programmatically opt out
```
vaind, line 149: This would apply for any other integration we also add by default... I wonder, again, do we **really** need to make Il2CppLineNumberSupportEnabled opt-in? What is the risk if we enabled it by default? IMO it would simply situations like this, where we normally enable all integrations and let users opt-out by removing them if needed.

Swatinem: The fact that this was crashing in Unity 2020 because of mismatched fn signatures is a clear indication that this does need a bit more conservative testing before we turn it on by default ;-)

I also wonder what the situation would look like going forward. As we call into undocumented/unsupported internals, a Unity update could potentially break us as well.

vaind: AFAIK an integration could add an event/exception processor during setup. That looks like a "proper" way to do it. But LGTM at the moment

vaind (approved):

PR #836, opened 2022-06-22
fix: Correctly get native stack trace on Unity 2020

Swatinem:
Turns out that the libil2cpp method did change between 2020 and 2021, where it got a new parameter. We now add an ifdef-d shim for 2020 to work with the same signature.

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -114,7 +114,11 @@ namespace Sentry.Unity
 #else
             = new Il2CppMethods(
                 il2cpp_gchandle_get_target,
+#if UNITY_2021_3_OR_NEWER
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,10 @@
   - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#220)
   - [diff](https://github.com/getsentry/sentry-cli/compare/1.71.0...2.2.0)
 
+### Fixes
+
+- Correctly get native stack trace on Unity 2020 ([#836](https://github.com/getsentry/sentry-unity/pull/836))
+
```
bitsandfoxes, line 26:
```suggestion
```
Since this has not been released yet it did technically not break. No break - no fix :D

PR #853, opened 2022-06-24
fix: Avoid crashes on Unity 2020 when capturing native stack

Swatinem:
Unity 2020 does not *return* a newly allocated string as out-parameter,
but rather expects a pre-allocated buffer it writes into.

#skip-changelog

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -145,9 +145,14 @@ namespace Sentry.Unity
             imageName = null;
             // Unity 2020 does not *return* a newly allocated string as out-parameter,
             // but rather expects a pre-allocated buffer it writes into.
+            // That buffer needs to have space for the hex-encoded uuid (32) plus
+            // terminating nul-byte.
             char[] uuidBuffer = new char[32 + 1];
             il2cpp_native_stack_trace(exc, out addresses, out numFrames, uuidBuffer);
-            imageUUID = new string(uuidBuffer);
+            // C-strings are nul-terminated, but the conversion here would
+            // normally keep that terminating nul-byte in the string, which
+            // we don't want.
+            imageUUID = new string(uuidBuffer).TrimEnd('\0');
         }
 
         // Definition from Unity `2020.3`:
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -13,6 +13,9 @@
 #endif
 
 using System;
+#if !UNITY_2021_3_OR_NEWER
+using System.Buffers;
+#endif
 using System.Runtime.InteropServices;
 using UnityEngine;
 using UnityEngine.Scripting;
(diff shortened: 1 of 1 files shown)


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -140,13 +143,17 @@ private Il2CppMethods _il2CppMethods
         private static void Il2CppNativeStackTraceShim(IntPtr exc, out IntPtr addresses, out int numFrames, out string? imageUUID, out string? imageName)
         {
             imageName = null;
-            il2cpp_native_stack_trace(exc, out addresses, out numFrames, out imageUUID);
+            // Unity 2020 does not *return* a newly allocated string as out-parameter,
+            // but rather expects a pre-allocated buffer it writes into.
+            char[] uuidBuffer = new char[32 + 1];
```
bitsandfoxes, line 148: Where is the `32+1` coming from?

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -140,13 +143,17 @@ private Il2CppMethods _il2CppMethods
...
+            il2cpp_native_stack_trace(exc, out addresses, out numFrames, uuidBuffer);
+            imageUUID = new string(uuidBuffer);
         }
 
         // Definition from Unity `2020.3`:
         // void il2cpp_native_stack_trace(const Il2CppException * ex, uintptr_t** addresses, int* numFrames, char* imageUUID)
         [DllImport("__Internal")]
-        private static extern void il2cpp_native_stack_trace(IntPtr exc, out IntPtr addresses, out int numFrames, out string? imageUUID);
+        private static extern void il2cpp_native_stack_trace(IntPtr exc, out IntPtr addresses, out int numFrames, [Out] char[] imageUUID);
```
bitsandfoxes, line 156: Why not `out char[]`?

bitsandfoxes (approved):

package-dev/Runtime/SentryInitialization.cs
Swatinem: its the uuid + terminating nul. I added a comment to explain.

package-dev/Runtime/SentryInitialization.cs
bruno-garcia: The caller doesn't pass a `string` then, does it? Because IIRC it'll compile passing to `char[]` and the link you shared says: 

>❌ DO NOT use [Out] string parameters. String parameters passed by value with the [Out] attribute can destabilize the runtime if the string is an interned string. See more information about string interning in the documentation for

Swatinem: It uses a fixed-size `char[]`, which corresponds partly to this doc comment:

> CONSIDER char[] or byte[] arrays from an ArrayPool when native code is expected to fill a character buffer. This requires passing the argument as [Out].

I tried `ArrayPool` at first, but that doesn’t seem necessary since the size of the buffer is statically known.

PR #858, opened 2022-06-27
chore: Cleaned Build.targets paths 

bitsandfoxes:
By convention, we try to end paths with a slash, so tried getting rid of most cases of double-slashes.

#skip-changelog

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -15,17 +15,21 @@
     <SentryArtifactsDestination>$(RepoRoot)package-dev/Plugins/</SentryArtifactsDestination>
     <!-- Cocoa -->
     <SentryCocoaRoot>$(RepoRoot)modules/sentry-cocoa/</SentryCocoaRoot>
-    <SentryiOSDeviceArtifactsDestination>$(SentryArtifactsDestination)/iOS/Device/Sentry.framework/</SentryiOSDeviceArtifactsDestination>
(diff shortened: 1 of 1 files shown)


Directory.Build.targets
```diff
@@ -15,18 +15,22 @@
     <SentryArtifactsDestination>$(RepoRoot)package-dev/Plugins/</SentryArtifactsDestination>
     <!-- Cocoa -->
     <SentryCocoaRoot>$(RepoRoot)modules/sentry-cocoa/</SentryCocoaRoot>
-    <SentryiOSDeviceArtifactsDestination>$(SentryArtifactsDestination)/iOS/Device/Sentry.framework/</SentryiOSDeviceArtifactsDestination>
-    <SentryiOSSimulatorArtifactsDestination>$(SentryArtifactsDestination)/iOS/Simulator/Sentry.framework/</SentryiOSSimulatorArtifactsDestination>
-    <SentrymacOSArtifactsDestination>$(SentryArtifactsDestination)/macOS/Sentry/</SentrymacOSArtifactsDestination>
+    <SentryiOSDeviceArtifactsDestination>$(SentryArtifactsDestination)iOS/Device/Sentry.framework/</SentryiOSDeviceArtifactsDestination>
```
vaind, line 18: personally, I find this less readable

bruno-garcia: It's best not to have variable+ `/path` because if for some reason we don't resolve the variable, it ends up as a rooted path.

bruno-garcia (approved):
LGTM, unless @vaind has a reason for us not to go ahead

PR #747, opened 2022-05-12
feat: Synchronize contexts from .NET to native

vaind:
closes #650 

* [x] [Windows](https://sentry.io/organizations/sentry-sdks/issues/3110234864/events/e664d539bd694b16b62c0a58874ddca7/?project=5439417)
* [x] [Linux](https://sentry.io/organizations/sentry-sdks/issues/3262191687/events/b12969783de64aff74e667b06dd6cf8c/?project=5439417)
* [x] [MacOS](https://sentry.io/organizations/sentry-sdks/issues/2677827744/events/4e116a0ea7a1464bbe7f72169b075c58/?project=5439417) - GPU & Unity contexts only - currently can't update existing contexts - see https://discord.com/channels/621778831602221064/814577827043082250/976931491357552741
* [x] [iOS](https://sentry.io/organizations/sentry-sdks/issues/2677827744/events/54f1ddd98d4247959d8e3c54eb0d8c29/?project=5439417) - same remarks as macOS
* [x] [Android java](https://sentry.io/organizations/sentry-sdks/issues/3245458320/events/83f3444310064093b438ce3226c547c0/?project=5439417)
* [x] [Android native](https://sentry.io/organizations/sentry-sdks/issues/3329070559/events/4cb3f2892a9e41e531b7a4cb027a1149/?project=5439417)

PR #861, opened 2022-06-29
fix: use LogWarning instead of LogError for potential SDK errors

### LINKED ISSUE #839: SDK internal error logged through Debug.LogError
We're logging an SDK internal error through the Unity logger here https://github.com/getsentry/sentry-unity/blob/6322d6b6e5701d8dad143eac0efe0453f1b84b20/src/Sentry.Unity.Android/SentryJava.cs#L154

`Error in SentryJava.ScopeCallback: UnityEngine.AndroidJavaException: java.lang.NoSuchMethodError: no non-static method with name='getContexts' signature='()Ljava/lang/Object;'`

https://sentry.io/organizations/sentry-sdks/issues/3369803070/?project=5439417&query=is%3Aunresolved

vaind:
closes #839 

#skip-changelog

### DIFF
--- a/src/Sentry.Unity.Android/SentryJava.cs
+++ b/src/Sentry.Unity.Android/SentryJava.cs
@@ -151,7 +151,8 @@ namespace Sentry.Unity.Android
                 }
                 catch (Exception e)
                 {
-                    UnityEngine.Debug.LogWarning($"Error in SentryJava.ScopeCallback: {e}");
+                    // Adding the Sentry logger prefix ensures we don't send this error to Sentry.
+                    UnityEngine.Debug.LogError($"{UnityLogger.LogPrefix}Error in SentryJava.ScopeCallback: {e}");
                 }
                 return null;
             }
--- a/src/Sentry.Unity.Android/SentryJava.cs
+++ b/src/Sentry.Unity.Android/SentryJava.cs
@@ -151,7 +151,7 @@ namespace Sentry.Unity.Android
                 }
                 catch (Exception e)
                 {
-                    UnityEngine.Debug.LogError($"Error in SentryJava.ScopeCallback: {e}");
+                    UnityEngine.Debug.LogWarning($"Error in SentryJava.ScopeCallback: {e}");
                 }
                 return null;
             }
--- a/src/Sentry.Unity/Json/SafeSerializer.cs
+++ b/src/Sentry.Unity/Json/SafeSerializer.cs
@@ -25,7 +25,7 @@ namespace Sentry.Unity.Json
             }
             catch (Exception e)
             {
-                logger?.LogError("Failed to serialize value of type \"{0}\"", e, value.GetType());
+                logger?.LogWarning("Failed to serialize value of type \"{0}\"", e, value.GetType());
                 return null;
             }
         }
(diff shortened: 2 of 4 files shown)


src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -151,7 +151,7 @@ public ScopeCallback(Action<AndroidJavaObject> callback) : base("io.sentry.Scope
                 }
                 catch (Exception e)
                 {
-                    UnityEngine.Debug.LogError($"Error in SentryJava.ScopeCallback: {e}");
+                    UnityEngine.Debug.LogWarning($"Error in SentryJava.ScopeCallback: {e}");
```
bitsandfoxes, line 154: I thought this was about using `UnityEngine.Debug` instead of the SentryLogger and not about LogError itself?

vaind: ah, right, diagnostic logger doesn't send errors, right?

src/Sentry.Unity/Json/SafeSerializer.cs
```diff
@@ -25,7 +25,7 @@ internal static class SafeSerializer
             }
             catch (Exception e)
             {
-                logger?.LogError("Failed to serialize value of type \"{0}\"", e, value.GetType());
+                logger?.LogWarning("Failed to serialize value of type \"{0}\"", e, value.GetType());
```
vaind, line 28:
```suggestion
                logger?.LogError("Failed to serialize value of type \"{0}\"", e, value.GetType());
```

src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -28,7 +28,7 @@ public SentryEvent Process(SentryEvent @event)
             }
             catch (Exception ex)
             {
-                _sentryOptions.DiagnosticLogger?.LogError("{0} processing failed.", ex, nameof(SentryEvent));
+                _sentryOptions.DiagnosticLogger?.LogWarning("{0} processing failed.", ex, nameof(SentryEvent));
```
vaind, line 31:
```suggestion
                _sentryOptions.DiagnosticLogger?.LogError("{0} processing failed.", ex, nameof(SentryEvent));
```

bruno-garcia (approved):

bruno-garcia:
Actually. Does it make sense to be an error? We'd like to warn the user that somethjing went wrong here so we can be aware

vaind:
> Actually. Does it make sense to be an error? We'd like to warn the user that somethjing went wrong here so we can be aware

It does make sense, that's why it was an error. Whether it should still be is up to you. See https://github.com/getsentry/sentry-unity/pull/747#discussion_r903926267

bitsandfoxes (approved):

src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -151,7 +151,8 @@ public ScopeCallback(Action<AndroidJavaObject> callback) : base("io.sentry.Scope
                 }
                 catch (Exception e)
                 {
-                    UnityEngine.Debug.LogError($"Error in SentryJava.ScopeCallback: {e}");
+                    // Adding the Sentry logger prefix ensures we don't send this error to Sentry.
```
bitsandfoxes, line 154: <3

PR #864, opened 2022-06-30
feat: Breadcrumbs on button clicks

bitsandfoxes:
The idea is to auto-instrument all buttons present in scenes during build time (the `OnProcessScene` also gets called within the Editor).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -43,10 +43,63 @@
     <Error Condition="'$(UnityManagedPath)' == ''" Text="'UnityManagedPath' not defined. Can't find UnityEngine.dll."></Error>
     <Error Condition="!Exists('$(UnityManagedPath)/UnityEngine.dll')" Text="Couldn't find UnityEngine at $(UnityManagedPath)/UnityEngine.dll."></Error>
     <ItemGroup>
+<!--      <Reference Include="UnityEngine">-->
+<!--        <HintPath>$(UnityManagedPath)/UnityEngine.dll</HintPath>-->
+<!--        <Private>false</Private>-->
+<!--      </Reference>-->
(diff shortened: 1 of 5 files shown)


Directory.Build.props
```diff
@@ -43,10 +43,63 @@
     <Error Condition="'$(UnityManagedPath)' == ''" Text="'UnityManagedPath' not defined. Can't find UnityEngine.dll."></Error>
     <Error Condition="!Exists('$(UnityManagedPath)/UnityEngine.dll')" Text="Couldn't find UnityEngine at $(UnityManagedPath)/UnityEngine.dll."></Error>
     <ItemGroup>
+<!--      <Reference Include="UnityEngine">-->
+<!--        <HintPath>$(UnityManagedPath)/UnityEngine.dll</HintPath>-->
```
bitsandfoxes, line 47: That here needs some cleaning up but the gist of it is: `ManagedPath/UnityEngine.dll` on the managed root contains all the type forwarders to the modules inside of `ManagedPath/UnityEngine/`. This does not seem to work 100% of the time like with LogAssert in the test projs. Instead, we could just reference the individual modules.

src/Sentry.Unity/SentryScene.cs
```diff
@@ -0,0 +1,10 @@
+using UnityEngine;
+
+namespace Sentry.Unity
+{
+    public class SentryScene : MonoBehaviour
```
bitsandfoxes, line 5: Adding a callback to a button relies on a MonoBehaviour being a permanent part of the scene. `SentryMonoBehaviour` get created during runtime and the callback would fail.

bitsandfoxes:
Closing this as `stale`.

PR #865, opened 2022-06-30
fix: Don't access disk on unknown platforms

bitsandfoxes:
Apparently, `Application.persistentDataPath` not only returns a path but also creates the directory. This leads to issues on platforms that impose restrictions on fileIO, like the Switch.


test/Sentry.Unity.Android.Tests/Sentry.Unity.Android.Tests.csproj
```diff
@@ -6,5 +6,6 @@
     <ProjectReference Include="../../src/sentry-dotnet/src/Sentry/Sentry.csproj" Private="false" />
     <ProjectReference Include="../../src/Sentry.Unity/Sentry.Unity.csproj" Private="false" />
     <ProjectReference Include="../../src/Sentry.Unity.Android/Sentry.Unity.Android.csproj" Private="false" />
+    <ProjectReference Include="../Sentry.Unity.Tests/Sentry.Unity.Tests.csproj" />
```
bruno-garcia, line 9: I believe this can cause issues. In .NET IIRC this would return the tests twice. Does it work OK with Unity?
Often in this case we have a test helper class. e.g: 

https://github.com/getsentry/sentry-dotnet/tree/main/test/Sentry.Testing

bruno-garcia (commented):
I feel we could get away without changing the SentryInitialization class. If we pass down the platform through inside the SDK we can just avoid assigning the application folder (reading from Application.get...

test/Sentry.Unity.Android.Tests/Sentry.Unity.Android.Tests.csproj
bitsandfoxes: It does not seem to cause any issues in Unity. We do have those `SharedClasses` so we could even move it there.

bitsandfoxes:
> I feel we could get away without changing the SentryInitialization class. If we pass down the platform through inside the SDK we can just avoid assigning the application folder (reading from Application.get...

Derp. I was so sure that it would be hidden behind the Switch SDK that I didn't even check for `Runtime.Switch`.

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -175,7 +176,10 @@ internal SentryUnityOptions(SentryMonoBehaviour behaviour, IApplication applicat
                 ? "editor"
                 : "production";
 
-            CacheDirectoryPath = application.PersistentDataPath.Trim();
+            if (application.Platform is not RuntimePlatform.Switch)
```
vaind, line 179: The first stab at this was through the `UnknownPlatform.cs`, for any platform that's not explicitly known to work. If the issue is with creating a directory, I'd suggest moving [that code](https://github.com/getsentry/sentry-unity/blob/f2c2d3aeda3afe94ca5bac5941e845c02b6f6b94/src/Sentry.Unity/Default/UnknownPlatform.cs#L24_L29) here instead

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 ### Fixes
 
 - Only use application-not-responding detection in C#, disable in native plugins ([#852](https://github.com/getsentry/sentry-unity/pull/852))
+- The SDK no longer accesses the disk on unknown platforms (i.e. Switch) ([#865](https://github.com/getsentry/sentry-unity/pull/865))
```
vaind, line 31:
```suggestion
- The SDK no longer accesses the disk on unknown platforms (e.g. Switch) ([#865](https://github.com/getsentry/sentry-unity/pull/865))
```

vaind (approved):
LGTM

src/Sentry.Unity/SentryUnityOptions.cs
bitsandfoxes: Fair point. I think we can use `ToSentryOptions` from now on for more than just mapping from the `ScriptableOptions`to the `SentryOptions`.

PR #870, opened 2022-07-05
chore: disable cocoa ANR detection

### LINKED ISSUE #838: ANR: Disable native in favor of C#
We're currently missing an opt-out for ANR on the native layer. `RemoveIntegration<>` does not work since the setup for mobile happens during build time.

- [x] Android
- [x] iOS

vaind:
closes #838 

#skip-changelog (no user visible change because the native cocoa SDK with the ANR enabled hasn't shipped yet)

### DIFF
--- a/package-dev/Plugins/macOS/SentryNativeBridge.m
+++ b/package-dev/Plugins/macOS/SentryNativeBridge.m
@@ -46,6 +46,8 @@ const void *SentryNativeBridgeOptionsNew()
 {
     NSMutableDictionary *dictOptions = [[NSMutableDictionary alloc] init];
     dictOptions[@"sdk"] = @ { @"name" : @"sentry.cocoa.unity" };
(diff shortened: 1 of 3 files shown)


package-dev/Plugins/macOS/SentryNativeBridge.m
```diff
@@ -46,6 +46,8 @@ int SentryNativeBridgeLoadLibrary()
 {
     NSMutableDictionary *dictOptions = [[NSMutableDictionary alloc] init];
     dictOptions[@"sdk"] = @ { @"name" : @"sentry.cocoa.unity" };
+    dictOptions[@"enableAutoSessionTracking"] = @NO;
```
vaind, line 49: I've noticed this could be here statically, instead of setting it with a native interop call

bitsandfoxes (approved):
Thanks!

PR #871, opened 2022-07-06
fix: Added a x64 guard to the IL2CPP methods

bitsandfoxes:
There is something off about the way we call native code on `x86` so I'm guarding the IL2CPP feature with `UNITY_64` for now:
```
error LNK2019: unresolved external symbol _il2cpp_gchandle_get_target@4 referenced in function _SentryUnityInfo_il2cpp_gchandle_get_target_m42BE043B96499DAF199C1045023BE6DE1B0FF6EA
  Hint on symbols that are defined and could potentially match:
    _il2cpp_gchandle_get_target
```
This can easily be reproduced with our own samples and `throw_cpp`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fixed linker issues for x86 IL2CPP builds on 2020.3 and newer ([#871](https://github.com/getsentry/sentry-unity/pull/871))
+- Explicitly disable Windows x86 native-error tracking and IL2CPP processing integration ([#871](https://github.com/getsentry/sentry-unity/pull/871))
 
 ## 0.20.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Fixed linker issues for x86 IL2CPP builds on 2020.3 and newer ([#871](https://github.com/getsentry/sentry-unity/pull/871))
```
vaind, line 5:
```suggestion
- Explicitly disable Windows x86 native-error tracking integration ([#871](https://github.com/getsentry/sentry-unity/pull/871))
```

vaind (approved):

PR #872, opened 2022-07-06
fix: Correctly byte-swap ELF build-ids

Swatinem:
Debug-IDs on Android/Linux need to have some parts byte-swapped, as explained here:
https://getsentry.github.io/symbolicator/advanced/symbol-server-compatibility/#identifiers


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -107,19 +107,45 @@ public string Platform
...
 
+        private static string SanitizeDebugId(string debugId)
+        {
+#if UNITY_ANDROID || UNITY_STANDALONE_LINUX
+            // For ELF platforms, the 20-byte `NT_GNU_BUILD_ID` needs to be
+            // turned into a "little-endian GUID", which means the first three
+            // components need to be byte-swapped appropriately.
+            // See: https://getsentry.github.io/symbolicator/advanced/symbol-server-compatibility/#identifiers
+
+            // As our id is already stringified, we will use substrings here,
```
bruno-garcia, line 127: It's strigified already below:

```
            imageUUID = new String(uuidBuffer, 0, Array.IndexOf(uuidBuffer, '\0'));
```

Is it possible to do this without the 18+ heap allocations?

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -142,12 +178,20 @@ private Il2CppMethods _il2CppMethods
...
-            // That buffer needs to have space for the hex-encoded uuid (32) plus terminating nul-byte.
-            var uuidBuffer = new char[32 + 1];
+            // Unity 2020 does not *return* a newly allocated string as out-parameter,
+            // but rather expects a pre-allocated buffer it writes into.
+            // That buffer needs to have space for either:
+            // - A hex-encoded `LC_UUID` on MacOS (32)
+            // - A hex-encoded GUID + Age on Windows (40)
+            // - A hex-encoded `NT_GNU_BUILD_ID` on ELF (Android/Linux) (40)
+            // plus a terminating nul-byte.
+            var uuidBuffer = new char[40 + 1];
```
vaind, line 188: `char` is 2 bytes in c# but we're passing to C `char` in `const Il2CppException * ex, uintptr_t** addresses, int* numFrames, char* imageUUID)`, We should be using `byte[]` here and in the other call to `il2cpp_native_stack_trace` instead. If we do, we can than make `SanitizeDebugId` also take the same `byte` array, making it possible to efficiently swap the bytes as needed.

Conversion to string can happen when needed.

bitsandfoxes: @Swatinem this might be why it's not going out of bounds previously when giving it a buffer of 32+1.

Swatinem: Indeed, internally `Il2CppChar` is defined as u16 like this:

```
#if _MSC_VER
typedef wchar_t Il2CppChar;
#elif __has_feature(cxx_unicode_literals)
typedef char16_t Il2CppChar;
#else
typedef uint16_t Il2CppChar;
#endif
```

So I’m wondering how this could have ever worked.

Swatinem:
I have spent yesterday going over this implementation, and now both 2020 and 2021 code paths deal with raw pointers `char*`, do the byte flipping based on those, and convert them to C# strings in the end.

@vaind @bruno-garcia please take another look.

@bitsandfoxes has been testing this on various platforms and things seem to work fine, in particular Android/Linux now gets a correct debug-id.

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -107,19 +107,60 @@ public string Platform
...
+        // nul-terminated C string. This function converts it to a C# string,
+        // and also byte-swaps/truncates on ELF platforms.
+        private static string? SanitizeDebugId(IntPtr debugIdPtr)
+        {
+            if (debugIdPtr == IntPtr.Zero)
+            {
+                return null;
+            }
+
+#if UNITY_ANDROID || UNITY_STANDALONE_LINUX
```
vaind, line 130: just checking - is there any diference between 32 and 64-bit android with regards to these UUIDs

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -131,32 +172,48 @@ private Il2CppMethods _il2CppMethods
...
+            // - A hex-encoded GUID + Age on Windows (40)
+            // - A hex-encoded `NT_GNU_BUILD_ID` on ELF (Android/Linux) (40)
+            // plus a terminating nul-byte.
+            var uuidBuffer = il2cpp_alloc(40 + 1);
             il2cpp_native_stack_trace(exc, out addresses, out numFrames, uuidBuffer);
-            // C-strings are nul-terminated, but the conversion here would normally keep that terminating nul-byte in the string, which we don't want.
-            imageUUID = new string(uuidBuffer).TrimEnd('\0');
+
+            imageUUID = SanitizeDebugId(uuidBuffer);
+            il2cpp_free(uuidBuffer);
```
vaind, line 203:
```suggestion
            try 
            {
                imageUUID = SanitizeDebugId(uuidBuffer);
            }
            finally
            {
                il2cpp_free(uuidBuffer);
            }
```

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/master/CHANGELOG.md#7200)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/7.19.0...7.20.0)
 
+### Fixes
+
+- Correctly byte-swap ELF build-ids ([#872](https://github.com/getsentry/sentry-unity/pull/872))
+
```
vaind, line 17: fixes usually come before features

vaind (approved):

bitsandfoxes (approved):

PR #873, opened 2022-07-06
set tfm to ns2.0

bruno-garcia:


vaind (changes requested):
What's the goal here? This breaks Unity 2021 build because it requires v2.1

vaind:
> > What's the goal here? This breaks Unity 2021 build because it requires v2.1
> 
> Goal is to skip Sentry (the dotnet project added via submodule) from building all targets. In particular the net6.0-android so we can remove the workload. And avoid breakage when we add other targets that require workloads installed.

Got it... so I guess yes that works. The few ugly things about that are:
* it depends on Unity version which TFW we want to build for - also we have [this](https://github.com/getsentry/sentry-unity/blob/76bbdcc3904ac160fbb05feae5f940c9d89a267b/Directory.Build.props#L27) and [this](https://github.com/getsentry/sentry-unity/blob/be7bdf5d0d2bdb47f2197a44a93b1a45e454b0d8/test/Directory.Build.props#L13)
* only works in CI (or rather - it requires you to specify the when building), so local developers who just run `dotnet build` still get the slow build **and** must have the android workflow installed

Is there no other way to instruct dotnet to only build what's necessary from sentry-dotnet when building in the unity repo? I thought something like specifying a prefix for projects to build (`Sentry.Unity*`) but that doesn't seem to work...

So if this is just about CI and local developers still must have the workflow, I am not entirely sure it's worth doing (for the <1 minute of build time it would save). It would be awesome if we could get this working locally when running `dotnet build` - that would be the real time-saver during development

bitsandfoxes:
@mattjohnsonpint and I had a chat recently and he suggested that we could add something like `<TargetFrameworks Condition="!$(UNITY)">...` to the .NET repo to exclude everything the Unity SDK does not need.

mattjohnsonpint:
Closing in favor of https://github.com/getsentry/sentry-dotnet/pull/1780

PR #891, opened 2022-07-15
fix: Wizard auto select

bitsandfoxes:
Wizard now auto selects organization and project if there is only one option available.


vaind (approved):
just to make sure: have you tested manually on an account where you have multiple options? so that still works.

PR #905, opened 2022-07-22
feat: IL2CPP programmatic access & logging

bitsandfoxes:
The `IL2CPP line numbers` flag requires the debug symbol upload to run to actually work but previously didn't state so.
Additionally, this adds programmatic access to the flag.

### DIFF
--- a/src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
+++ b/src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
@@ -196,7 +196,7 @@ namespace Sentry.Unity.Editor.Android
                     symbolsUpload.TryCopySymbolsToGradleProject();
                 }
 
-                symbolsUpload.AppendUploadToGradleFile(sentryCliPath, _options!.Il2CppLineNumberSupportEnabled);
+                symbolsUpload.AppendUploadToGradleFile(sentryCliPath);
             }
             catch (Exception e)
             {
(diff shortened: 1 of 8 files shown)


src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
```diff
@@ -175,6 +175,12 @@ internal void SetupSymbolsUpload(string unityProjectPath, string gradleProjectPa
             if (disableSymbolsUpload)
             {
                 symbolsUpload.RemoveUploadFromGradleFile();
+
+                if (_options!.Il2CppLineNumberSupportEnabled)
```
bruno-garcia, line 179: This `!` looks dangerous. 

On line 160 you do `if (_options is null ||` then you `disableSymbolsUpload = true`
So it could be that options is null here. And the compiler knows, that's why you had to do `!`.

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -95,6 +95,11 @@ public new CompressionLevelWithAuto RequestBodyCompressionLevel
         /// </summary>
         public bool LinuxNativeSupportEnabled { get; set; } = true;
 
+        /// <summary>
+        /// Whether the SDK should add IL2CPP line number support
+        /// </summary>
```
bruno-garcia, line 100:
```suggestion
        /// </summary>
        /// <remarks>
        /// To give line numbers, Sentry requires the debug symbols Unity generates during build
        /// For that reason, uploading debug information files must be enabled. 
        /// For that, Org Slut, Project Slug and Auth token are required.
        /// </remarks>
```

bruno-garcia (changes requested):

src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
bitsandfoxes: Of course! Thanks for the catch!

bruno-garcia (approved):

PR #910, opened 2022-07-26
feat: Additional editor window utility

bitsandfoxes:
Added logging and icon highlights.

Highlight if the DSN is missing since it's the most crucial/fundamental part of the config:
image

Checking if the symbol upload is properly setup to set expectations:
image

In `Unity 2019`:
image

Symbol Upload Highlights for the required fields (best effort without actual validation of the input):
Editor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Features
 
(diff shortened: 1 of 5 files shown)


bruno-garcia (approved):
❤️

PR #913, opened 2022-07-27
feat: Adding il2cpp as integration to SdkVersion

bitsandfoxes:
#skip-changelog

### DIFF
--- a/src/Sentry.Unity/UnityEventProcessor.cs
+++ b/src/Sentry.Unity/UnityEventProcessor.cs
@@ -23,6 +23,9 @@ namespace Sentry.Unity
             try
             {
                 PopulateDevice(@event.Contexts.Device);
+                // Populating the SDK Integrations here (for now) instead of UnityScopeIntegration because it cannot be guaranteed
+                // that it got added last or that there was not an integration added at a later point
+                PopulateSdkIntegrations(@event.Sdk);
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):

src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -66,6 +69,13 @@ private void PopulateDevice(Device device)
             }
         }
 
+        private void PopulateSdkIntegrations(SdkVersion sdkVersion)
+        {
+            foreach (var integrationName in _sentryOptions.SdkIntegrationNames)
```
bruno-garcia, line 74: `SdkIntegrationNames` is of a mutable type, and if we iterate on it here while it's being changed on another thread, it'll throw.

Looking at the code here, it's only mutated in the event processor's constructor so it will be fine. But as a follow up we should take that into account (make `SdkIntegrationNames` copy on write

bruno-garcia (approved):

bruno-garcia:
> Unity failed because it couldn't acquire a license.
WebGL, just merging

src/Sentry.Unity/UnityEventProcessor.cs
bitsandfoxes: Good point! Thanks

PR #916, opened 2022-07-28
fix: Save SampleRate to Options.asset

bitsandfoxes:
Due to being nullable `SampleRate` never actually got saved to the SentryOptions.asset. 
We pass it as `null` to the actual SentryUnityOptions to skip event drop checks in case it's `1.0f`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+
+### Fixes
+
 - Save SampleRate to Options.asset ([#916](https://github.com/getsentry/sentry-unity/pull/916))
 
 ## 0.22.0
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Save SampleRate to Options.asset ([#916](https://github.com/getsentry/sentry-unity/pull/916))
```
bruno-garcia, line 5:
```suggestion

### Fixes

- Save SampleRate to Options.asset ([#916](https://github.com/getsentry/sentry-unity/pull/916))
```

bruno-garcia (approved):

Issue #866, opened 2022-07-01
Exporting to iOS project failed to build


shunia:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)

SaaS

Which version of the SDK?

0.19.0

How did you install the package? (Git-URL, Assetstore)

Git-URL

Which version of Unity?

2021.1.25f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?

Error only occurs in iOS project, we can not build the exported project, if remove Sentry, everything works fine;
Unity editor works as expected, exported Android Project works too. We are not aiming for other platforms.

### Steps to Reproduce
1. Setup Sentry followed by documentation;
2. Export iOS project to an existing iOS project (replace);
3. Build the iOS project in xCode as usual;

### Expected Result
Build pass and everything works.

### Actual Result
Failed when build iOS project:
```
ld: warning: object file (/Users/Clover/Desktop/_/ios_exported/Libraries/lib_burst_generated.a(lib_burst_generated_part_0_merged.o)) was built for newer iOS version (15.2) than being linked (11.0)
ld: warning: object file (/Users/Clover/Desktop/_/ios_exported/Libraries/lib_burst_generated.a(lib_burst_generated_part_4.o)) was built for newer iOS version (15.2) than being linked (11.0)
ld: warning: object file (/Users/Clover/Desktop/_/ios_exported/Libraries/lib_burst_generated.a(lib_burst_generated_part_1.o)) was built for newer iOS version (15.2) than being linked (11.0)
ld: warning: object file (/Users/Clover/Desktop/_/ios_exported/Libraries/lib_burst_generated.a(lib_burst_generated_part_3.o)) was built for newer iOS version (15.2) than being linked (11.0)
ld: warning: object file (/Users/Clover/Desktop/_/ios_exported/Libraries/lib_burst_generated.a(lib_burst_generated_part_2.o)) was built for newer iOS version (15.2) than being linked (11.0)
ld: warning: object file (/Users/Clover/Desktop/_/ios_exported/Libraries/lib_burst_generated.a(lib_burst_generated_part_0.o)) was built for newer iOS version (15.2) than being linked (11.0)
Undefined symbols for architecture arm64:
  "__NSNumberOrNil", referenced from:
      ___SentryNativeBridgeWriteScope_block_invoke in SentryNativeBridge.o
  "__NSBoolOrNil", referenced from:
[...]

PR #932, opened 2022-08-12
fix: Made inline methods in Sentry Native Bridge static

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/866

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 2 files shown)


bruno-garcia (approved):

vaind:
@bitsandfoxes can you reproduce the mentioned issue locally? If so, maybe try removing the leading underscore, which I guess could be the cause why the symbols are stripped somewhere in the build chain.

bitsandfoxes:
@vaind yes, it's reproducible locally. I gave your suggestion a try but the outcome is the exact same.

Issue #925, opened 2022-08-03
automatic screenshot feature creates an exception


johnfoconnor:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io

Which version of the SDK?
0.22.2
How did you install the package? Git-URL

Which version of Unity?
2021.3.5f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows? Editor (only tested in editor)

### Steps to Reproduce

1. Enable automatic screenshot capturing
2. "Play" my unity scene
3. Sentry detects a `Sentry: (Info) Detected an ANR event: Application not responding for at least 5000 ms.`

### Expected Result

Sentry should either

A) correctly capture the screenshot
B) not try to capture a screenshot if it cant (e.g. if it's not on the main thread)

### Actual Result

Sentry produces an exception and fails to capture a screenshot
```
Sentry: (Error) Failed to add attachment: screenshot.jpg. System.Exception: Sentry: cannot capture screenshot attachment on other than the main (UI) thread.
  at Sentry.Unity.ScreenshotAttachmentContent.GetStream () [0x0002d] in /sentry-unity/src/Sentry.Unity/ScreenshotAttachment.cs:38 
  at Sentry.Protocol.Envelopes.EnvelopeItem.FromAttachment (Sentry.Attachment attachment) [0x00000] in /sentry-unity/src/sentry-dotnet/src/Sentry/Envelopes/EnvelopeItem.cs:264 
  at Sentry.Protocol.Envelopes.Envelope.FromEvent (Sentry.SentryEvent event, Sentry.Extensibility.IDiagnosticLogger logger, System.Collections.Generic.IReadOnlyCollection`1[T] attachments, Sentry.SessionUpdate sessionUpdate) [0x00036] in /sentry-unity/src/sentry-dotnet/src/Sentry/Envelopes/Envelope.cs:153 
UnityEngine.DebugLogHandler:LogFormat (UnityEngine.LogType,UnityEngine.Object,string,object[])
Sentry.Unity.Integrations.UnityLogHandlerIntegration:LogFormat (UnityEngine.LogType,UnityEngine.Object,string,object[]) (at /sentry-unity/src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs:80)
UnityEngine.Debug:LogError (object)
[...]

PR #937, opened 2022-08-17
fix: Removed exception from screenshot attachment

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/925

With the .NET SDK filtering empty streams from being added to events, we can now return an empty `MemoryStream` instead of throwing an exception.


vaind (approved):
Finally, it's possible. 

Maybe mention dotnet SDK update in the changelog too? Even though it's a pre-release version.

bruno-garcia (approved):

bitsandfoxes:
> Maybe mention dotnet SDK update in the changelog too? Even though it's a pre-release version.

We're going to bump the .NET SDK this week so the changelog is going to be all nice and tidy with the next release.

PR #939, opened 2022-08-18
fix: Screenshot quality

bitsandfoxes:
Currently, the screenshot quality was relative to the current screen size which creates issues when a project supports a wide range of different resolutions, making "half the current resolution" on small screens barely usable.

Instead, we can map `High, Medium, Low` to "ranges" of resolutions akin to what Youtube videos offer.
```
ScreenshotQuality.Full => // Don't touch
ScreenshotQuality.High => 1920,     // 1080p
ScreenshotQuality.Medium => 1280,   // 720p
ScreenshotQuality.Low => 854,       // 480p
```
We take the wider side (portrait or landscape) and scale the screenshot to match while preserving the aspect ratio. (defacto @vaind's original implementation) with an user-facing enum instead of a `vec2 max resolution`.


vaind (approved):
Glad to see this :)

PR #944, opened 2022-08-19
ci: dangerfile improvements

vaind:
I've made some changes to dangerJS integration when copying to Unreal.

e.g. `#skip-changelog` wouldn't be necessary here because the PR starts with "ci:"


bitsandfoxes (approved):
Supercool! 🤯

PR #964, opened 2022-09-07
feat: Package update notification

bitsandfoxes:
With something like this, we can detect updates and also from which version. So we can notify about things like experimental features no longer being experimental on the wizard.

`Package changed from io.sentry.unity.dev-0.0.1 to io.sentry.unity.dev-0.0.2`


bitsandfoxes:
Closing this as `stale`. We're now also publishing to the new UPM PackageManager, which aims to make this obsolete.

PR #978, opened 2022-09-16
Add transaction processor

### LINKED ISSUE #439: Transactions have the wrong SDK name
Transactions send from the Unity SDK are reported as .NET's:

image

SimonCropp:
fixes #439


src/Sentry.Unity/UnityEventProcessor.cs
```diff
@@ -6,66 +6,80 @@
...
-
-            @event.ServerName = null;
-
-            return @event;
         }
 
         private void PopulateDevice(Device device)
         {
-            if (_mainThreadData.IsMainThread())
+            if (!_mainThreadData.IsMainThread())
```
bruno-garcia, line 56: Just to confirm: This is just a refactor? Or am I missing a change here I can't spot?

bruno-garcia (approved):
Thanks Simon!

src/Sentry.Unity/UnityEventProcessor.cs
SimonCropp: I inverted so as to have an early exit and make it easier to reason about

PR #989, opened 2022-09-28
fix: Clang formatter

bitsandfoxes:
#skip-changelog

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -13,7 +13,9 @@ jobs:
           submodules: recursive
 
       - name: Install clang-format
-        run: brew install clang-format
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/format-code.yml
```diff
@@ -13,7 +13,9 @@ jobs:
           submodules: recursive
 
       - name: Install clang-format
-        run: brew install clang-format
+        run: |
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+          brew install clang-format
```
vaind, line 18: just remove the step altogether? clang-format is already available https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#language-and-runtime

vaind (approved):

PR #991, opened 2022-09-28
feat: Runtime Init instrumentation 

bitsandfoxes:
Span & Transaction names need some discussion.

The goal here is to capture runtime initialization via spans i.e.
Screenshot 2022-09-30 at 14 21 22


vaind:
Just from a quick look at the picture first, so that I don't forget: maybe drop the "unity" prefix? Seems unnecessary

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Features
 
+- Automated Performance Instrumentalisation for Runtime Initialization ([#991](https://github.com/getsentry/sentry-unity/pull/991))
```
vaind, line 11: instrumentation maybe?

vaind (approved):
LGTM.

What's wrong with WebGL? Maybe add a comment to the class definition why it doesn't work there?

bitsandfoxes:
> LGTM.
> 
> What's wrong with WebGL? Maybe add a comment to the class definition why it doesn't work there?

Tracking the WebGL limitation here https://github.com/getsentry/sentry-unity/issues/1000
It seems to be an issue with the `UnityWebRequestTransport` that errors when trying to send events when the SDK is initialized during `SubsystemRegistration`.

PR #1002, opened 2022-10-05
fix: Consume Cocoa SDK as `.xcframework`

bitsandfoxes:
Currently, we build the Cocoa SDK and then copy the device and simulator `.framework` directories into the Plugins/iOS folder. They get excluded from Unity through the `.meta` files because, for the generated XCode project, we set the SDK up ourselves.

On the M1 the Carthage build artifacts are named slightly different which breaks our hardcoded [paths](https://github.com/getsentry/sentry-unity/blob/4439798181a535f3fc2b4f9264eee903bef4b217/Directory.Build.targets#L167) within the `.xcframework`.

Since we're currently setting the `.frameworks` up to be ignored anyway and we manually copy them to the generated Xcode project, the idea here is:
- [x] Copy whole the generated `.xcframework` instead of the individual `.framework` into the plugins folder
- [x] Append `~` to it so Unity ignores it (and does not copy the frameworks from within during XCode generation)
- [x] Set up the missing linking to the framework within the Xcode project
- [x] Update package snapshot test


Directory.Build.targets
```diff
@@ -156,25 +154,27 @@ Expected to exist:
...
           And '$(MSBuildProjectName)' == 'Sentry.Unity'
-          And (!Exists('$(SentryiOSDeviceArtifactsDestination)') OR !Exists('$(SentryiOSSimulatorArtifactsDestination)'))" BeforeTargets="BeforeBuild">
+          And (!Exists('$(SentryiOSArtifactsDestination)'))" BeforeTargets="BeforeBuild">
     <Error Condition="!Exists('$(SentryCocoaRoot)')" Text="Couldn't find the Cocoa root at $(SentryCocoaRoot)."></Error>
     <Message Importance="High" Text="Building Sentry iOS SDK."></Message>
 
-    <Exec WorkingDirectory="$(SentryCocoaRoot)" Command="carthage build --use-xcframeworks --no-skip-current --platform iOS"></Exec>
+    <!--    Workaround to enable building the Cocoa SDK with Bitcode enabled -->
+    <Message Importance="High" Text="Creating xcconfig file to enable Bitcode." />
+    <WriteLinesToFile File="$(SentryCocoaRoot)unity.xcconfig" Lines="ENABLE_BITCODE = YES" Overwrite="true" />
```
bitsandfoxes, line 163: The Cocoa builds with Bitcode `disabled` by default. Unity Xcode projects have Bitcode `enabled` by default. Bitcode has been deprecated but while we wait for Unity to update the generation we build the Cocoa SDK with a temporary config file to enable it.

test/Scripts.Tests/test-pack-contents.ps1
```diff
@@ -26,7 +26,11 @@ Can't compare package contents against baseline."
 $zip = [IO.Compression.ZipFile]::OpenRead($packageFile)
 try {
     $snapshotContent = $zip.Entries.FullName.Replace("\", "/")
+    # Override the generated GUID as part of the .bcsymbolmap
+    $snapshotContent = $snapshotContent -replace "BCSymbolMaps\/.+?(\.)","BCSymbolMaps/*."
```
bitsandfoxes, line 30: Filtering here, since Bitcode is now part of the `.xcframework` and the Guid changes with every build.

.github/workflows/sdk.yml
```diff
@@ -25,17 +25,17 @@ jobs:
...
+      #   uses: actions/cache@v2
+      #   id: cache
+      #   with:
+      #     # Note: native SDKs are cached and only built if the respective 'package-dev/Plugins/' directories are empty.
+      #     # Output changes only depending on the git sha of the submodules
+      #     # hash of package/package.json for cache busting on release builds (version bump)
+      #     path: |
+      #       package-dev/Plugins
+      #       modules/sentry-java/sentry-android-ndk/build/intermediates/merged_native_libs/release/out/lib
+      #     key: sdk=${{ inputs.target }}-${{ hashFiles('submodules-status', 'package/package.json', 'Directory.Build.targets', 'sdk-static/**') }}
```
bitsandfoxes, line 38:
```suggestion
      - name: Restore from cache
        uses: actions/cache@v2
        id: cache
        with:
          # Note: native SDKs are cached and only built if the respective 'package-dev/Plugins/' directories are empty.
          # Output changes only depending on the git sha of the submodules
          # hash of package/package.json for cache busting on release builds (version bump)
          path: |
            package-dev/Plugins
            modules/sentry-java/sentry-android-ndk/build/intermediates/merged_native_libs/release/out/lib
          key: sdk=${{ inputs.target }}-${{ hashFiles('submodules-status', 'package/package.json', 'Directory.Build.targets', 'sdk-static/**') }}
```

Directory.Build.targets
vaind, line 163: what happens if we have it enabled here and users want to disable it (in Unity) - would everything work normally? Just checking because nothing coming from xcode would surprise me

vaind (approved):
LGTM

Directory.Build.targets
bitsandfoxes: Yes. I did test that and Xcode just rolls with it.

PR #1024, opened 2022-10-18
ci: update deprecated actions

vaind:
outstanding issue: `setup-unity` action needs to be updated to node16 & latest `@actions/core` - I've tried doing it quickly in [a fork](https://github.com/vaind/setup-unity/tree/chore/update-deps) but windows build didn't work so let's wait for the upstream maintainer to do it


bitsandfoxes (approved):
<3

Issue #1029, opened 2022-10-20
Have a way to set log type that can be captured as a breadcrumb


eomene:
What problem could Sentry solve that it doesn't?
Right now I see all logs, normal logs, warning and logerrors are captured by sentry as breadcrumbs. But in my project there are alot of logwarnings that I just want to be able to ignore. 
I know we can technically do a filter. But there are alot. I would rather have a setting where I tell sentry not to automatically send breadcrumbs. Because I want to do it myself. 
Or set it to ignore warnings, logs and so on. 

It would be very handy...
Or maybe this is already possible. Would be nice if you can tell me how or send me in the right direction.

bitsandfoxes:
Hey @eomene. 
You're basically asking for a way to disable the automatic breadcrumb creation from logs? Or to capture only one specific type of log as breadcrumb?

We do have plans for making breadcrumbs configurable based on levels here:https://github.com/getsentry/sentry-unity/issues/60. Would that work for you?

bitsandfoxes:
While having granular control over breadcrumbs is still on the to-do, an opt-out absolutely makes sense to me.

PR #1030, opened 2022-10-20
feat: Added MinimumBreadcrumbLevel

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/1029

Because there is no inherent order to the log type and we're already transforming `Unity LogType` to `Sentry LogType` I added individual checkboxes to opt-out of adding breadcrumbs automatically.
Screenshot 2022-10-24 at 16 16 00


vaind (changes requested):
Not sure it makes sense to do the partial solution. Would it be much more trouble to implement the real solution as described in #60 - that's a superset of this

samples/unity-of-bugs/Assets/Resources/Sentry/SentryOptions.asset
```diff
@@ -25,6 +25,7 @@ MonoBehaviour:
   <AttachScreenshot>k__BackingField: 1
   <ScreenshotQuality>k__BackingField: 2
   <ScreenshotCompression>k__BackingField: 75
+  <AddLogsAsBreadcrumbs>k__BackingField: 1
```
vaind, line 28: Doesn't seem a correct name? Some older version of the file? Or where does this come from?

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -149,6 +152,32 @@ private void OnQuitting()
             _hub?.FlushAsync(_sentryOptions?.ShutdownTimeout ?? TimeSpan.FromSeconds(1)).GetAwaiter().GetResult();
         }
 
+
+        internal static bool PassesMinimumBreadcrumbLevel(LogType minimumLogType, LogType logType)
```
vaind, line 156: Maybe also document the chosen older in the comment to make it easier to read? What is the order you've chosen anyway? From my naive look just from the names I'd expect the following order:

```
    Log,
    Warning,
    Exception,
    Assert,
    Error,
```

or maybe `Exception` & `Assert` swapped, but they're actually orthogonal.

Also, `Error` isn't really defined in Unity docs so it's not clear to me whether it's any worse than `Exception` or not.

vaind: On a second thought... maybe we should give user a set of checkboxes to pick which they want to handle? So we don't have to come up with arbitrary ordering that's always going to make someone unhappy.

bitsandfoxes: Actually, this is a very fair point.

bitsandfoxes:
> Looks great. Sorry for being such a nuisance on the previous two versions

You were absolutely right tho. It's much more usable this way.

Issue #914, opened 2022-07-27
Add missing screenshot tests


bitsandfoxes:
Since we've got the quality controls back in.

PR #1047, opened 2022-10-25
fix: Added missing screenshot tests

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/914
Note: We already discussed if we should not have an API to query the `IsMainThread` the SDK uses internally. Maybe this might be reused?

#skip-changelog

### DIFF
--- a/src/Sentry.Unity/ScreenshotAttachment.cs
+++ b/src/Sentry.Unity/ScreenshotAttachment.cs
@@ -27,7 +27,7 @@ namespace Sentry.Unity
         {
             // Note: we need to check explicitly that we're on the same thread. While Unity would throw otherwise
             // when capturing the screenshot, it would only do so on development builds. On release, it just crashes...
-            if (_behaviour.IsMainThread is null || !_behaviour.IsMainThread())
+            if (_behaviour.MainThreadData.IsMainThread())
             {
                 _options.DiagnosticLogger?.LogDebug("Can't capture screenshots on other than main (UI) thread.");
                 return Stream.Null;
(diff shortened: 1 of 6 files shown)


src/Sentry.Unity/ScreenshotAttachment.cs
```diff
@@ -27,16 +27,16 @@ public Stream GetStream()
         {
             // Note: we need to check explicitly that we're on the same thread. While Unity would throw otherwise
             // when capturing the screenshot, it would only do so on development builds. On release, it just crashes...
-            if (!_behaviour.MainThreadData.IsMainThread())
+            if (_behaviour.IsMainThread is null || !_behaviour.IsMainThread())
```
vaind, line 30: maybe 
```suggestion
            if (_behaviour.IsMainThread?.Invoke() is not true)
```

src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -118,6 +118,7 @@ internal void OnApplicationFocus(bool hasFocus)
     internal partial class SentryMonoBehaviour
     {
         internal readonly MainThreadData MainThreadData = new();
+        internal Func<bool>? IsMainThread;
```
vaind, line 121: Maybe we should instead change `MainThreadData` to an interface so it's mockable for tests, instead of pulling this one function?

test/Sentry.Unity.Tests/ScreenshotAttachmentContentTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            var stream = sut.GetStream();
+
+            Assert.IsNotNull(stream);
+        }
+
+        [Test]
+        public void GetStream_IsNonMainThread_ReturnsNullStream()
+        {
+            _fixture.IsMainThread = false;
```
vaind, line 55: And just in case you didn't want to do the above changes & fixture - you could just run this test in a different thread?

bitsandfoxes: Wow, that was a real case of tunnel vision right there.

PR #1048, opened 2022-10-25
ci: Add action to generate stackaid.json

wescarr:
This action will generate a stackaid.json file, listing the dependencies that Sentry wants to fund on StackAid. FYI, I'm one of the @stackaid founders helping @chadwhitacre setup Sentry's funding.

### DIFF
--- a/.github/workflows/stackaid.yml
+++ b/.github/workflows/stackaid.yml
@@ -1,0 +1,11 @@
+name: 'fund-on-stackaid'
+on:
+  push:
+    branches:
+      - main
+jobs:
+  stackaid-json:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: stackaid/generate-stackaid-json@v1.9


.github/workflows/stackaid.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  push:
+    branches:
+      - main
+jobs:
+  stackaid-json:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: stackaid/generate-stackaid-json@v1.9
```
vaind, line 11: We pin actions in this repo. And checkout@v2 uses outdated Node, triggering notices.
```suggestion
      - uses: actions/checkout@v3
      - uses: stackaid/generate-stackaid-json@bbccf9adf6ab4acd69aa979b0cdc7e93a5e83d78 # pin@v1.9
```

.github/workflows/stackaid.yml
```diff
@@ -0,0 +1,11 @@
+name: 'fund-on-stackaid'
+on:
+  push:
+    branches:
+      - main
```
vaind, line 5: maybe update temporarily to trigger on this branch so we can see & check the recognized dependencies here?

chadwhitacre:
Closing. See https://github.com/getsentry/sentry/pull/40553#issuecomment-1292120935. 🙏

.github/workflows/stackaid.yml
wescarr: Thanks for the feedback @vaind We'll update our README as well.

PR #1055, opened 2022-10-28
fix: Updated expected sym count for iOS

bitsandfoxes:
#skip-changelog


vaind:
I've just hijacked the dotnet update branch for this

PR #1066, opened 2022-11-03
fix: Awake instrumentation

bitsandfoxes:
The issue:
Previously, I hooked up to the compilation finished callback to modify the `Assembly-CSharp.dll`, which worked fine on a minimal project. When testing this on the dogfooding game that uses a lot more Unity packages, it failed to load precompiled assemblies (such as `UnityEngine.TextRenderingModule.dll`) that get shipped with the engine installation and copied to different cache directories during the build.

The "solution":
The new approach is to implement the `IPostBuildPlayerScriptDLLs` interface. The report contains a buildfile array with all the used assemblies that got copied into `Temp/StagingArea/Managed/Data/` to be further processed. 
Because `Mono.Cecil` still fails to resolve those precompiled assemblies, I added the `SentryAssemblyResolver` that picks those files up from the working directory.


src/Sentry.Unity.Editor/AutoInstrumentation/SentryAssemblyResolver.cs
```diff
@@ -0,0 +1,40 @@
...
+    public class SentryAssemblyResolver : BaseAssemblyResolver
+    {
+        private readonly string _workingDirectory;
+        private readonly DefaultAssemblyResolver _defaultResolver;
+
+        public SentryAssemblyResolver(string workingDirectory)
+        {
+            _workingDirectory = workingDirectory;
+            _defaultResolver = new DefaultAssemblyResolver();
+        }
```
mattjohnsonpint, line 15: Nothing wrong here, but this is a good place to use a field initializer with targeted-type-new.

```suggestion
        private readonly string _workingDirectory;
        private readonly DefaultAssemblyResolver _defaultResolver = new();

        public SentryAssemblyResolver(string workingDirectory)
        {
            _workingDirectory = workingDirectory;
        }
```

src/Sentry.Unity.Editor/AutoInstrumentation/SentryAssemblyResolver.cs
```diff
@@ -0,0 +1,40 @@
...
+        public override AssemblyDefinition Resolve(AssemblyNameReference name)
+        {
+            AssemblyDefinition assembly;
+            try
+            {
+                assembly = _defaultResolver.Resolve(name);
+            }
+            catch (AssemblyResolutionException e)
+            {
+                var path = Path.GetFullPath(Path.Combine(_workingDirectory, e.AssemblyReference.Name + ".dll"));
```
mattjohnsonpint, line 26: Is `Path.GetFullPath` necessary?  I think both `File.Exists` and `AssemblyDefinition.ReadAssembly` would both resolve relative paths the same way.

src/Sentry.Unity.Editor/AutoInstrumentation/SentryAssemblyResolver.cs
```diff
@@ -0,0 +1,40 @@
...
+            _workingDirectory = workingDirectory;
+            _defaultResolver = new DefaultAssemblyResolver();
+        }
+
+        public override AssemblyDefinition Resolve(AssemblyNameReference name)
+        {
+            AssemblyDefinition assembly;
+            try
+            {
+                assembly = _defaultResolver.Resolve(name);
```
mattjohnsonpint, line 22: Would be cleaner as:

```suggestion
                return _defaultResolver.Resolve(name);
```

Same below in the catch.  Then you don't have to declare the variable at all, and it also turns `else throw` into just `throw`.

src/Sentry.Unity.Editor/AutoInstrumentation/ModuleReadWriter.cs
```diff
@@ -46,15 +132,65 @@ private static bool TryReadSymbols(this ModuleDefinition module)
...
+                        case null:
+                            continue;
+                    }
+
+                    if (method.Parameters.Count != requiredParameters.Length)
+                    {
+                        continue;
+                    }
+
+                    // We go over all the required parameters and compare them to the parameters found in the method.
```
mattjohnsonpint, line 165: Is this logic correct? It means that if each of the required parameters is found *anywhere* in the method signature, then it would match.

For example, `Foo(string, string, int, int)` would also match `Foo(int, int, string, string)`

src/Sentry.Unity.Editor/AutoInstrumentation/ModuleReadWriter.cs
```diff
@@ -46,15 +132,65 @@ private static bool TryReadSymbols(this ModuleDefinition module)
...
+                    }
+                }
+            }
 
-            module.Write(file, parameters);
+            throw new Exception(
+                $"Failed to find method '{name}' " +
+                $"in '{typeDefinition.FullName}' " +
+                $"with parameters: '{(requiredParameters is not null ? string.Join(",", requiredParameters.AsEnumerable()) : "none")}'");
         }
```
mattjohnsonpint, line 194: Here's a cleaner version of the above method.

```csharp
internal MethodDefinition GetMethodDefinition(TypeDefinition typeDefinition, string name, Type[]? requiredParameters = null)
{
    requiredParameters ??= Array.Empty<Type>();

    var matchingMethod = typeDefinition.Methods
        .FirstOrDefault(method =>
            method.Name == name &&
            requiredParameters.Length == method.Parameters.Count &&
            
            // This is the equivalent to the current code
            // requiredParameters.All(parameter =>
            //     method.Parameters.Any(p => p.ParameterType.FullName == parameter.FullName))

            // I think you want this instead, as it will match in position
            method.Parameters.Select(p => p.ParameterType.FullName)
                .SequenceEqual(requiredParameters.Select(x => x.FullName))
        );

    if (matchingMethod == null)
    {
        throw new Exception(
            $"Failed to find method '{name}' " +
            $"in '{typeDefinition.FullName}' " +
            $"with parameters: '{string.Join<Type>(",", requiredParameters)}'");
    }

    return matchingMethod;
}
```

src/Sentry.Unity.Editor/AutoInstrumentation/ModuleReadWriter.cs
```diff
@@ -1,39 +1,125 @@
...
+            var moduleReaderWriter = new SentryPlayerReaderWriter(workingDirectory, playerAssemblyPath, sentryUnityAssemblyPath);
+            moduleReaderWriter.ReadFromWorkingDirectory();
+
+            return moduleReaderWriter;
+        }
+
+        public void Write()
+        {
+            var parameters = new WriterParameters { WriteSymbols = _playerModuleHasSymbols };
+            _playerModule?.Write(_playerAssemblyPath, parameters);
```
mattjohnsonpint, line 62:
```suggestion
            _playerModule.Write(_playerAssemblyPath, parameters);
```

src/Sentry.Unity.Editor/AutoInstrumentation/ModuleReadWriter.cs
```diff
@@ -1,39 +1,125 @@
...
+using System.IO;
+using System.Linq;
 using Mono.Cecil;
-using Mono.Cecil.Cil;
-using UnityEngine;
 
 namespace Sentry.Unity.Editor
 {
-    static class ModuleReaderWriter
+    internal class SentryPlayerReaderWriter
```
mattjohnsonpint, line 9: nit: Rename file to match.

src/Sentry.Unity.Editor/AutoInstrumentation/ModuleReadWriter.cs
```diff
@@ -1,39 +1,125 @@
...
+        private readonly string _playerAssemblyPath;
+        private readonly string _sentryUnityAssemblyPath;
+
+        private bool _playerModuleHasSymbols;
+        private ModuleDefinition _playerModule = null!;         // Set when reading the assemblies
+        private ModuleDefinition _sentryUnityModule = null!;    // Set when reading the assemblies
+
+        private readonly SentryAssemblyResolver _assemblyResolver;
+
+        internal SentryPlayerReaderWriter(
```
mattjohnsonpint, line 23: This constructor and several methods are marked `internal` but seem like they could be `private` - or were you planning to use them elsewhere?

src/Sentry.Unity.Editor/AutoInstrumentation/ModuleReadWriter.cs
mattjohnsonpint: See my suggested change below also.  It's a bit easier to see in LINQ.

src/Sentry.Unity.Editor/AutoInstrumentation/SentryAssemblyResolver.cs
bitsandfoxes: Apparently not. I did a double-check and I get around the whole `SentryResolver` by adding to the search directories of a `DefaultAssemblyResolver`! 👍 
Thanks for pointing that out.

src/Sentry.Unity.Editor/AutoInstrumentation/SentryPerformanceAutoInstrumentation.cs
```diff
@@ -0,0 +1,113 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using Mono.Cecil.Cil;
+using Sentry.Extensibility;
+using UnityEditor.Build;
+using UnityEditor.Build.Reporting;
+using UnityEngine;
+using Debug = UnityEngine.Debug;
```
mattjohnsonpint, line 10:
```suggestion
```

src/Sentry.Unity.Editor/AutoInstrumentation/SentryPlayerReaderWriter.cs
```diff
@@ -0,0 +1,170 @@
...
+                module.ReadSymbols();
+                return true;
+            }
+            catch
+            {
+                return false;
+            }
+        }
+
+        private TypeDefinition GetTypeDefinition(ModuleDefinition module, Type type)
```
mattjohnsonpint, line 136:
```suggestion
        private static TypeDefinition GetTypeDefinition(ModuleDefinition module, Type type)
```

src/Sentry.Unity.Editor/AutoInstrumentation/SentryPlayerReaderWriter.cs
```diff
@@ -0,0 +1,170 @@
...
+
+                return (module, hasSymbols);
+            }
+            catch (Exception exception)
+            {
+                throw new ArgumentException($"Failed to read: '{file}'", exception);
+            }
+        }
+
+        private bool TryReadSymbols(ModuleDefinition module)
```
mattjohnsonpint, line 123:
```suggestion
        private static bool TryReadSymbols(ModuleDefinition module)
```

src/Sentry.Unity.Editor/AutoInstrumentation/SentryPlayerReaderWriter.cs
```diff
@@ -0,0 +1,170 @@
...
+            var typeDefinition = module.GetType(type.FullName);
+            if (typeDefinition is null)
+            {
+                throw new ArgumentException($"Failed to get requested type definition in {module.Name}", type.FullName);
+            }
+
+            return typeDefinition;
+        }
+
+        private MethodDefinition GetMethodDefinition(TypeDefinition typeDefinition, string name, Type[]? requiredParameters = null)
```
mattjohnsonpint, line 147:
```suggestion
        private static MethodDefinition GetMethodDefinition(TypeDefinition typeDefinition, string name, Type[]? requiredParameters = null)
```

mattjohnsonpint (approved):
A few minor suggestions, but looks great!

PR #1077, opened 2022-11-10
ci: don't run all the integration tests on the release branch

vaind:
It's all tested already when a release is launched. The integration tests are flaky no matter what we do... This regularly breaks releases. 

Assuming we don't run release when the main branch is not green, it shouldn't be necessary to run integration tests again on the release branch - thus we can make a release faster and without failing all the time.


bitsandfoxes (approved):
Good point.

PR #1091, opened 2022-11-24
fix: Symbol upload logging for iOS

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/1071
- Forcing the symbol upload in the foreground to have the logs available in the Xcode console.
- Removing the log level from the command because we write it into the `.properties` already
- Removed `note:` because it was messing with the order of logs

### DIFF
--- a/src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
+++ b/src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
@@ -22,14 +22,8 @@ namespace Sentry.Unity.Editor.iOS
         private readonly string _mainPath = Path.Combine("MainApp", "main.mm");
         private readonly string _optionsPath = Path.Combine("MainApp", OptionsName);
         private readonly string _uploadScript = @"
-
 process_upload_symbols() {{
-    ./{0} --log-level=debug upload-dif {1} --force-foreground $BUILT_PRODUCTS_DIR &> sentry-symbols-upload.log
-    if [ ""$?"" -eq ""0"" ] ; then
-        echo ""note: Debug symbol upload successful.""
-    else
-        echo ""error: Debug symbol upload failed. See './sentry-symbols-upload.log' for more information.""
-    fi
+    ./{0} upload-dif --force-foreground {1} $BUILT_PRODUCTS_DIR 2>&1 | tee ./sentry-symbols-upload.log
 }}
 
 export SENTRY_PROPERTIES=sentry.properties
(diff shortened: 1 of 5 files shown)


src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -25,7 +25,7 @@ internal class SentryXcodeProject : IDisposable
 
 process_upload_symbols() {{
     ./{0} --log-level=debug upload-dif {1} $BUILT_PRODUCTS_DIR &> sentry-symbols-upload.log &
-    if [ $? -eq 0] ; then
+    if [ $? -eq ""0"" ] ; then
```
vaind, line 28:
```suggestion
    if [ ""$?"" -eq ""0"" ] ; then
```

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -22,22 +22,21 @@ internal class SentryXcodeProject : IDisposable
         private readonly string _mainPath = Path.Combine("MainApp", "main.mm");
         private readonly string _optionsPath = Path.Combine("MainApp", OptionsName);
         private readonly string _uploadScript = @"
-
 process_upload_symbols() {{
-    ./{0} --log-level=info upload-dif {1} $BUILT_PRODUCTS_DIR 2>&1 | tee ./sentry-symbols-upload.log
+    ./{0} upload-dif --force-foreground {1} $BUILT_PRODUCTS_DIR 2>&1 | tee ./sentry-symbols-upload.log
```
vaind, line 26: why the removal of `--log-level=info`? is it defined elsewhere?

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
```diff
@@ -22,22 +22,21 @@ internal class SentryXcodeProject : IDisposable
...
 }}
 
 export SENTRY_PROPERTIES=sentry.properties
 if [ ""$ENABLE_BITCODE"" = ""NO"" ] ; then
-    echo ""note: Uploading debug symbols (Bitcode disabled).""
+    echo ""Uploading debug symbols (Bitcode disabled).""
```
vaind, line 31: I remember `note: ` prefix having to do something with how the logs were displayed in xcode. Why removing? I may be dreaming though...

vaind (approved):

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
bitsandfoxes: Yes. There are `note:`, `warning:` and `error:` afaik but we don't use them for anything and since sentry cli just plain logs it messes with the order.

src/Sentry.Unity.Editor.iOS/SentryXcodeProject.cs
bitsandfoxes: Yes. We write the `log-level` into `sentry.properties`. Just like on Android.

PR #1092, opened 2022-11-24
fix: Native layer not closing

bitsandfoxes:
On mobile, the native layer is getting set up during buildtime and is self-initializing.
If the SDK gets disabled during runtime (i.e. in the `Configure` callback, then the native layer needs to be closed down too.

Something like `SentryUnity.Close()` is still missing, which allows for the programmatic closing of all of Sentry.

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -108,6 +108,15 @@ namespace Sentry.Unity
                 }
 #endif
             }
+            else
+            {
+                // Closing down the native layer that are set up during build and self-initialize
+#if SENTRY_NATIVE_COCOA
+                    SentryNativeCocoa.Close(options.DiagnosticLogger);
+#elif SENTRY_NATIVE_ANDROID
+                    SentryNativeAndroid.Close(options.DiagnosticLogger);
+#endif
+            }
         }
     }
 
--- a/src/Sentry.Unity.Android/SentryNativeAndroid.cs
+++ b/src/Sentry.Unity.Android/SentryNativeAndroid.cs
@@ -53,15 +53,20 @@ namespace Sentry.Unity.Android
                         "Failed to reinstall backend. Captured native crashes will miss scope data and tag.", e);
                 }
 
-                ApplicationAdapter.Instance.Quitting += () =>
-                {
-                    // Sentry Native is initialized and closed by the Java SDK, no need to call into it directly
-                    options.DiagnosticLogger?.LogDebug("Closing the sentry-java SDK");
-                    SentryJava.Close();
-                };
+                ApplicationAdapter.Instance.Quitting += () => Close(options.DiagnosticLogger);
 
                 options.DefaultUserId = SentryJava.GetInstallationId();
             }
+        }
+
+        /// <summary>
+        /// Closes the native Android support.
+        /// </summary>
+        public static void Close(IDiagnosticLogger? logger = null)
+        {
+            // Sentry Native is initialized and closed by the Java SDK, no need to call into it directly
+            logger?.LogDebug("Closing the sentry-java SDK");
+            SentryJava.Close();
         }
     }
 }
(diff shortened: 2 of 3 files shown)


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -106,6 +106,15 @@ public static void Init()
                     options.DiagnosticLogger?.LogDebug("Creating '{0}' span.", SubSystemSpanOperation);
                     SubSystemRegistrationSpan = InitSpan.StartChild(SubSystemSpanOperation, "subsystem registration");
                 }
+#endif
+            }
+            else
+            {
+                // Closing down the native layer that are set up during build and self-initialize
+#if SENTRY_NATIVE_COCOA
+                SentryNativeCocoa.Close(options.DiagnosticLogger);
```
vaind, line 115: AFAICT [Cocoa native close](https://github.com/getsentry/sentry-cocoa/blob/fbeb49aa4bf63d1ce6537cbc4d509564d1510617/Sources/Sentry/SentrySDK.m#L405-L427) would crash here if sentry wasn't initialized. (Or maybe I understand objective-c wrong and it would first init the hub before closing it)

I think we cannot know here whether it was initialized - we could check options but that doesn't mean these are the same as the ones during build time. 

Therefore, we should check with the Cocoa SDK team & maybe update close() first to check whether hub is nil.

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -13,6 +15,9 @@ namespace Sentry.Unity
     /// </summary>
     public static class SentryUnity
```
vaind, line 16: Should this be `IDisposable` too?

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -33,75 +38,89 @@ public static void Init(Action<SentryUnityOptions> sentryUnityOptionsConfigure)
...
-                            options.DiagnosticLogger?.Log(SentryLevel.Warning,
-                                "Exception while releasing the lockfile on the config directory.", ex);
-                        }
-                    }
-                };
+                    Options.NativeSupportCloseCallback?.Invoke();
+                    Options.NativeSupportCloseCallback = null;
+                }
+
+                DotnetSdk?.Dispose();
```
vaind, line 110: clear instance afterwards so that another Close() wouldn't do anything.

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -33,75 +38,89 @@ public static void Init(Action<SentryUnityOptions> sentryUnityOptionsConfigure)
...
+            }
+            finally
```
vaind, line 112: Do we expect this to throw? If so, maybe catch & log, same as below for LockFile, instead of having a `finally` branch?

vaind (changes requested):

src/Sentry.Unity/SentryUnity.cs
bitsandfoxes: I'm not sure how that would work right now or what we would dispose. It can't be the `SentryUnity` class because it'd have to no longer be `static`. So the `Init` method should return some intermediate `IDisposable`.

vaind (approved):
It's OK, though maybe we could clean `SentryUnity` class a bit by extracting a non-static class with all the current code and have a single variable `private static SentryUnitySdk? Instance` that we set in init() and remove in close() (+ call its close of course).
`SentryUnitySdk` could have non-static and non-nullable Options, SentrySdk and FileStream lockfile` 
WDYT?

bitsandfoxes:
> It's OK, though maybe we could clean `SentryUnity` class a bit by extracting a non-static class with all the current code and have a single variable `private static SentryUnitySdk? Instance` that we set in init() and remove in close() (+ call its close() of course). `SentryUnitySdk` could have non-static and non-nullable Options, SentrySdk and FileStream lockfile` WDYT?

Excellent point.

Issue #1112, opened 2022-12-13
The referenced script on this Behaviour (Game Object '<null>') is missing!


ArnauKokoro:
### Environment
Sentry SaaS (sentry.io)

Which version of the SDK?
0.27.0

How did you install the package?
Git-URL

Which version of Unity?
2020.3.15f2

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Android, iOS

### Steps to Reproduce

1. Build Android & iOS

### Expected Result

No Warning Logs appears

### Actual Result
```
The referenced script (Unknown) on this Behaviour is missing!
    UnityEngine.StackTraceUtility:ExtractStackTrace () (at /Users/bokken/buildslave/unity/build/Runtime/Export/Scripting/StackTrace.cs:37)
    UnityEngine.ResourcesAPI:Load (string,System.Type) (at /Users/bokken/buildslave/unity/build/Runtime/Export/Resources/Resources.bindings.cs:66)
    UnityEngine.Resources:Load (string,System.Type) (at /Users/bokken/buildslave/unity/build/Runtime/Export/Resources/Resources.bindings.cs:117)
    UnityEngine.Resources:Load<Sentry.Unity.ScriptableSentryUnityOptions> (string) (at /Users/bokken/buildslave/unity/build/Runtime/Export/Resources/Resources.bindings.cs:112)
    Sentry.Unity.ScriptableSentryUnityOptions:LoadSentryUnityOptions (Sentry.Unity.ISentryUnityInfo) (at /sentry-unity/src/Sentry.Unity/ScriptableSentryUnityOptions.cs:96)
    Sentry.Unity.SentryInitialization:Init () (at /Library/PackageCache/io.sentry.unity@b562146d73/Runtime/SentryInitialization.cs:59)

The referenced script on this Behaviour (Game Object '<null>') is missing!
    UnityEngine.StackTraceUtility:ExtractStackTrace () (at /Users/bokken/buildslave/unity/build/Runtime/Export/Scripting/StackTrace.cs:37)
    UnityEngine.ResourcesAPI:Load (string,System.Type) (at /Users/bokken/buildslave/unity/build/Runtime/Export/Resources/Resources.bindings.cs:66)
    UnityEngine.Resources:Load (string,System.Type) (at /Users/bokken/buildslave/unity/build/Runtime/Export/Resources/Resources.bindings.cs:117)
[...]

PR #1128, opened 2022-12-22
fix: Warning when loading the options object

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/1112

We have two OptionConfigurations and by default:
- One lives in `Scripts/RuntimeOptionsConfig.cs`
- The other in `Editor/BuildtimeOptionsConfig.cs`

During runtime, when the `SentryScriptableOptions` gets loaded and it has a reference to the buildtime scriptable config it cannot resolve the class (since it lives inside of `Editor`) which leads to warnings:
```
The referenced script (Unknown) on this Behaviour is missing!
The referenced script on this Behaviour (Game Object '<null>') is missing!
```

To make life easier I moved every scriptable object we have to `Sentry.Unity` and the options and options configs now all live inside of `Resource/Sentry/`.
There were also a couple of cases where we used the `is` operator which is not recommended for all Unity objects.


src/Sentry.Unity.Editor/ConfigurationWindow/OptionsConfigurationTab.cs
```diff
@@ -71,21 +69,18 @@ internal static class OptionsConfigurationItem
...
-        private static string SentryAssetPath(string scriptName, bool isEditorScript) =>
-            string.Format("Assets/{0}/Sentry/{1}.asset", isEditorScript ? "Plugins" : "Resources", scriptName);
+        private static string SentryAssetPath(string scriptName) => $"Assets/Resources/Sentry/{scriptName}.asset";
 
         private static void CreateScript<T>(string fieldName, string scriptName)
         {
-            var isEditorScript = typeof(T) == typeof(BuildtimeConfiguration);
-            var directory = isEditorScript ? "Assets/Editor" : "Assets/Scripts";
-
+            const string directory = "Assets/Scripts";
```
vaind, line 76: Have you tested this with the integration test? The build-time script had buiold-time Unity (not Sentry) dependencies, so it had to reside in `Assets/Editor`

bitsandfoxes: Yes. That's why I moved the `CliOptions` to `Sentry.Unity` to get rid of the `Sentry.Unity.Editor` dependency in the options. Effectively, this means that the Buildtime scriptable object now lives inside of `Resources` and is part of the build but we got rid of all the blind casting and weird namespacing/dependencies.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- The SDK no longer creates warnings when loading the options ([#1128](https://github.com/getsentry/sentry-unity/pull/1128))
```
vaind, line 7: I think this should get a more prominent entry - it is a breaking change and users need to actively migrate their runtime options configs.

samples/unity-of-bugs/Assets/Resources/Sentry/SentryOptions.asset
```diff
@@ -47,7 +47,7 @@ MonoBehaviour:
   <MacosNativeSupportEnabled>k__BackingField: 1
   <LinuxNativeSupportEnabled>k__BackingField: 1
   <Il2CppLineNumberSupportEnabled>k__BackingField: 1
-  <OptionsConfiguration>k__BackingField: {fileID: 11400000, guid: 407b4d5f2fef2c845bf2a3dcdf6b00fb,
+  <RuntimeOptionsConfiguration>k__BackingField: {fileID: 11400000, guid: 407b4d5f2fef2c845bf2a3dcdf6b00fb,
```
vaind, line 50: I'm not entirely happy about the rename of the option variable because that means when users upgrade, they'll silently lose their existing configurations (they won't be run anymore), doesn't it?

bitsandfoxes: Reverted the renaming.

CHANGELOG.md
bitsandfoxes: I thought so too and wondered if that should happen during release. But you're right. I added the fix as breaking change "changelog.. type?" (which I took from the Cocoa SDK)

vaind (approved):

Issue #1135, opened 2023-01-04
UnityLogHandlerIntegration ignored message formatting


YLahin:
### Environment

How do you use Sentry?
self-hosted (22.11.0)

Which version of the SDK?
0.27.0

How did you install the package?
Git-URL

Which version of Unity?
2021.3.12f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
All platforms

### Steps to Reproduce

1. Run game
2. Invoke Debug.unityLogger.LogError("Test", "Hello World")
3. Open sentry dashboard

### Expected Result

Message "Test: Hello World"

### Actual Result

Message "Test"

### Any logs or screenshots
<img width="453" alt="000" src="https://user-images.githubusercontent.com/3340145/210533299-7d2ead0c-6fed-4cb2-bc4f-94ad82cfe053.png">
<img width="493" alt="002" src="https://user-images.githubusercontent.com/3340145/210533050-16876c11-a3a1-4976-b239-565e0c0c463f.png">
<img width="785" alt="003" src="https://user-images.githubusercontent.com/3340145/210533063-5f7d702c-2d2a-499d-9c4a-b7c98516afbc.png">
<img width="299" alt="005" src="https://user-images.githubusercontent.com/3340145/210533360-ea0ea5a2-7043-42c6-ae7e-7f4176fcf337.png">
<img width="380" alt="007" src="https://user-images.githubusercontent.com/3340145/210535159-4ddd5367-1311-47e7-9084-550e34281815.png">

PR #1150, opened 2023-01-24
fix: Logging integration captures tags and messages

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/1135

`Debug.unityLogger.LogError("HUE", "Something exploded!");` would only capture `HUE`, but now it looks like this: 
image

Previously: We put `Sentry: ` as prefix to all logs coming from within the SDK. And then we checked all logs coming in through the LoggingIntegration for starting with that prefix.

The `Format` changes from `{0}` to `{0}: {1}` if there is a tag present. Which means if the SDK makes use of the tag instead of adding a prefix we save on checking every message.

And if we make use of the `unityLogger` in our logger we save on having to route everything through an interceptor for our tests. 

win-win!


src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -125,6 +113,7 @@ internal void CaptureLogFormat(LogType logType, UnityEngine.Object? context, str
                 }
             }
 
+            var logMessage = string.Format(format, args);
```
mattjohnsonpint, line 116: At minimum, you should still check that args length is > 0.  Otherwise you might possibly run into issues if the string being formatted has braces in it.  (similar to https://github.com/getsentry/sentry-dotnet/pull/2004)

vaind (approved):
changelog conflicts

PR #1165, opened 2023-02-01
feat: Setting the `instruction_addr_adjustment` for IL2CPP exceptions

bitsandfoxes:
Based on https://github.com/getsentry/rfcs/pull/43
Fixing C# line numbers for IL2CPP builds.


vaind (approved):

src/Sentry.Unity/Il2CppEventProcessor.cs
```diff
@@ -55,6 +51,11 @@ public void Process(Exception incomingException, SentryEvent sentryEvent)
                     continue;
                 }
 
+                sentryStacktrace.AddressAdjustment =
+                    Application.platform == RuntimePlatform.Android
+                        ? InstructionAddressAdjustment.None
+                        : InstructionAddressAdjustment.All;
```
vaind, line 57: Have you tested other platforms with @Swatinem? This comment only mentioned Android, macOS and iOS: https://github.com/getsentry/rfcs/pull/43#issuecomment-1387361867

Swatinem: I think we have also tried Windows as well. Android seemed to be the only platform where the instruction addresses are already adjusted coming from Unity.

vaind: Any info on Linux?

Swatinem: I don’t think we have ever looked at that directly. @bitsandfoxes can you prepare an example? Given that all the other platforms are consistent across Unity versions, I believe its fine to only have one example.

bitsandfoxes: Can confirm it works as intended: https://sentry-sdks.sentry.io/issues/3910273139/events/0e1f0e6af5ab4b5d8641018092872af5/?project=4504604988538880

PR #1188, opened 2023-02-21
fix: Removed `Overrride Sentry URL` from editor window

bitsandfoxes:
We didn't intend to put this on the editor window and it caused some confusion as to `what to put there`. Since it's a very specific edge case and it's accessible from the `BuildTimeOptionsConfig` I think it can be removed from the editor window.


vaind (approved):
other than the changelog, finally!

PR #1208, opened 2023-03-01
feat: Tests add breadcrumbs during setup

bitsandfoxes:
We're receiving a whole lot of `System.Net.WebException` that seem to originate from our tests. But the issue itself it not helpful (yet) 
image

#skip-changelog


bitsandfoxes:
Ridiculous idea. The SDK is disabled during `SetUp`. Where did I expect the breadcrumb to go?

PR #1209, opened 2023-03-01
fix: Updated expected `main.mm`

bitsandfoxes:
Slipped through during the Cocoa 8.1 update.
It is odd (read: concerning) tho. Why are the tests failing locally but not in CI?

#skip-changelog


bruno-garcia (commented):
> It is odd (read: concerning) tho. Why are the tests failing locally but not in CI?

Definitely concerning, worth looking at this before merging. Get it to fail first, then merge.

.github/workflows/ci.yml
```diff
@@ -159,7 +159,7 @@ jobs:
           docker exec unity dotnet msbuild /t:UnityPlayModeTest /p:Configuration=Release /p:OutDir=other test/Sentry.Unity.Tests
 
       - name: Upload test artifacts (playmode)
-        if: ${{ failure() }}
+        if: ${{ always() }}
```
bitsandfoxes, line 162:
```suggestion
        if: ${{ failure() }}
```

.github/workflows/ci.yml
```diff
@@ -169,7 +169,7 @@ jobs:
         run: docker exec unity dotnet msbuild /t:UnityEditModeTest /p:Configuration=Release /p:OutDir=other test/Sentry.Unity.Editor.Tests
 
       - name: Upload test artifacts (editmode)
-        if: ${{ failure() }}
+        if: ${{ always() }}
```
bitsandfoxes, line 172:
```suggestion
        if: ${{ failure() }}
```

Issue #1214, opened 2023-03-03
Make ANR detection timescale agnostic


bitsandfoxes:
The ANR detection right now uses a coroutine to detect if the main thread failed to update for `ANR-timeout` amount of time.
It uses `WaitForSeconds` to sleep but this is based on [`Time.timescale`](https://docs.unity3d.com/ScriptReference/Time-timeScale.html) and will therefore not update if the scale is set to 0, effectively giving false positives in those cases.

Instead of relying on `WaitForSeconds` we could calculate the sleep duration based on `Time.realtimeSinceStartup`.

mattjohnsonpint:
Or [`Time.unscaledTime`](https://docs.unity3d.com/ScriptReference/Time-unscaledTime.html)?

PR #1225, opened 2023-03-13
fix: Make ANR detection unaffected by `Time.timescale`

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/1214


vaind (approved):
wow, didn't even think this would exist 👍

PR #1226, opened 2023-03-14
fix: Uploading all `.pdb` at `Temp/ManagedSymbols` for Mono on Windows

bitsandfoxes:
Unity allows for the use of [Assembly Definitions](https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html) which leads to the creation of separate `.dll` and `.pdb` files. Currently, those get skipped due to the naming restriction.

### DIFF
--- a/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
+++ b/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
@@ -178,7 +178,7 @@ namespace Sentry.Unity.Editor.Native
 
             if (isMono)
             {
-                addFilesMatching($"{projectDir}/Temp", new[] { "**/UnityEngine.*.pdb", "**/Assembly-CSharp.pdb" });
+                addFilesMatching($"{projectDir}/Temp/ManagedSymbols", new[] { "*.pdb" });
             }
 
             var cliArgs = "upload-dif ";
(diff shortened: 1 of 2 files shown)


src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -178,7 +178,7 @@ private static void UploadDebugSymbols(IDiagnosticLogger logger, BuildTarget tar
 
             if (isMono)
             {
-                addFilesMatching($"{projectDir}/Temp", new[] { "**/UnityEngine.*.pdb", "**/Assembly-CSharp.pdb" });
```
vaind, line 181: Are you sure the original files are still matched **for all supported versions of Unity**?
I remember some of these rules (not saying it relates to this one specifically) being very different for each version I've tested this on.

Maybe instead just add the new rule instead of removing the old one unless you've verified we're not losing useful files.

bitsandfoxes: I double checked and at the very least for Windows this works.
Line numbers on OSX and mono don't work right now. And that directory does not exist either. Checking for Linux.

vaind (approved):

PR #1242, opened 2023-03-21
ci: Android build in a separate job

vaind:
* Enables android project export from Unity (for the integration test) 
* fixes issues and some bugs present with that option
* I've also got a cool idea to move android retry logic to a separate workflow so we can actually retry the whole thing on a different runner, since if it failed once, it was likely to fail again on the same machine.


bitsandfoxes (approved):
Pure 🔥

PR #1263, opened 2023-03-31
fix: Don't add `Sentry Android` with Android Native Support disabled

bitsandfoxes:
The base idea is to make unity games as a library work. For that, the SDK can't move any native SDK into the gradle output project.
That means
- The SDK no longer needs to set `self-init = false` in the manifest because there is no android SDK in the output project
- Manually copy/remove the Android SDK from the output project

### DIFF
--- a/src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
+++ b/src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
@@ -229,7 +229,7 @@ namespace Sentry.Unity.Editor.Android
 
         internal void CopyAndroidSdkToGradleProject(string gradlePath)
         {
-            var sentrySdkPath = Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins", "Android", "Sentry");
+            var sentrySdkPath = Path.Combine("Packages", SentryPackageInfo.GetName(), "Plugins", "Android", "Sentry~");
             var targetPath = Path.Combine(gradlePath, "unityLibrary", "libs");
 
             if (_options is { Enabled: true, AndroidNativeSupportEnabled: true })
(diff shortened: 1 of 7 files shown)


test/Scripts.Tests/package-release.zip.snapshot
```diff
@@ -202,13 +202,9 @@ Plugins/Linux/Sentry/libsentry.so
...
-Plugins/Android/Sentry.meta
-Plugins/Android/Sentry/sentry-android-core-release.aar
-Plugins/Android/Sentry/sentry-android-core-release.aar.meta
-Plugins/Android/Sentry/sentry-android-ndk-release.aar
-Plugins/Android/Sentry/sentry-android-ndk-release.aar.meta
-Plugins/Android/Sentry/sentry.jar
-Plugins/Android/Sentry/sentry.jar.meta
+Plugins/Android/Sentry~/sentry-android-core-release.aar
+Plugins/Android/Sentry~/sentry-android-ndk-release.aar
+Plugins/Android/Sentry~/sentry.jar
```
vaind, line 207: where are the .meta files?

src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
```diff
@@ -226,6 +234,93 @@ internal void SetupProguard(string gradleProjectPath)
...
+                    return;
+                }
+
+                throw new ArgumentException($"Failed to add Sentry Android dependencies to 'build.gradle'.\n{gradle}", nameof(gradle));
+            }
+            else
+            {
+                if (gradle.Contains(SDKDependencies))
+                {
+                    _logger.LogInfo("Adding Android SDK dependencies have previously been added. Removing them.");
```
vaind, line 308:
```suggestion
                    _logger.LogInfo("Android SDK dependencies have previously been added. Removing them.");
```

test/Scripts.Tests/package-release.zip.snapshot
bitsandfoxes: We don't need/have meta files for those anymore. Appending `~` to a directory makes Unity ignore it and we're copying and setting up the Android SDK ourselves.

vaind (approved):

PR #1292, opened 2023-04-25
chore: Added override to build locally with 2022

bitsandfoxes:
From testing this locally it worked fine until I tried building with 2022.

```
error CS1705: Assembly 'UnityEngine' with identity 'UnityEngine, uses 'netstandard, Version=2.1.0.0", 
which has a higher version than referenced assembly 'netstandard' 
with identity 'netstandard, Version=2.0.0.0' [/Users/bitfox/Workspace/sentry-unity/src/Sentry.Unity/Sentry.Unity.csproj]
```

#skip-changelog

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,7 +11,7 @@
     <SignAssembly>false</SignAssembly>
 
     <!-- This is the default for our lowest supported version 2019 LTS: netstandard2.0 -->
-    <!-- It gets overridden in the test/props -->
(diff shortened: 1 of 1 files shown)


vaind (approved):

mattjohnsonpint:
We'll want a .NET Standard 2.1 version of Sentry to be used as well.  Adding this in https://github.com/getsentry/sentry-dotnet/pull/2328.  Not sure if there's any additional work to ensure it's used for Unity 2022 or not.

mattjohnsonpint:
... probably not, since it's just a `ProjectReference`

PR #1294, opened 2023-04-25
fix: Filter `Web` and `SocketException`

bitsandfoxes:
We're seeing a huge number of barely actionable [events](https://sentry-sdks.sentry.io/issues/3952628308/events/d35a1d6557d741d994e46fc1de8dadd8/). They are being captured through the `UnobservedTaskException` mechanism but the stacktrace alone does not help with the `System.Net.WebException` and `System.Net.Sockets.SocketException` because the exception themselves  don't have any information about the address they were trying to connect to.

### DIFF
--- a/src/Sentry.Unity/Integrations/UnitySocketExceptionFilter.cs
+++ b/src/Sentry.Unity/Integrations/UnitySocketExceptionFilter.cs
@@ -1,0 +1,14 @@
+using System;
+using Sentry.Extensibility;
+
+namespace Sentry.Unity.Integrations
+{
+    internal class UnitySocketExceptionFilter : IExceptionFilter
+    {
+        internal const string Message = "The requested address is not valid in this context";
+
+        public bool Filter(Exception ex) =>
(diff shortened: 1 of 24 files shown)


src/Sentry.Unity/Integrations/UnitySocketExceptionFilter.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using Sentry.Extensibility;
+
+namespace Sentry.Unity.Integrations
+{
+    internal class UnitySocketExceptionFilter : IExceptionFilter
+    {
+        internal const string Message = "The requested address is not valid in this context";
```
mattjohnsonpint, line 8: Since these are human-readable messages, I wonder if they are ever localized to non-English languages?  Does Unity change these messages if the Editor is configured for some other language, or perhaps if the `CultureInfo.CurrentUICulture` is not English?

mattjohnsonpint:
Is everything in this PR related to this change? Seems like maybe a bad merge?

vaind (approved):

PR #1310, opened 2023-05-05
feat: Added options for debouncing

bitsandfoxes:


src/Sentry.Unity.Editor/ConfigurationWindow/CoreTab.cs
```diff
@@ -23,10 +24,35 @@ internal static void Display(ScriptableSentryUnityOptions options)
...
 
-            options.EnableLogDebouncing = EditorGUILayout.Toggle(
-                new GUIContent("Enable Log Debouncing", "The SDK debounces log messages of the same type if " +
-                                                        "they are more frequent than once per second."),
-                options.EnableLogDebouncing);
+            EditorGUILayout.Space();
+            EditorGUI.DrawRect(EditorGUILayout.GetControlRect(false, 1), Color.gray);
+            EditorGUILayout.Space();
+
+            {
```
brustolin, line 31: l: What is the deal with this block here?

brustolin (approved):
LGTM

src/Sentry.Unity.Editor/ConfigurationWindow/CoreTab.cs
bitsandfoxes: It's just visually blocking all those option blocks to make it easier skimming through.

PR #1315, opened 2023-05-09
chore: Downgrading iOS version

bitsandfoxes:
Someone is lying here.
Based on :https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md there should be `15.5` for macOS-12. But based on CI there isn't.

#skip-changelog


krystofwoldrich (approved):
LGTM 🚀

PR #1345, opened 2023-05-30
feat: Filter Unity telemetry `BadGatewayException` on iOS native layer

bitsandfoxes:
We had filtering of `BadGatewayExceptions` on the C# layer. If enabled, these need to be propagated down to the native layer as well.

### DIFF
--- a/src/Sentry.Unity.Editor.iOS/NativeOptions.cs
+++ b/src/Sentry.Unity.Editor.iOS/NativeOptions.cs
@@ -41,6 +41,11 @@ static SentryOptions* getSentryOptions()
         return nil;
     }}
 
+    {(options.FilterBadGatewayExceptions ? @"options.beforeSend = ^SentryEvent * _Nullable(SentryEvent * _Nonnull event) {
+        if ([event.request.url containsString:@""operate-sdk-telemetry.unity3d.com""]) return nil;
+        return event;
+    };" : "")}
+
     return options;
 }}";
 
--- a/src/Sentry.Unity/SentryUnityOptions.cs
+++ b/src/Sentry.Unity/SentryUnityOptions.cs
@@ -140,6 +140,11 @@ namespace Sentry.Unity
         public TimeSpan AnrTimeout { get; set; } = TimeSpan.FromSeconds(5);
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


src/Sentry.Unity.Editor.iOS/NativeOptions.cs
```diff
@@ -41,6 +41,11 @@ internal static string Generate(SentryUnityOptions options)
         return nil;
     }}
 
+    {(options.FilterBadGatewayExceptions ? @"options.beforeSend = ^SentryEvent * _Nullable(SentryEvent * _Nonnull event) {
+        if ([event.request.url containsString:@""operate-sdk-telemetry.unity3d.com""]) return nil;
```
vaind, line 45: any chance `event.request` or `.url` may be null? Would that still work without crashing/throwing?

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -139,6 +139,11 @@ public new CompressionLevelWithAuto RequestBodyCompressionLevel
         /// </summary>
         public TimeSpan AnrTimeout { get; set; } = TimeSpan.FromSeconds(5);
 
+        /// <summary>
+        /// Whether the SDK should automatically filter `Bad Gateway Exceptions` caused by Unity.
```
bruno-garcia, line 143: Above it indicates that it does this only work for iOS. Should we indicate this in the docs here?
```suggestion
        /// Whether the SDK should automatically filter `BadGatewayException` caused by Unity on iOS.
```

bruno-garcia (approved):

src/Sentry.Unity.Editor.iOS/NativeOptions.cs
bitsandfoxes: In case they are `nil` the if would fail. But that's oke. We don't want to filter those anyway.

src/Sentry.Unity/SentryUnityOptions.cs
bitsandfoxes: There are a whole bunch of places these exceptions are coming from. We already filter them on the C# layer but now, with the same flag because it's the same noise, we also filter the ones coming from native iOS.

PR #1352, opened 2023-05-31
fix: Xcode build with iOS native support disabled

bitsandfoxes:
The Unity SDK expects the bridge to be there and the Xcode build fails during linking.
We now copy a no-op bridge to the output project when the iOS native support is disabled.


src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs
```diff
@@ -51,41 +51,74 @@ internal static bool IsNativeSupportEnabled(SentryUnityOptions options, IDiagnos
...
+                CopyFile(nativeBridgePath, Path.Combine(pathToProject, "Libraries", SentryPackageInfo.GetName(), SentryXcodeProject.BridgeName), logger);
 
                 using var sentryXcodeProject = SentryXcodeProject.Open(pathToProject);
                 sentryXcodeProject.AddSentryFramework();
                 sentryXcodeProject.AddSentryNativeBridge();
                 sentryXcodeProject.AddNativeOptions(options, NativeOptions.CreateFile);
                 sentryXcodeProject.AddSentryToMain(options);
 
-                if (cliOptions?.IsValid(logger, EditorUserBuildSettings.development) is true)
+                if (cliOptions != null && cliOptions.IsValid(logger, EditorUserBuildSettings.development))
```
bitsandfoxes, line 121: `cliOptions` is a Unity object. They don't "like" null-coalescing operators and create warnings.

vaind (approved):

PR #1361, opened 2023-06-06
fix: Image UUID normalization

bitsandfoxes:
Trimming `0` on UUIDs with a trailing `0` leads to a malformed UUID and failure to symbolicate.


kahest (approved):
LGTM 👍

Issue #1322, opened 2023-05-15
JNI:GetDeviceName:java.lang.SecurityException when initializing Sentry when targeting Android API level 32 or later


travis-tap-3d:
### Environment

How do you use Sentry?

Sentry SaaS (sentry.io)

Which version of the SDK?

1.2.0

How did you install the package? (Git-URL, Assetstore)

Git-URL

Which version of Unity?

2021.3.23f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?

Android player

### Steps to Reproduce

Integrate Sentry into a Unity Android project and set the API minimum and target level to 32.

### Expected Result

Initializes without an exception. API 32 is a requirement for some devices in future such as Oculus Quest

### Actual Result

Logs show this exception when  trying to initialize the Sentry SDK:

```
_AndroidPlayer "Oculus_Quest" JNI:AndroidPlayer "Oculus_Quest" JNI:GetDeviceName:java.lang.SecurityException: Settings key: <bluetooth_name> is only readable to apps with targetSdkVersion lower than or equal to: 31
Sentry.Unity.SentryMonoBehaviour:CollectData()
UnityEngine.GameObject:AddComponent()
Sentry.Unity.SentryMonoBehaviour:get_Instance()
Sentry.Unity.ScriptableSentryUnityOptions:ToSentryUnityOptions(Boolean, ISentryUnityInfo, IApplication)
Sentry.Unity.SentryInitialization:Init()

[PlatformDependent/AndroidPlayer/Source/DVM.h line 1703435192]_
```

bitsandfoxes:
Thanks for reporting this. I'm giving this a go locally.

bitsandfoxes:
Looks like requesting `[SystemInfo.deviceName](https://docs.unity3d.com/ScriptReference/SystemInfo-deviceName.html)` and `targetSdkVersion 32` don't get along.

bitsandfoxes:
Hey, I just looked at putting in the workaround but I can't seem to reproduce this on an actual phone. Is this happening for you on the Oculus Quest only?

travis-tap-3d:
Hmm I only have a Quest as our product isn't on mobile so unfortunately I'm unable to test that for you. That'd be surprising since I assume it would be specific to any Android device with that version.

bitsandfoxes:
Resolved by https://github.com/getsentry/sentry-unity/pull/1365 and shipped with `1.5.0`

[...]

PR #1365, opened 2023-06-09
fix: `DeviceName` exception during initialization

bitsandfoxes:
"Resolves" https://github.com/getsentry/sentry-unity/issues/1322


bitsandfoxes:
> Hello, not sure how your release schedule works, but is this targets for a future release version? Thanks!

Yes. It'll land this week.

src/Sentry.Unity/SystemInfoAdapter.cs
```diff
@@ -56,7 +57,32 @@ private SentrySystemInfoAdapter()
...
+                {
+                    using var version = new AndroidJavaClass("android.os.Build$VERSION");
+                    var sdkVersion = version.GetStatic<int>("SDK_INT");
+                    if (sdkVersion >= 32)
+                    {
+                        try
+                        {
+                            return SystemInfo.deviceName;
+                        }
+                        catch (AndroidJavaException)
```
bitsandfoxes, line 75:
```suggestion
                        catch
```

src/Sentry.Unity/SystemInfoAdapter.cs
```diff
@@ -56,7 +57,32 @@ private SentrySystemInfoAdapter()
...
+                        }
+                        catch
+                        {
+                            // Unity's built-in fallback
+                            return "<unknown>";
+                        }
+                    }
+                }
+
+                return SystemInfo.deviceName;
```
vaind, line 83: Maybe try-catch instead of checking a specific version? Also, it may just start working again with a higher SDK version, can't it? In that case this would also prevent returning the actual value.

vaind: lol I'm blind. I've ignored the try-catch :D

vaind (approved):

Issue #1467, opened 2023-10-13
Investigate support for Linux and Windows dedicated server build crashes


InterstellarStella:
Customers have reported that they are able to receive exceptions and errors in Sentry SaaS for Linux and Windows dedicated server builds, but they are unable to receive crashes. It has been confirmed that they can get crashes correctly on Windows client build.

We should investigate whether it is possible to send crashes from Linux and Windows dedicated servers, either out of the box or after some configuration.

Issue #1462, opened 2023-10-11
`Server` are missing from known platforms


bitsandfoxes:
We're missing the `servers` added with 2022 from our known platforms check:
https://github.com/getsentry/sentry-unity/blob/27fb9b8a35fcb0a0917a7b041c84b2249caad4e3/src/Sentry.Unity/ScriptableSentryUnityOptions.cs#L274

bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-unity/issues/1467

PR #1468, opened 2023-10-13
feat: Dedicated Server support

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/1467, https://github.com/getsentry/sentry-unity/issues/1462

Versions newer than `2021.3` have additional enums covering the `dedicated server` runtimes. These are basically just headless builts with assets stripped. 

I moved the Runtime Platform check into `UnityInfo` that we compile with the actual game to have access to the precompile directives.

The only actual difference is on macOS where building a `dedicated macOS server` does not generate an `.app`. The Cocoa Native Bridge now does an additional lookup when trying to load the `dylib`.

App:
```
├── Frameworks
│   ├── GameAssembly.dylib
│   └── UnityPlayer.dylib
├── Info.plist
├── MacOS
│   └── unity-of-bugs
├── PlugIns
│   └── Sentry.dylib
├── Resources
│   ├── Data
│   │   └── Resources
└── _CodeSignature
```
Server:
```
├── Data
│   └── Resources
├── GameAssembly.dylib
├── PlugIns
│   └── Sentry.dylib
├── UnityPlayer.dylib
└── unity-of-bugs
```
and loading fails with `Sentry (bridge): Couldn't load Sentry.dylib - dlopen() failed`


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## Fixes
```
bitsandfoxes, line 5:
```suggestion
## Feature
```

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -17,59 +17,58 @@ public static class SentryNative
         /// Configures the native SDK.
         /// </summary>
         /// <param name="options">The Sentry Unity options to use.</param>
-        public static void Configure(SentryUnityOptions options)
+        /// <param name="sentryUnityInfo">Infos about the current Unity environment</param>
+        public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentryUnityInfo)
         {
-            if (ApplicationAdapter.Instance.Platform switch
+            if (!sentryUnityInfo.IsNativeSupportEnabled(options, ApplicationAdapter.Instance.Platform))
```
vaind, line 23: the original check only worked on Windows and Linux, i.e. platforms where we bundle sentry-native. Now, `IsNativeSupportEnabled` returns true for a lot more platforms. Is that desirable? I guess it's fine because we don't trigger this code unless it's one of the supported platforms?

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -228,10 +231,11 @@ private void HandlePlatformRestrictions(SentryUnityOptions options, IApplication
                     options.BackgroundWorker = new WebBackgroundWorker(options, SentryMonoBehaviour.Instance);
                 }
 
+                // Disable offline caching regardless whether is was enabled or not.
+                options.CacheDirectoryPath = null;
```
vaind, line 235: I don't get why we should do this now.

src/Sentry.Unity.Native/SentryNative.cs
bitsandfoxes: The `IsNativeSupportEnabled` still does the check of the current runtime vs the flag on the options.
But yeah, we only get there if we support the platform in the first place.

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
bitsandfoxes: On "unknown" platforms like Nintendo Switch, we need to turn off any and all file IO as to not crash.

vaind (approved):

PR #1529, opened 2024-01-03
feat: Allow for manual application pause update

bitsandfoxes:
We continuously have issues regarding false positive ANR events coming in combination with Ad-SDKs from providers like AppLovin and Ironsource. These SDKs put the game in the background and or pause them. This seems to circumvent our `SentryMonoBehavious` hooks for pause/focus events.
Making the `UpdatePauseStatus` public allows users to manually unblock themselves by setting the SDK's pause status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features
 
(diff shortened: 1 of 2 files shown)


vaind (approved):
Should we have a doc entry? Or a debug log when an ANR is detected, mentioning the situation with Ad integrations? Or better both?

PR #1567, opened 2024-02-20
Fix missing line numbers in `CaptureMessage` stacktraces in IL2CPP builds

### LINKED ISSUE #1509: CaptureMessage stacktraces in IL2CPP builds are missing line numbers
Events created by capturing messages with `AttachStacktrace = true` do have a stacktrace but are missing line numbers.

tustanivsky:
Closes #1509

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -16,6 +16,7 @@
 
 using System;
 using Sentry.Extensibility;
+using AOT;
 #if UNITY_2020_3_OR_NEWER
 using System.Buffers;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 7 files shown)


munkiki7:
Latest findings:
- if you use the il2cpp API functions to get the current thread stacktrace frames (il2cpp_current_thread_get_stack_depth and iterate using il2cpp_current_thread_get_frame_at)
image
the call is successful only for the lowest frame (less meaningful);
- the method name and function address look valid using the c++ to c# struct mapping for the valid frame;
image
- the actual frames are only captured for the messages sent on Unity UI interactions (by other event processors);
image
- even for the only valid frame we still could not extract the info like on the image above (module and package).

bitsandfoxes:
Closing as wildly out-of-date and stale.

Issue #1558, opened 2024-02-09
Duplicate "embed framework" task added to XCode project if Unity build started with "Append" option


dmytro-kukushkin:
### Environment

How do you use Sentry?
Sentry SaaS 

Which version of the SDK?
1.8

How did you install the package? (Git-URL, Assetstore)
Git-URL

Which version of Unity?
2022.3.14

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Error happens in XCode during build

### Steps to Reproduce

1. Goto "Build Settings" menu, press "Build and Run". If Append/Replace dialog is presented - choose "Replace". Unity builds the project, XCode builds the project.
2. Goto "Build Settings" menu, press "Build and Run". Choose "Append". Unity build is fine, bit XCode project fails to build.
3. Any successive attempts to build project with "Append" option also fail.

### Expected Result

Project builds after choosing "Append" option.

### Actual Result

Project fails to build with error.

### Any logs or screenshots

XCode error is:
Unexpected duplicate tasks:
(next line is duplicated twice)
Target ‘Unity-iPhone’ (project ‘Unity-iPhone’): CodeSign /Users/XXX/Library/Developer/Xcode/DerivedData/Unity-iPhone-dtbfzzrpsmizsaaerhjthdlbizqx/Build/Products/ReleaseForRunning-iphoneos/XXX.app/Frameworks/Sentry.framework

![sentry-bug-1](https://github.com/getsentry/sentry-unity/assets/109065758/8beb0dca-70d7-479f-84e1-2d86b43ff653)

In project settings, there are now two Embed Framework tasks:

![sentry-bug-2](https://github.com/getsentry/sentry-unity/assets/109065758/3560edff-b7c4-4371-8043-674d59009f8a)

New task is added each time project is built with Append option.

bitsandfoxes:
Hey @dmytro-kukushkin, thanks for opening this. It looks like we're missing a check whether the SDK has already been added in our setup method https://github.com/getsentry/sentry-unity/blob/2f323ef4cc2748c91294b4c400b43bf18f732835/src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs#L99

bitsandfoxes:
Fixed by https://github.com/getsentry/sentry-unity/pull/1595

PR #1595, opened 2024-03-22
fix: remove ios embed framework build phase duplication (#1558)

munkiki7:
image

The linking phase is still added multiple times, but I could not find a way to check if it is already added.
But this does not break the build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- remove ios embed framework build phase duplication (#1558) ([#1595](https://github.com/getsentry/sentry-unity/pull/1595))
+
 ### Dependencies
 
 - Bump CLI from v2.30.1 to v2.30.2 ([#1589](https://github.com/getsentry/sentry-unity/pull/1589))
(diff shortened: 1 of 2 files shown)


bitsandfoxes:
> The linking phase is still added multiple times, but I could not find a way to check if it is already added.

Can't we just check for one of the things we do (i.e. adding the framework) and bail early instead of linking it twice?

munkiki7:
> > The linking phase is still added multiple times, but I could not find a way to check if it is already added.
> 
> Can't we just check for one of the things we do (i.e. adding the framework) and bail early instead of linking it twice?

This is a completely separate point, so, for example, if the user builds, then opens the XCode project, and deletes the link task, and builds again, it won't be added. This is a low-probability case, and a clean build will fix it.

So, do you want me to leave it as is, or do only one check for framework and linking as you have proposed?

bitsandfoxes:
> This is a completely separate point, so, for example, if the user builds, then opens the XCode project, and deletes the link task, and builds again, it won't be added. This is a low-probability case, and a clean build will fix it.

I think this is very much self-inflicted then and we're good.

bitsandfoxes (approved):
Thanks!

PR #1603, opened 2024-04-03
fix: Reinstall native backend 

bitsandfoxes:
I'm not sure (yet) when that happened but Unity is overwriting the signal handler in newer versions. This prevents `sentry-native` from capturing native crashes. 
We have a similar situation on Android where we reinstall the backend to take back control so I'm taking the same approach to resolve this.
An open TODO: We have checks in CI specifically to make sure that does not break. Why does CI pass?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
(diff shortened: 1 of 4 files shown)


munkiki7 (approved):

bruno-garcia:
Is there a ticket this is related to?

Issue #1620, opened 2024-04-15
Failing to capture native crashes on Windows


bitsandfoxes:
Originally raised on [discord](https://discord.com/channels/621778831602221064/1221853579376201808)

The SDK fails to capture native crashes triggered through `Utils.ForceCrash(ForcedCrashCategory.Abort);` on the current LTS (2021.3.37f1, 2022.3.24f1)
Other relevant crashes are
```
AccessViolation
FatalError
Abort
```

It looks like these Unity versions overwrite our backend. We've attempted to reinstall the backend in [this PR](https://github.com/getsentry/sentry-unity/pull/1603). That fixes the issue for C `SEGFAULTS` and `FatalError` but does not help with C++ exceptions. Turns out we're reinstalling the backend too early and the input system overwrites some of the handlers again.
We need to find a place to reliably re-install the sentry backend. (i.e. before/after the splashscreen?)

PR #1622, opened 2024-04-15
fix: Reinstall backend to capture native crashes

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/1620

Reinstalling the backend during `RuntimeInitializeLoadType.SubsystemRegistration` is still too early and the handler still gets overwritten.

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -131,7 +131,6 @@ namespace Sentry.Unity
 #endif
     }
 
-
     public class SentryUnityInfo : ISentryUnityInfo
     {
         public bool IL2CPP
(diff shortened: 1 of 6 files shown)


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -118,8 +118,20 @@ public static void Init()
...
+        {
+            // At this point Unity has taken the signal handler and will not invoke our handler. So we register our
+            // backend once more to make sure user-defined data is available in the crash report and the SDK is able
+            // to capture the crash.
+            SentryNative.ReinstallBackend();
+        }
+#endif
     }
 
+
```
bitsandfoxes, line 134:
```suggestion
```

munkiki7 (approved):

PR #1623, opened 2024-04-15
fix: Screenshot in CI

bitsandfoxes:
Looks like just setting the capacity is flaky.

#skip-changelog

### DIFF
--- a/src/Sentry.Unity/ScreenshotAttachment.cs
+++ b/src/Sentry.Unity/ScreenshotAttachment.cs
@@ -33,9 +33,10 @@ namespace Sentry.Unity
 
             if (Screen.width == 0 || Screen.height == 0)
             {
-                _options.DiagnosticLogger?.LogDebug("Can't capture screenshots on a screen with a resolution of '{0}x{1}'.", Screen.width, Screen.height);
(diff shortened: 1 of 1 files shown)


src/Sentry.Unity/ScreenshotAttachment.cs
```diff
@@ -33,9 +33,10 @@ public Stream GetStream()
 
             if (Screen.width == 0 || Screen.height == 0)
             {
-                _options.DiagnosticLogger?.LogDebug("Can't capture screenshots on a screen with a resolution of '{0}x{1}'.", Screen.width, Screen.height);
+                _options.DiagnosticLogger?.LogWarning("Can't capture screenshots on a screen with a resolution of '{0}x{1}'.", Screen.width, Screen.height);
                 // Returning a memory stream with a capacity of 1 so we can smoke-test the attempt to capture a screenshot in CI
-                return new MemoryStream(1);
+                return new MemoryStream(new byte[] { 0x00 });
```
bruno-garcia, line 38: A bit odd we need to return a byte array with `0` in it though

bruno-garcia (approved):

PR #1637, opened 2024-05-02
fix: Setup Wizard

bitsandfoxes:
The project format in the response changed: `platform` and `name` seem to be gone.


bitsandfoxes:
Fixed via https://github.com/getsentry/sentry/pull/70130 instead

Issue #1591, opened 2024-03-19
SIGSEGV errors on old Android devices when Sentry is enabled.


rodolfoBee:
### Environment

How do you use Sentry?
Sentry SaaS 

Which version of the SDK?
>=1.7.0

Which version of Unity?
2022.3.13f1

### Steps to Reproduce
Reproduction app is available in the internal issue. 

### Expected Result
The application runs with no errors on all devices. 

### Actual Result
Android 5 and older devices experience a SIGSEGV error. Link to event also in the internal ticket.

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2127) by [Unito](https://www.unito.io)

rodolfoBee:
@bitsandfoxes @kahest the affected user is asking for an update, is there any?

kahest:
@rodolfoBee I don't have any updates, I'm not working on this currently

bitsandfoxes:
I think we're a bit lost here without a proper way to reproduce this.
It is possible that the changes that have been made to get rid of our custom sigaltstacks in `sentry-native` could cause an increase in reported `SIGSEGV`. But for that to happen there needs to be a signal first. So basically: Sentry is crashing while capturing a crash.

Talibt:
Hey @bitsandfoxes 
Thank you!
Yeah sure, that would be easy to do.

bitsandfoxes:
Hi @Talibt,
I've created an "experimental" package with the rollback on the sentry-native [here](https://drive.google.com/file/d/1SXDZ6I2g_nO6Q8FtlPV2LMyHA6hbbk-4/view?usp=sharing).
This is strictly intended to work on and for those old Android devices. If you could give that a try and see if it has any effect on the crashes you're seeing, that'd be hugely appreciated.

Talibt:
Hey @bitsandfoxes, 
sorry to inform you that the hand-made package didn't work as expected. The lower-end devices are still crashing. 
However, we also tested your Unity Lite EPK, available here: https://github.com/getsentry/sentry-unity-lite, 
and it worked smoothly. 
What do you think is needed to move forward with fixing this issue?

bitsandfoxes:
One thing you could do immediately to unblock yourself from those issues on low-end devices is to disable android-native support. That way you get all of the offerings the Unity SDK has without having to rely on the Lite package.
Could you provide us with a build that keeps crashing?

Talibt:
[...]

PR #1652, opened 2024-05-14
fix: Android 5 attachment issues

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/1591

It looks like for older Android versions we have to make sure to manually detach from the JVM to not end up blowing up.


src/Sentry.Unity.Android/AndroidJavaScopeObserver.cs
```diff
@@ -14,63 +14,74 @@ public class AndroidJavaScopeObserver : ScopeObserver
...
-            AndroidJNI.AttachCurrentThread();
-            using var sentry = GetSentryJava();
-            using var javaBreadcrumb = new AndroidJavaObject("io.sentry.Breadcrumb");
-            javaBreadcrumb.Set("message", breadcrumb.Message);
-            javaBreadcrumb.Set("type", breadcrumb.Type);
-            javaBreadcrumb.Set("category", breadcrumb.Category);
-            using var javaLevel = breadcrumb.Level.ToJavaSentryLevel();
-            javaBreadcrumb.Set("level", javaLevel);
-            sentry.CallStatic("addBreadcrumb", javaBreadcrumb, null);
+            SentryJniExecutor.FireAndForget(() =>
```
vaind, line 17: I understand the issue is that we must call `DetachCurrentThread()`. Why then make the code asynchronous? That is likely to have unintended consequences elsewhere.

Howe about we just wrap all these functions in closures (which you already did) and instead of running on a separate thread, do it synchronously as it is now, but also detach in the finally block?

vaind (changes requested):

src/Sentry.Unity.Android/AndroidJavaScopeObserver.cs
bitsandfoxes: I tried that but kept running into `attempting to detach while still running code` errors.

bitsandfoxes: Which, fair, we could do that for the `FireAndForget` specifically.

vaind: Alright, I've searched a bit about the error in the original issue and found some solutions. The problem with launching a separate thread in each call is that it has a significant cost.

Instead, I can think of two (and a half) solutions:
1. roll back to running on the same thread as the caller and use `Android_JNI_SetupThread` the first time it's used (guarded by a thread-local boolean flag) & `Android_JNI_GetEnv` as needed. It's decribed [here](https://android-ndk.narkive.com/uqy1gqKj/can-t-execute-javavm-detachcurrentthread-attempting-to-detach-while-still-running-code), with the native [source code](https://github.com/ennorehling/android-sdl/blob/master/jni/SDL2/src/core/android/SDL_android.c#L429), if it's feasible to include those (probably a pain, right?)
2. add a custom worker thread we let this run on, with a blocking `wait()` call so we don't make this asynchronous.
3. number 2 for old platforms we know this breaks on, original code elsewhere.

src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -10,53 +9,63 @@ namespace Sentry.Unity.Android
     /// </summary>
     public static class SentryNativeAndroid
     {
+<<<<<<< HEAD
+        private static JniExecutor? JniExecutor;
+=======
+        private static readonly JniExecutor JniExecutor = new();
+>>>>>>> 919a2cbed3f67a4342036097c1c1fa9c55361eda
+
```
bitsandfoxes, line 17:
```suggestion
        private static JniExecutor? JniExecutor;
```

src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -10,53 +9,58 @@ namespace Sentry.Unity.Android
     /// </summary>
     public static class SentryNativeAndroid
     {
+        private static JniExecutor? JniExecutor;
```
bitsandfoxes, line 12:
```suggestion
        private static JniExecutor? JniExecutor;
        
```

vaind (approved):
👏 looks great!

PR #1655, opened 2024-05-17
Improve C++ exceptions stacktrace on iOS

### LINKED ISSUE #1628: C++ exception & std::terminate() support in iOS
1. Now when a C++ exception is thrown in iOS, sentry cannot collect the call stack of the exception thread, but can only collect the stacks of other unrelated threads. This issue has been mentioned in following issues: 
- https://github.com/getsentry/sentry-cocoa/issues/551
- https://github.com/getsentry/sentry-unity/issues/1273

I tried [the solution proposed by eakoli](https://github.com/getsentry/sentry-cocoa/issues/551#issuecomment-684089619) but still didn't solve the problem. However, I'm inspired from his solution. We can export a `__sentry_cxa_throw` function in sentry.framework:
img_v3_02a3_5173ea99-aa0f-41ec-b698-38f8ce4e32fg

And then add the following code through unity’s c++ plugin mechanism to ensure that all c++ exceptions thrown in UnityFramework can be collected by sentry:
```
// This file should be added into UnityFramework target in xcode
#include <dlfcn.h>


typedef void (*cxa_throw_type)(void *, std::type_info *, void (*)(void *));

extern "C" 
{
    void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(void *)) __attribute__((weak));

    void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(void *))
    {
        static cxa_throw_type orig_cxa_throw = NULL;
        if (orig_cxa_throw == NULL)
        {
            orig_cxa_throw = (cxa_throw_type)dlsym(RTLD_DEFAULT, "__sentry_cxa_throw");

            if (orig_cxa_throw == NULL)
                orig_cxa_throw = (cxa_throw_type)dlsym(RTLD_NEXT, "__cxa_throw");
        }
        orig_cxa_throw(thrown_exception, tinfo, dest);
        __builtin_unreachable();
    }
}
```

2. In addition, calling `std::terminate()` directly in C++ cannot be correctly captured by sentry. I investigate the sentry-cocoa logs, found that sentry tried to re-throw the current exception while processing terminate(`CPPExceptionTerminate`), but the re-throwed exception was not captured correctly. Because there is currently no exception. Here's a solution I've tested that works:
img_v3_02a3_641d4482-b9b9-46c3-8878-6925559e38cg

munkiki7:
Fixes #1628 

There are still some cases that are not covered by this fix.
If the exception is thrown not from UnityFramework, but from another library, the hook will not be called.

The details can be found in the conversation of the issue.

### DIFF
--- a/package-dev/Plugins/iOS/SentryCxaThrowHook.cpp
+++ b/package-dev/Plugins/iOS/SentryCxaThrowHook.cpp
@@ -1,5 +1,6 @@
 #include <dlfcn.h>
 #include <typeinfo>
+#include <pthread.h>
 
 extern "C"
 {
--- a/package-dev/Plugins/iOS/SentryCxaThrowHook.cpp
+++ b/package-dev/Plugins/iOS/SentryCxaThrowHook.cpp
@@ -1,0 +1,51 @@
+#include <dlfcn.h>
+
+extern "C"
+{
+    // Function type __cxa_throw
+    typedef void(*cxa_throw_type)(void*, std::type_info*, void(*)(void*));
+    // Function type __cxa_rethrow
+    typedef void(*cxa_rethrow_type)(void);
+
+    static pthread_mutex_t guard = PTHREAD_MUTEX_INITIALIZER;
+    static cxa_throw_type orig_cxa_throw = NULL;
+    static cxa_rethrow_type orig_cxa_rethrow = NULL;
+
+    void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(void *)) __attribute__((weak));
+
+    void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(void *))
+    {
+        pthread_mutex_lock(&guard);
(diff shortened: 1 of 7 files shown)


bitsandfoxes:
That does `UnityFramework` mean. Are native plugins bundled in the output part of the `UnityFramework`?

munkiki7:
> That does `UnityFramework` mean? Are native plugins bundled in the output part of the `UnityFramework`?

All native plugin files in the Unity Project (in the Assets directory, as well as in Packages) will be linked to UnityFramework target, and will be properly handled.

package-dev/Plugins/iOS/SentryCxaThrowHook.cpp
bitsandfoxes, line 1: So this gets copied to the generated Xcode project by Unity. Any chance this causes issues with Sentry disabled? The includes look fine to me.

bitsandfoxes (approved):

PR #1657, opened 2024-05-21
change gradle task syntax and placement

munkiki7:
Reason for changes: random android smoke tests failures during CI tasks.

We have noticed that sometimes the upload task is executed before the assemble task is finished, which resulted in not all .so files being generated at the moment it was executed. After that the uploaded symbols validation was failing.

Now we are attaching the symbol upload task to either `assembleDebug` or `assembleRelease` tasks of the launcher build.gradle instead of the root build.gradle file.

One of the other advantages is that with the previous approach the symbol upload task was executed during the gradle sync process in the Android Studio (if the project was exported), as it was bound to the last task being executed (even though it was not a build-related task).

Previous syntax:
```
gradle.taskGraph.whenReady { taskGraph ->
    gradle.taskGraph.allTasks[-1].doLast { task ->
        //task body
    }
}
```
New syntax (needs to be wrapped in the afyterEvaluate call, because we have to wait until the task list is built):
```
afterEvaluate {
    task sentryUploadSymbols {
        doLast {
            //task body
        }
    }
    
    tasks.assembleDebug.finalizedBy sentryUploadSymbols
    tasks.assembleRelease.finalizedBy sentryUploadSymbols
}
```


src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
```diff
@@ -53,8 +54,8 @@ private string _symbolUploadTaskFormat
                     stringBuilder.AppendLine($"        def sentryLogFile = new FileOutputStream('{logsDir}/sentry-symbols-upload.log')");
                 }
                 stringBuilder.AppendLine("        exec {");
-                stringBuilder.AppendLine("            environment 'SENTRY_PROPERTIES', './sentry.properties'");
-                stringBuilder.AppendLine($"            executable '{SentryCliMarker}'");
+                stringBuilder.AppendLine("            environment 'SENTRY_PROPERTIES', file(\"${rootDir}/sentry.properties\").absolutePath");
```
bitsandfoxes, line 57: Where is the change to the path coming from?

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
```diff
@@ -81,7 +86,7 @@ private string _symbolUploadTaskFormat
 
             _unityProjectPath = unityProjectPath;
             _gradleProjectPath = gradleProjectPath;
-            _gradleScriptPath = Path.Combine(_gradleProjectPath, "build.gradle");
+            _gradleScriptPath = Path.Combine(_gradleProjectPath, "launcher/build.gradle");
```
bitsandfoxes, line 89: Ah, I see. Because the SDK now modifies the `launcher/build.gradle`. Why?

bitsandfoxes:
Thanks for tackling this!

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
munkiki7: The task is now added to the topmost project - launcher that depends on unityFramework target.

We are now executing the symbol upload task after the `assembleDebug` or `assembleRelease` tasks, which do not exist in the scope of the `build.gradle` file we were modifying previously.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- change gradle task syntax and placement ([#1657](https://github.com/getsentry/sentry-unity/pull/1657))
+
 ### Fixes
 
 - The SDK no longer crashes on Android versions 5 and 6 with native support enabled ([#1652](https://github.com/getsentry/sentry-unity/pull/1652))
```
bitsandfoxes, line 9:
```suggestion
### Fixes

- Updated the symbol upload task on Gradle to be guaranteed to run last ([#1657](https://github.com/getsentry/sentry-unity/pull/1657))
- The SDK no longer crashes on Android versions 5 and 6 with native support enabled ([#1652](https://github.com/getsentry/sentry-unity/pull/1652))
```

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
munkiki7: That depends on whether anything else is also using the locations of those files. But we can check.

bitsandfoxes: Sentry CLI is the one looking for the `.properties` so I think it's safe to keep all that in once place.

bitsandfoxes (approved):

PR #1659, opened 2024-05-22
chore: Log when configuring native SDKs

bitsandfoxes:
I keep running into situations where I want to know whether the native SDK was configured or enabled but the logs won't tell me.

#skip-changelog


munkiki7 (approved):
LGTM

PR #1676, opened 2024-06-03
chore: Add `2023` to CI

bitsandfoxes:
#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,7 +259,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


bruno-garcia:
Looks like we really need this, does removing 2019 from CI help?
Saw the bug fixes for Unity 6 also in place, would CI catch those?

bitsandfoxes:
> Saw the bug fixes for Unity 6 also in place, would CI catch those?

It would have. We're updating CI next. I'll try to gather some data regarding what Unity versions are actually in use to make some more informed decisions.

bitsandfoxes:
Supersede by https://github.com/getsentry/sentry-unity/pull/1775

PR #1677, opened 2024-06-03
ci: fix unity update script

vaind:
There have been changes to the download page. This should fix the updater CI


bitsandfoxes (approved):
Thanks for taking care of this!

PR #1679, opened 2024-06-04
do not set exception on task failure, improve single call enforcement

### LINKED ISSUE #1675: v2.0.6 version Android initialization SentrySDK will be stuck
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)
Sentry SaaS (sentry.io)

Which version of the SDK?
v2.0.6

Git-URL

Which version of Unity?
2021.3.33f1

Android

### Steps to Reproduce

SDK upgrade to the latest version initialization will get stuck （v2.0.6）

### Expected Result

What you thought would happen.

### Actual Result

What actually happened. Maybe a screenshot/recording? Maybe some logs?

munkiki7:
Fixes #1675.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- The SDK no longer causes deadlocks during synchronization with the native layer on Android ([#1679](https://github.com/getsentry/sentry-unity/pull/1679))
+
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


bitsandfoxes:
We definitely need a changelog entry such as "The SDK no longer causes deadlocks during synchronization with the native layer on Android".

bitsandfoxes (approved):

PR #1687, opened 2024-06-10
feat: add a flag whether to ignore the CLI errors during Android symbol upload

munkiki7:
Added a `IgnoreCliErrors` flag to Sentry CLI Options, that will wrap the symbol and mapping upload Gradle tasks in `try/catch` blocks if it is toggled, so if they fail, the Unity build will still succeed. It is disabled by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Added a flag whether to ignore CLI errors during symbol and mapping uploads ([#1687](https://github.com/getsentry/sentry-unity/pull/1687))
+- Added an `IgnoreCliErrors` to the Sentry-CLI options, allowing you to ignore errors during symbol and mapping upload ([#1687](https://github.com/getsentry/sentry-unity/pull/1687))
 
 ### Dependencies
 


src/Sentry.Unity.Editor/ConfigurationWindow/DebugSymbolsTab.cs
```diff
@@ -21,6 +21,10 @@ internal static void Display(SentryCliOptions cliOptions)
                 new GUIContent("Upload Sources", "Whether your source code should be uploaded to Sentry, so the stack trace in Sentry has the relevant code next to it."),
                 cliOptions.UploadSources);
 
+            cliOptions.IgnoreCliErrors = EditorGUILayout.Toggle(
+                new GUIContent("Ignore CLI Errors", "Whether to ignore the Sentry CLI errors during the build. If this is ON, the errors will be logged, but the build will succeed."),
```
bitsandfoxes, line 25:
```suggestion
                new GUIContent("Ignore Symbol Upload Errors", "Whether to ignore the Sentry CLI errors during the build. If this is ON, the errors will be logged, but the build will succeed."),
```

I'm not sure if there's enough space for that "name" but otherwise we can drop the "symbol".

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Added a flag whether to ignore CLI errors during symbol and mapping uploads ([#1687](https://github.com/getsentry/sentry-unity/pull/1687))
```
```suggestion
- Added an `IgnoreCliErrors` to the Sentry-CLI options, allowing you to ignore errors during symbol and mapping upload ([#1687](https://github.com/getsentry/sentry-unity/pull/1687))
```

bitsandfoxes (approved):
Awesome!

PR #1691, opened 2024-06-17
Pass CLI logs to Unity in any case for Android build

munkiki7:
The logs will be passed as warnings if the IgnoreCliErrors flag is set in the settings, and will be logged as info if the task was successful.

#skip-changelog


bitsandfoxes (approved):
Awesome!

PR #1710, opened 2024-07-02
fix: User ID fallback on mobile

bitsandfoxes:
User reported events without any user ID coming from the C# layer. This should work out of the box. In case of `user.Id` being null the `enricher` would set the installation ID.
https://github.com/getsentry/sentry-dotnet/blob/82dcbcdadaf37670729b394bf47489cc73fc7aa1/src/Sentry/Internal/Enricher.cs#L83

But if the ID was an empty string, it would not get overwritten by the enricher, it would not get serialized, and it would not show up on the issue.

To prevent those edge-cases I added fallbacks to both Android and iOS, relying on the same mechanism as we do on desktop to create a new ID and pass it down.


src/Sentry.Unity.Android/IJniExecutor.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace Sentry.Unity.Android
+{
```
bruno-garcia, line 4: can u use top level `namespace` here already?

src/Sentry.Unity.iOS/SentryNativeCocoa.cs
```diff
@@ -58,6 +59,21 @@ internal static void Configure(SentryUnityOptions options, ISentryUnityInfo sent
             if (sentryUnityInfo.IL2CPP)
             {
                 options.DefaultUserId = SentryCocoaBridgeProxy.GetInstallationId();
+                if (string.IsNullOrEmpty(options.DefaultUserId))
+                {
+                    // In case we can't get an installation ID we create one and sync that down to the native layer
+                    options.DiagnosticLogger?.LogDebug("Failed to fetch 'Installation ID' from the native SDK. Creating new 'Default User ID'.");
+
+                    options.DefaultUserId = AnalyticsSessionInfo.userId;
```
bruno-garcia, line 67: `AnalyticsSessionInfo.userId`

Is this something devs would need to add to their app manifest now? Might be a problem to use something that can be used to track users across apps. That's why we use our own `guid` everywhere

bitsandfoxes: I picked `AnalyticsSessionInfo.userId` as we currently use it for the desktop platforms. It's a GUID generated during installation. Based on the [Unity Docs] it only gets regenerated during a re-install. (https://docs.unity3d.com/2018.3/Documentation/ScriptReference/Analytics.AnalyticsSessionInfo-userId.html)

For the Unity SDK to use the .NET SDKs installation ID we'd need to make that public.
https://github.com/getsentry/sentry-dotnet/blob/d88635187ba02b4e47906a38b5bc0ebb19548355/src/Sentry/SentryOptions.cs#L46

src/Sentry.Unity.Android/IJniExecutor.cs
bitsandfoxes: I might, but everything in that `.csproj` is inside the `Sentry.Unity.Android` namespace.

bitsandfoxes:
The user confirmed that the build from this PR fixed the issue for them.

src/Sentry.Unity.iOS/SentryNativeCocoa.cs
bruno-garcia: oh OK, so it uses the same principal as we do.

Worth adding a big block of comment clarifying that btw for anyone reasing the code and being worried about tracking and privacy

bruno-garcia (approved):
Approving to unblock, strongly suggest the comment about how the user-id

Issue #1713, opened 2024-07-04
App freezes at startup with Sentry disabled


aleh-patseyeu:
### Environment
Android

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
2.1.0

How did you install the package? (Git-URL, Assetstore)
Git-URL

Which version of Unity?
2022.3.28f

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Android

### Steps to Reproduce

1. Select Android as a build platform
2. Set scripting backend to IL2CPP
3. Disable Sentry in SDK Options

### Expected Result
Apps starts normally. Sentry is disabled.

### Actual Result
App freezes at startup.

What actually happened. Maybe a screenshot/recording? Maybe some logs?
Seems like some Sentry code is being stripped during build process. Adding `io.sentry.unity.runtime` assembly to `link.xml` or configuring progruad rules doesn't help.

```
2024/07/04 17:47:09.748 17067 17124 Error Unity Sentry: Error during JNI execution: UnityEngine.AndroidJavaException: java.lang.ClassNotFoundException: io.sentry.Sentry
2024/07/04 17:47:09.748 17067 17124 Error Unity   at UnityEngine.AndroidJNISafe.CheckException () [0x00000] in <00000000000000000000000000000000>:0 
2024/07/04 17:47:09.748 17067 17124 Error Unity   at UnityEngine.AndroidJNISafe.FindClass (System.String name) [0x00000] in <00000000000000000000000000000000>:0 
2024/07/04 17:47:09.748 17067 17124 Error Unity   at UnityEngine.AndroidJavaClass._AndroidJavaClass (System.String className) [0x00000] in <00000000000000000000000000000000>:0 
2024/07/04 17:47:09.748 17067 17124 Error Unity   at Sentry.Unity.Android.SentryJava.GetSentryJava () [0x00000] in <00000000000000000000000000000000>:0 
2024/07/04 17:47:09.748 17067 17124 Error Unity   at Sentry.Unity.Android.SentryJava+<>c.<Close>b__2_0 () [0x00000] in <00000000000000000000000000000000>:0 
2024/07/04 17:47:09.748 17067 17124 Error Unity   at Sentry.Unity.Android.JniExecutor.DoWork () [0x00000] in <00000000000000000000000000000000>:0 
[...]

PR #1714, opened 2024-07-05
Fix: Android Native initialization without SDK

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/1713

This PR covers and prevents exceptions when the SDK tries to initialize Sentry Java SDK at runtime but is not present. (i.e. disabled during build-time)


tustanivsky:
Looks good!

PR #1749, opened 2024-07-30
deps: update actions/upload-artifact to v4

joshuarli:
Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.github/workflows/android-smoke-test.yml
+++ b/.github/workflows/android-smoke-test.yml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Upload artifacts on failure
         if: ${{ failure() }}
-        uses: actions/upload-artifact@v3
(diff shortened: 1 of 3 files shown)


bitsandfoxes (approved):

bitsandfoxes:
I don't think this has anything to do with craft but with https://github.com/vaind/download-artifact forking v2

vaind:
> I don't think this has anything to do with craft but with [vaind/download-artifact](https://github.com/vaind/download-artifact) forking v2

I haven't been tagged here so only now noticed this by accident. [Here's an updated version of the download action](https://github.com/vaind/download-artifact/releases/tag/v4-with-wait-timeout) & I've pushed a commit to this branch to try & fix CI

bitsandfoxes (approved):
Thanks @vaind, that works!

PR #1754, opened 2024-08-05
Fix `SentryMonoBehavior` instance being removed on scene switch in WebGL

### LINKED ISSUE #1678: Events not captured from Unity WebGL build after scene change
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
2.0.6

Git-URL

Which version of Unity?
2021.3.32f1

It is happening on WebGL build(dont know if on another platforms also, did not tested). In Editor bug is NOT reproduced

### Steps to Reproduce

1. Integrate Sentry SDK events/messages/errors/unhandledExceptions capturing on first and second scenes via SentrySdk.Capture[] methods. 
2. Capturing executed from UniTask loop in custom non MonoBehaviour classes.
2. Make game build on WebGL platform(dev/usual, doesn't matter)
3. Run build on any browser(tested on Chrome and Firefox, both reproduced bug).

### Expected Result
All calls of SentrySdk.Capture[] methods and unhandled exceptions will be gathered and displayed on sentry.io.

### Actual Result
While first scene is loaded and active all events sending to cloud correctly. But after a moment of second scene loaded all events stops to send. After this, each sending try throws NullReference exception to browser console from internal SentrySDK code(see the screenshots).
image

Exceptions thrown from internal DoSendEvent method and catched by try/catch block.
image

### Any logs or screenshots
SentryConfiguration on which build was tested:
image
image
image
image
image
image
image

Detailed log of Error message:
image

tustanivsky:
This PR adds `DontDestroyOnLoad` call to `SentryMonoBehavior.Awake` to ensure its instance won't be destroyed during the scene switch in WebGL builds.

Closes #1678

### DIFF
--- a/src/Sentry.Unity/SentryMonoBehaviour.cs
+++ b/src/Sentry.Unity/SentryMonoBehaviour.cs
@@ -122,12 +122,12 @@ public partial class SentryMonoBehaviour
     // We want to do it this way instead of a StartCoroutine() so that we have the context info ASAP.
     private void Awake()
     {
-        CollectData();
-
         // This prevents object from being destroyed when unloading the scene since using HideFlags.HideAndDontSave
         // doesn't guarantee its persistence on all platforms i.e. WebGL
         // (see https://github.com/getsentry/sentry-unity/issues/1678 for more details)
         DontDestroyOnLoad(gameObject);
+        
+        CollectData();
     }
 
     internal void CollectData()
--- a/src/Sentry.Unity/SentryMonoBehaviour.cs
+++ b/src/Sentry.Unity/SentryMonoBehaviour.cs
@@ -120,7 +120,15 @@ public partial class SentryMonoBehaviour
 
     // Note: Awake is called only once and synchronously while the object is built.
     // We want to do it this way instead of a StartCoroutine() so that we have the context info ASAP.
-    private void Awake() => CollectData();
+    private void Awake()
+    {
+        CollectData();
+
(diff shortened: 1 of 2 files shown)


src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -120,7 +120,15 @@ internal ISentrySystemInfo SentrySystemInfo
...
     // We want to do it this way instead of a StartCoroutine() so that we have the context info ASAP.
-    private void Awake() => CollectData();
+    private void Awake()
+    {
+        CollectData();
+
+        // This prevents object from being destroyed when unloading the scene since using HideFlags.HideAndDontSave
+        // doesn't guarantee its persistence on all platforms i.e. WebGL
+        // (see https://github.com/getsentry/sentry-unity/issues/1678 for more details)
+        DontDestroyOnLoad(gameObject);
```
bitsandfoxes, line 130:
```suggestion
        // This prevents object from being destroyed when unloading the scene since using HideFlags.HideAndDontSave
        // doesn't guarantee its persistence on all platforms i.e. WebGL
        // (see https://github.com/getsentry/sentry-unity/issues/1678 for more details)
        DontDestroyOnLoad(gameObject);
        
        CollectData();
```

bitsandfoxes: I think we can call this first. But that's just a random nitpick.

bitsandfoxes (approved):

PR #1794, opened 2024-09-10
chore: Local integration respects Unity version

bitsandfoxes:
What this allows is to call 
```
pwsh ./test/Scripts.Integration.Test/integration-test.ps1 -Platform "Android-Export" -UnityVersion "6000" -Rebuild
```

with different versions and have them show up in the `/samples` like so
```
samples git:(feat/integration-test-variable) ✗ tree
.
├── IntegrationTest2019
├── IntegrationTest2020
├── IntegrationTest2021
├── IntegrationTest2022
├── IntegrationTest6000
└── unity-of-bugs
```
making looking for diffs or running tests on different versions a lot less painful and tedious.

#skip-changelog


scripts/smoke-test-android.ps1
```diff
@@ -6,6 +6,11 @@ param (
     [string] $UnityVersion = ""
 )
 
+if (-not $Global:NewProjectPathCache)
```
vaind, line 9: You could maybe move this check to `globals.ps1`, making it return early

vaind (approved):

scripts/smoke-test-android.ps1
bitsandfoxes: I don't think that works. It's not about returning early it's about the dot-scope. The whole script gets executed in the caller's scope, messing with the global variables.

PR #1802, opened 2024-09-12
feat: Collect `MainThreadData` before initial scene load

bitsandfoxes:
Initially, we deferred the collection of `MainThreadData` into our MonoBehaviour's Awake call. Since that data does not change at runtime, fetching it doesn't impose any performance impact (we were just doing it at a later point) we can rely on the `[RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]` attribute. That way we have the data available right after initialization and irrespective whether an error gets captured on the main/UI or a background thread.


bruno-garcia (approved):
makes sense

PR #1803, opened 2024-09-12
feat: Added `Func<bool> IsMainThread` to options

bitsandfoxes:
Added a hook to the options to allow overwriting the check whether the current thread is the main thread.


src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -194,6 +194,11 @@ public sealed class SentryUnityOptions : SentryOptions
     /// </summary>
     public bool PerformanceAutoInstrumentationEnabled { get; set; } = false;
 
+    /// <summary>
+    /// Allows to overwrite the check to determine whether the current thread is the main thread
```
bruno-garcia, line 198: I'm lacking an understanding of why would you need to override it. We should ask folks who need this hook to elaborate on the use case so we can expand on this summary here

bitsandfoxes:
Closing this as stale

PR #1806, opened 2024-09-18
Change access modifier to `private` for stack trace mode option setter

### LINKED ISSUE #1783: WebGL exception during IL2CPP processing
Reported on [Discord](https://discord.com/channels/621778831602221064/1246915557731860511/1278277335615012935)

Triggering an exception causes the IL2CPP backend to throw one of its own.

Image
Image

We might need to disable the automatic integration of the IL2CPP exception processor for WebGL.

tustanivsky:
This PR should prevent users from accidentally enabling `Enhanced` stack trace mode via the `SentryUnityOptions` parent class property. Enhanced mode is not supported in IL2CPP builds (see #1033 for more details) and thus could cause crashes like in #1783.

Closes #1783

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,15 +13,12 @@
 - The SDK now uses the .NET SDK's `Options.FileWriteDisabled` to opt-out on any writing operations on 'unknown' platforms such as the Nintendo Switch ([1814](https://github.com/getsentry/sentry-unity/pull/1814))
 - The SDK no longer freezes the game during shutdown when targeting WebGL ([#3619](https://github.com/getsentry/sentry-dotnet/pull/3619))
 - The SDK no longer passes a mangled `gpu.vendorId` to the Android native layer ([#1813](https://github.com/getsentry/sentry-unity/pull/1813))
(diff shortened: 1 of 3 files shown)


bruno-garcia (changes requested):
Looks like the whole property was `private` though (getter and setter). We're making the getter public but the setter private. Is this helping somehow?

bruno-garcia:
> 
What we wanted to achieve was take away the footgun for users to switch mode and break their game by hiding the property. But it's still accessible via the base class. With this we're giving the "no-footgun!" attempt a second chance.


As per my comment, isnt' the whole thing `private` and now we're making half private? If not, I might misundersatnd the code.

bitsandfoxes:
> As per my comment, isnt' the whole thing private and now we're making half private? If not, I might misundersatnd the code.

Making the property private on the child only hides it on the child. Accessing the property i.e. in the configure callback falls back directly to the `SentryOptions`. 

Screenshot 2024-10-07 at 16 23 37

Keeping the property `public` but overwriting the accessor on the setter allows us to put a hurdle in front of the footgun.

Screenshot 2024-10-07 at 16 22 48

bitsandfoxes (approved):
Thanks for finding this!

bruno-garcia:
Thank you for clarifying that

Issue #1804, opened 2024-09-12
Application crashes on Nintendo Switch when reporting an error via Sentry


abond-sd:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which version of the SDK?
2.1.3

How did you install the package? (Git-URL, Assetstore)
Git URL

Which version of Unity?
2022.3.37f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Nintendo Switch

### Steps to Reproduce

1. Start the application on Switch
2. Have Switch report an error via Sentry
3. --> The application closes

### Expected Result

Reporting an error should not terminate the application

### Actual Result

Based on previous advice for Sentry on Switch we had already set the following in a SentryBuildTimeOptionsConfiguration class
```
options.CacheDirectoryPath = null;
options.AutoSessionTracking = false;
```

It seems like the issue is that even when the above options are set, it still tries to use TryGetPersistentInstallationId which will always crash since files & directories cannot be created on the platform.

The error log that appears in the console when it generates a dump:
```
Error: nn::fs::CreateDirectory() failed because the operation is unsupported.
Search '2002-6364' on the following Error Code List for more information.
https://developer.nintendo.com/html/online-docs/g1kr9vj6-ja/Packages/Docs/Lists/ErrorCodeList/NintendoSwitch-ErrorCodeList-ja/2002.html
https://developer.nintendo.com/html/online-docs/g1kr9vj6-en/Packages/Docs/lists/NintendoSwitch-ErrorCodeList/2002.html

------------
AbortObserver called: Failed: nnFsResult
  Module: 2
  Description: 6364
  InnerValue: 0x0031b802
Condition: nn::Result::IsSuccess()
Function: CreateDirectory in D:\home\teamcity\work\sdk\Programs\Chris\Sources\Libraries\fs\fsa\fs_UserFileSystem.cpp:58
```

### Any logs or screenshots
[SymbolizedSwitchCrash.txt](https://github.com/user-attachments/files/16982569/SymbolizedSwitchCrash.txt)

bitsandfoxes:
Sorry to see you run into issues @abond-sd.
This should work out-of-the-box. We're handling platform-restrictions for `unknown` platforms (i.e. Switch) here:
https://github.com/getsentry/sentry-unity/blob/f4e129b68bfac22cb9bdafdc104b28dd67797f28/src/Sentry.Unity/ScriptableSentryUnityOptions.cs#L238

[...]

Issue #1779, opened 2024-08-28
WebGL freeze during SDK closing


ziggyeye:
https://github.com/getsentry/sentry-unity/issues/1761 is still happening in version 2.1.4
WebGL freeze during SDK closing

bitsandfoxes:
Would you be able to provide us with a minimal repro? I'd love to fix this.
It works for me with our sample, so there's something else going on.

ziggyeye:
That's strange. We haven't had the bandwidth to create a sample yet but it's definitely still freezing up like before.
I'd be curious if it's happening with others using the SDK.

bitsandfoxes:
I'll give this another try.

ziggyeye:
[web.zip](https://github.com/user-attachments/files/17004069/web.zip)
[UnityWebglSentryTestAssets.zip](https://github.com/user-attachments/files/17004071/UnityWebglSentryTestAssets.zip)

Attached are two zips with a a sample project that replicates the issue. SentryTestAssets.zip is the Assets folder of the test project. It's just an image displayed in webgl. But the web template used has a LOAD button to load the unity app. When loaded, there will be a QUIT button to close the Unity app. Pressing QUIT you will notice the web page freezes. The console log shows "Quitting..." but it never completes.

web.zip is the built version of this project. You can run this on a localhost easily to test it.
Please expedite this issue as it is a launch blocker for us that is imminent.

bitsandfoxes:
Thanks a lot for the repro! It's hugely appreciated!

bitsandfoxes:
It looks like this is still related to deadlocking during shutdown of the SDK. We'll try to minimize the repro on the .NET SDK. Hopefully, this is just the Metrics acting up.

ziggyeye:
thanks for the update

bitsandfoxes:
I've made the necessary changes to the underlying .NET SDK and it works fine now in our samples.
I'd like to give it a go in your repro but the actual project you shared only contains an `Assets` directory with a whole bunch of `.png` and the Sentry options. Could you share a functional repro I can open up in the editor and build myself? I'd like to avoid releasing it and still have it not working for you.

ziggyeye:
[sentrytest copy.zip](https://github.com/user-attachments/files/17079869/sentrytest.copy.zip)

Try this

bitsandfoxes:
I can build this project but the template fails to load the Unity game. :(

ziggyeye:
[...]

PR #1814, opened 2024-09-26
fix: Disable file-write on unknown platforms

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/1804, https://github.com/getsentry/sentry-unity/issues/1779

Also reverts https://github.com/getsentry/sentry-unity/pull/1728

Tracing itself has nothing to do with file-write. Turns out that simply making the request requires some access to the file system 
https://github.com/getsentry/sentry-dotnet/blob/12c7d7ca046ab985b0267f43c5fd35f2caad108c/src/Sentry/Internal/Http/HttpTransport.cs#L42


bruno-garcia (approved):

bruno-garcia:
Seems like only changelog is missing now?

PR #1826, opened 2024-10-07
fix: Builder

bitsandfoxes:
#skip-changelog

### DIFF
--- a/test/Sentry.Unity.Tests/IntegrationTests.cs
+++ b/test/Sentry.Unity.Tests/IntegrationTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
 using NUnit.Framework;
 using Sentry.Unity.Tests.TestBehaviours;
 using UnityEditor;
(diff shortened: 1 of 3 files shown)


test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -274,6 +275,8 @@ public IEnumerator DebugLogException_OnMainThread_IsCapturedAndIsMainThreadIsTru
     [UnityTest]
     public IEnumerator DebugLogException_InTask_IsCapturedAndIsMainThreadIsFalse()
     {
+        LogAssert.Expect(LogType.Exception, new Regex("Exception: .* Test Event"));
```
bitsandfoxes, line 278: Fixed a flaky test while at it.

bruno-garcia (approved):
This removes the integration tests on the built UPM package, right? please add a ticket to bring it back so we don't forget

bitsandfoxes:
> This removes the integration tests on the built UPM package, right? please add a ticket to bring it back so we don't forget 

The tests stay the same. I removed the duplicated logic of `is/is not main` between the `create project` and the `build project` steps.

test/Sentry.Unity.Tests/IntegrationTests.cs
bruno-garcia: Not sure this fixed it: 
image
https://github.com/getsentry/sentry-unity/actions/runs/11224732057/job/31202068887

PR #1828, opened 2024-10-07
fix: main CI

bitsandfoxes:
The issue is having to set `buildPlayerOptions.options = BuildOptions.AcceptExternalModificationsToPlayer;` only when exporting a Gradle project. Otherwise this breaks `main` in the `Build without SDK` step.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -316,7 +316,7 @@ jobs:
 
       - name: Build without Sentry SDK
         # This hasn't broken for many months, so disabling on PRs to speed up CI. And also to test a clean build with Sentry SDK included.
-        if: ${{ github.ref_name == 'main' }}
+        # if: ${{ github.ref_name == 'main' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -466,6 +461,11 @@ jobs:
       - name: Add Sentry to the project
         run: ./test/Scripts.Integration.Test/add-sentry.ps1 -UnityPath "${{ env.UNITY_PATH }}"
 
+      - name: Build without Sentry SDK
+        # This hasn't broken for many months, so disabling on PRs to speed up CI. And also to test a clean build with Sentry SDK included.
+        if: ${{ github.ref_name == 'main' }}
+        run: ./test/Scripts.Integration.Test/build-project.ps1 -UnityPath "${{ env.UNITY_PATH }}"
+
```
vaind, line 468: says "build without sentry sdk" after the sdk has been added to the project

vaind (approved):

PR #1850, opened 2024-10-16
fix: Missing contexts

bitsandfoxes:
We moved the collection of `MainThreadData` out of the `MonoBehaviour`'s Awake call in https://github.com/getsentry/sentry-unity/pull/1802
I'm actually not 100% why that worked in the first place. The Awake happens past the SceneLoad so the contexts should not have been available at that time either.

### Why is it broken

- The SDK initializes during `RuntimeInitializeLoadType.SubsystemRegistration`
https://github.com/getsentry/sentry-unity/blob/03af53dcd6462bc2525e8abfcbef7f4ad8060c8e/package-dev/Runtime/SentryInitialization.cs#L54
- The SDK registers the integrations on the options
- The UnityScopeIntegration gets registered and populates the contexts
https://github.com/getsentry/sentry-unity/blob/e7a45a96ab82e20394c6ad8711e3e0011594798c/src/Sentry.Unity/Integrations/UnityScopeIntegration.cs#L46-L58
- The `MainThreadData` has not been collected at this point because this happens during `BeforeSceneLoad`
https://github.com/getsentry/sentry-unity/blob/f1fe17a033b2a0861c937efd931dd19645908bda/src/Sentry.Unity/MainThreadData.cs#L75-L76
- The contexts are missing

### How to fix it

Instead of relying on the Unity `RuntimeInitialize` we can just collect the contexts right before initialization
https://github.com/getsentry/sentry-unity/blob/e7a45a96ab82e20394c6ad8711e3e0011594798c/src/Sentry.Unity/SentryUnitySDK.cs#L31
Everything is nice and explicit.
Updated our smoketest so contexts won't go missing anymore.

### DIFF
--- a/src/Sentry.Unity/SentryUnitySDK.cs
+++ b/src/Sentry.Unity/SentryUnitySDK.cs
@@ -28,6 +28,8 @@ internal class SentryUnitySdk
             return null;
         }
 
+        MainThreadData.CollectData(SentrySystemInfoAdapter.Instance);
+
         // On Standalone, we disable cache dir in case multiple app instances run over the same path.
         // Note: we cannot use a named Mutex, because Unit doesn't support it. Instead, we create a file with `FileShare.None`.
         // https://forum.unity.com/threads/unsupported-internal-call-for-il2cpp-mutex-createmutex_internal-named-mutexes-are-not-supported.387334/
--- a/src/Sentry.Unity/SentryUnitySDK.cs
+++ b/src/Sentry.Unity/SentryUnitySDK.cs
@@ -28,7 +28,7 @@ internal class SentryUnitySdk
             return null;
         }
 
-        MainThreadData.CollectData(SentrySystemInfoAdapter.Instance);
+        MainThreadData.CollectData();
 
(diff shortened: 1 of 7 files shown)


src/Sentry.Unity/SentryUnitySDK.cs
```diff
@@ -28,6 +28,8 @@ private SentryUnitySdk(SentryUnityOptions options)
             return null;
         }
 
+        MainThreadData.CollectData();
```
vaind, line 31: Do they sync to native when moved here?

bitsandfoxes: Yes. For our purposes the `MainThreadData` can be considered immutable and needs to be collected and then synched once. The sync to native happens after initialization
https://github.com/getsentry/sentry-unity/blob/f1fe17a033b2a0861c937efd931dd19645908bda/src/Sentry.Unity/SentryUnitySDK.cs#L70-L77

vaind (approved):

Issue #1871, opened 2024-10-30
Warning due to `SampleRate` passed on to Android as `int`


bitsandfoxes:
We're setting the `SampleRate` in the Android Manifest here
https://github.com/getsentry/sentry-unity/blob/4df953801eaf87d6de850c9756942dbf51637c4f/src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs#L422-L423

causing

```
Warn Bundle Key io.sentry.sample-rate expected Float but value was a java.lang.Integer.  The default value -1.0 was returned.
2024/10/30 18:00:55.358 21665 21665 Warn Bundle Attempt to cast generated internal exception:
2024/10/30 18:00:55.358 21665 21665 Warn Bundle java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Float
2024/10/30 18:00:55.358 21665 21665 Warn Bundle at android.os.BaseBundle.getFloat(BaseBundle.java:1364)
2024/10/30 18:00:55.358 21665 21665 Warn Bundle at android.os.Bundle.getFloat(Bundle.java:798)
2024/10/30 18:00:55.358 21665 21665 Warn Bundle at io.sentry.android.core.e1.f(ManifestMetadataReader.java:3)
```

It looks like we might be running into an edge case of `sampleRate.ToString()` being written as "1" instead of "1.0"?

PR #1872, opened 2024-10-30
fix: Write `SampleRate` as `float` to manifest

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/1871


vaind (approved):
Looks like this issue https://github.com/getsentry/sentry-java/issues/3603
which has already been addressed and should come out in the next sentry-java release. 

I don't see why merging this would hurt though.

PR #1875, opened 2024-10-31
fix: link sentry native against static windows CRT

### LINKED ISSUE #1440: System.DllNotFoundException caused by dynamically linked CRT
### Environment

How do you use Sentry?
sentry.io through Unity

Which version of the SDK?
1.5.1

Unity Package Manager

Which version of Unity?
2022.3.8f1

Player, Windows 10 IL2CPP Production build

### Steps to Reproduce

Difficult to reproduce, because this is happening in the wild, but the issue seems to be because you're compiling sentry.dll with the dynamic CRT linked in. This means we're required to ensure that the user has the correct CRT runtimes installed too.

Would be great if it was possible to statically link the CRT instead, so that we don't need additional runtimes installed:

Using depends you can see the currently dynamically linked dlls (see below)

On systems that don't have this installed, but run our IL2CPP Unity game just fine, we experience the System.DllNotFoundException

### Expected Result

No exception

### Actual Result

System.DllNotFoundException
Unable to load DLL 'sentry'. Tried the load the following dynamic libraries: Unable to load dynamic library 'sentry' because of 'Failed to open the requested dynamic library (0x06000000) - The specified module could not be found.  (WinError:0000007e)

### Any logs or screenshots

image

vaind:
Fixes #1440

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Fixed an issue where the SDK would write the `SampleRate` as an `int` instead of a `float` to the Android Manifest, causing issues during the Android SDK's initialization ([#1872](https://github.com/getsentry/sentry-unity/pull/1872))
 - The SDK no longer calls into `Application.persistentDataPath` on unknown platforms. This prevents crashes during startup on platforms with restricted disk access like the Nintendo Switch ([#1870](https://github.com/getsentry/sentry-unity/pull/1870))
-- Link sentry-native DLL against static Windows CRT ([#1875](https://github.com/getsentry/sentry-unity/pull/1875))
+- `sentry-native` DLL is not statically linked against Windows CRT. This allows the SDK to provide native support on some Windows systems that do not have CRT installed locally ([#1875](https://github.com/getsentry/sentry-unity/pull/1875))
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fixed an issue where the SDK would write the `SampleRate` as an `int` instead of a `float` to the Android Manifest, causing issues during the Android SDK's initialization ([#1872](https://github.com/getsentry/sentry-unity/pull/1872))
 - The SDK no longer calls into `Application.persistentDataPath` on unknown platforms. This prevents crashes during startup on platforms with restricted disk access like the Nintendo Switch ([#1870](https://github.com/getsentry/sentry-unity/pull/1870))
+- Link sentry-native DLL against static Windows CRT ([#1875](https://github.com/getsentry/sentry-unity/pull/1875))
```
bitsandfoxes, line 9:
```suggestion
- `sentry-native` DLL is not statically linked against Windows CRT. This allows the SDK to provide native support on some Windows systems that do not have CRT installed locally ([#1875](https://github.com/getsentry/sentry-unity/pull/1875))
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fixed an issue where the SDK would write the `SampleRate` as an `int` instead of a `float` to the Android Manifest, causing issues during the Android SDK's initialization ([#1872](https://github.com/getsentry/sentry-unity/pull/1872))
 - The SDK no longer calls into `Application.persistentDataPath` on unknown platforms. This prevents crashes during startup on platforms with restricted disk access like the Nintendo Switch ([#1870](https://github.com/getsentry/sentry-unity/pull/1870))
+- `sentry-native` DLL is not statically linked against Windows CRT. This allows the SDK to provide native support on some Windows systems that do not have CRT installed locally ([#1875](https://github.com/getsentry/sentry-unity/pull/1875))
```
bitsandfoxes, line 9:
```suggestion
- `sentry-native` DLL is now statically linked against Windows CRT. This allows the SDK to provide native support on some Windows systems that do not have CRT installed locally ([#1875](https://github.com/getsentry/sentry-unity/pull/1875))
```

bitsandfoxes (approved):

PR #1880, opened 2024-11-04
CI: use REST API in the unity updater

vaind:
The original code relied on parsing a releases web page, which became flaky now that it is dynamic (loads actual release data after load has completed). I've found out there's an API so I'm using that instead. See https://services.docs.unity.com/release/v1/

also, I've updated the action dependency because the CI complains about node depreciation

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -3,8 +3,8 @@ name: Update Dependencies
 on:
   # Run every day on multiple occasions, with each job running at a different time to limit license-acquisition issues.
   schedule:
-    - cron: "0 2 * * *"
-    - cron: "0 3 * * *"
+    - cron: '0 2 * * *'
+    - cron: '0 3 * * *'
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -181,3 +181,6 @@ csharp_new_line_before_finally = true
 csharp_new_line_before_members_in_object_initializers = true
 csharp_new_line_before_members_in_anonymous_types = true
 csharp_new_line_between_query_expression_clauses = true
+
+[*.yml]
+quote_type = single
```
vaind, line 186: for some reason VSCode kept replacing single-quotes with double-quotes in the yaml despite all the local setting for using singles. This helped: https://stackoverflow.com/questions/48864985/vscode-single-to-double-quote-automatic-replace#comment105479981_48866406

vaind: I hope it's alright to keep it here

bitsandfoxes (approved):
🔥

Issue #1621, opened 2024-04-15
Simplify `BuildTime` and `Runtime` Options


bitsandfoxes:
## Problem

Because the native SDKs on mobile get initialized even before the game even starts, the SDK is providing two `configure` hooks to programmatically configure the options, one for runtime and one for build-time. This is creating a lot of confusion as to which options apply to what and when. 
Examples:
-  _Even tho I set the default tags right during build-time options they do not get applied during runtime._
- I've set the sample rate in the runtime options but it does not seem to affect events coming from the native SDK
- 

### Serialized Options and programmatic configuration

We needed a place to keep all the options we surface in the editor config window. Instead of a config file the SDK makes use of the built-in `ScriptableObject` functionality. To allow for programmatic configuration the SDK provides two additional scriptable objects users can inherit and overwrite their `Configure` callback

- https://github.com/getsentry/sentry-unity/blob/2085934e83c2f0f7242a6fd64bc21ec1a675a053/src/Sentry.Unity/SentryRuntimeOptionsConfiguration.cs#L13
- https://github.com/getsentry/sentry-unity/blob/2085934e83c2f0f7242a6fd64bc21ec1a675a053/src/Sentry.Unity/SentryBuildTimeOptionsConfiguration.cs#L13

### Initialization

Before initialization we're "loading" the `scriptable options` and create `SentryUnityOptions` from there

https://github.com/getsentry/sentry-unity/blob/2085934e83c2f0f7242a6fd64bc21ec1a675a053/src/Sentry.Unity/ScriptableSentryUnityOptions.cs#L121-L130

and we're invoking the `RuntimeConfiguration` during the option creation.

https://github.com/getsentry/sentry-unity/blob/2085934e83c2f0f7242a6fd64bc21ec1a675a053/src/Sentry.Unity/ScriptableSentryUnityOptions.cs#L219

### Build Time Options

In comparison, the build time options get invoked in a bunch of different places during the build process. Any options set here do **not** apply to the C# layer at runtime.

- https://github.com/getsentry/sentry-unity/blob/2085934e83c2f0f7242a6fd64bc21ec1a675a053/src/Sentry.Unity.Editor/Android/PostBuildCheck.cs#L17
- https://github.com/getsentry/sentry-unity/blob/2085934e83c2f0f7242a6fd64bc21ec1a675a053/src/Sentry.Unity.Editor.iOS/BuildPostProcess.cs#L22
[...]

PR #1888, opened 2024-11-06
feat: Add `OptionConfiguration` to replace `Runtime/BuildTime` 

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/1621

Based on the issue description: The aim is to have users no longer need to know or care about `Runtime` and `BuildTime` differences. They write their code in once place, we use the same options at any time.

Platform specifics can be controlled via the Unity precompile directives. We include a sample of those with the generated template:

```
using Sentry;
using Sentry.Unity;

public class SentryConfiguration : SentryOptionsConfiguration
{
    public override void Configure(SentryUnityOptions options)
    {
        // Here you can programmatically modify the Sentry option properties used for the SDK's initialization

#if UNITY_ANDROID || UNITY_IOS || UNITY_STANDALONE_OSX
        // NOTE!
        // On Android, iOS, and macOS, ALL options configured here will be "baked" into the exported project
        // during the build process.
        // Changes to the options at runtime will not affect the native SDKs (Java, C/C++, Objective-C)
        // and only apply to the C# layer.

        // Examples:

        // Works as expected and will disable all debug logging of the native SDKs
        // options.Debug = false;

        // Will NOT work as expected as this will need to get validated at runtime
        // options.Debug = SystemInfo.deviceName.Contains("Pixel");
#endif
    }
}
```

The Runtime/BuildTime config got deprecated but will still be called last as keep backwards compatibility. I've updated the config window with a warning label that shows as long as there are either Runtime or BuildTime configs.
ScriptableConfig


samples/unity-of-bugs/Assets/Scripts/SentryOptionConfiguration.cs
```diff
@@ -0,0 +1,15 @@
...
+
+public class SentryOptionConfiguration : SentryOptionsConfiguration
+{
+    public override void Configure(SentryUnityOptions options)
+    {
+        // Here you can programmatically modify the Sentry option properties used for the SDK's initialization
+
+        options.Environment = "dev";
+
+        options.NativeOverrides.Environment = "native_dev";
```
bruno-garcia, line 12: Is this a real use case though?
Shouldn't we just take the `dev` above and automatically override native?

samples/unity-of-bugs/Assets/Scripts/SentryOptionConfiguration.cs
```diff
@@ -6,10 +6,19 @@ public class SentryOptionConfiguration : SentryOptionsConfiguration
...
     {
-#if UNITY_ANDROID
-        options.SampleRate = 0.1f;
-#elif UNITY_IOS
-        options.SampleRate = 1.0f;
+        // Here you can programmatically modify the Sentry option properties used for the SDK's initialization
+
+#if UNITY_ANDROID || UNITY_IOS
+        // Important!
+        // Android and iOS options are getting validated and set at build time. Changes to them while the game
```
bruno-garcia, line 13: "while the game .. " missing a bit at the end?

samples/unity-of-bugs/Assets/Scripts/SentryOptionConfiguration.cs
```diff
@@ -6,10 +6,19 @@ public class SentryOptionConfiguration : SentryOptionsConfiguration
 {
     public override void Configure(SentryUnityOptions options)
     {
-#if UNITY_ANDROID
-        options.SampleRate = 0.1f;
-#elif UNITY_IOS
-        options.SampleRate = 1.0f;
+        // Here you can programmatically modify the Sentry option properties used for the SDK's initialization
+
+#if UNITY_ANDROID || UNITY_IOS
```
bruno-garcia, line 11: Should we just make the C# layer initialize the native layer by default? 

And make this "Native inits first" thing be opt-in? It seems still hard for folks to understand what's going on, even with the comments below. (I don't expect a beginner to get it)

src/Sentry.Unity.Editor/ConfigurationWindow/OptionsConfigurationTab.cs
```diff
@@ -17,23 +17,10 @@ public static void Display(ScriptableSentryUnityOptions options)
...
-            "\n" +
-            "Use the 'Build Time Configuration Script' in case you need to change build-time behavior, " +
-            "e.g. specify custom Sentry-CLI options or change settings for native SDKs that start before the " +
-            "managed layer does (such as Android, iOS, macOS).",
-            MessageType.Info);
-
-        EditorGUILayout.Space();
-
-        EditorGUILayout.HelpBox("Deprecated and will be removed in a future version.\nPlease use the 'Option Config Script' below instead.", MessageType.Warning);
+        EditorGUILayout.HelpBox("The Runtime/BuildTime scriptable objects have been deprecated and will be removed in a future version." +
```
bruno-garcia, line 20: Should we link to someplace (docs?) that explains in more details what's going on?

If I see a deprecation I always worry about what that means to me: "OK so it won't work anymore what I did before?"

Could we show this message only if it's an upgrade, where they have the old, 2 files thingy. But not for new installs where there are no files yet created? (avoid confusion to new users and don't scare old users?)

src/Sentry.Unity.Editor/SentryScriptableObject.cs
```diff
@@ -37,25 +37,20 @@ internal static (SentryUnityOptions?, SentryCliOptions?) ConfiguredBuildTimeOpti
         if (scriptableOptions is not null)
         {
             options = scriptableOptions.ToSentryUnityOptions(isBuilding: true, unityInfo: null);
+
+            // TODO: Move this into `Load` once we remove Runtime- and BuildTimeConfig
+            // We're calling `Configure` here and not in `Load` so the new Config does not overwrite the BuildTimeConfig
```
bruno-garcia, line 42: helpful comment, thanks

src/Sentry.Unity/SentryOptionsConfiguration.cs
```diff
@@ -14,13 +14,26 @@ public class {{ScriptName}} : SentryOptionsConfiguration
...
+                // build time. Any runtime changes to these options will not take effect.
+
+                // Examples:
+
+                // Works as expected and will disable all debug logging of the native SDKs
+                // options.Debug = false;
+
+                // Will NOT work as expected as this will need to get validated at runtime
+                // options.Debug = SystemInfo.deviceName.Contains("Pixel");
+        #endif
```
bruno-garcia, line 30: The "will NOT WORK" note doesn't expand on what to do in that case. 

Not arguying this isnt' a step forward, just saying I fear it's still confusing

```suggestion
#if UNITY_ANDROID || UNITY_IOS
                // Important!
                // On Android and iOS, ALL options configured here will be "baked" into the exported project during
                // build time. Any runtime changes to these options will not take effect.

                // Examples:

                // Works as expected and will disable all debug logging of the native SDKs
                // options.Debug = false;

                // Will NOT work as expected as this will need to get validated at runtime
                // options.Debug = SystemInfo.deviceName.Contains("Pixel");
#endif
```

test/Scripts.Integration.Test/Editor/BuildTimeOptions.cs
```diff
@@ -11,33 +12,10 @@ public override void Configure(SentryUnityOptions options, SentryCliOptions cliO
...
-        options.PerformanceAutoInstrumentationEnabled = true;
-
-        cliOptions.UploadSymbols = !string.IsNullOrEmpty(cliOptions.UrlOverride);
-        cliOptions.UploadSources = cliOptions.UploadSymbols;
-        cliOptions.Organization = "sentry-sdks";
-        cliOptions.Project = "sentry-unity";
-        cliOptions.Auth = "dummy-token";
-
+        // Make sure the config is still called and the DSN has been set
+        Assert.IsFalse(string.IsNullOrWhiteSpace(options.Dsn));
```
bruno-garcia, line 16: seems like a downgrade where we don't assert the actual value anymore, just that it's not set.

If we somehow end up passing: "Disable SDK" which is `""`, tests would pass.

Also, if test fails we'd get: "Expected true but got false". So ideally use a better assertion API like "Assert.Same(A,B)" or whatever this lib has (NUnit is it?)

test/Scripts.Integration.Test/Scripts/CliConfiguration.cs
```diff
@@ -0,0 +1,23 @@
...
+using UnityEngine.Assertions;
+
+public class CliConfiguration : SentryCliOptionsConfiguration
+{
+    public override void Configure(SentryCliOptions cliOptions)
+    {
+        Debug.Log("Sentry: CliConfiguration::Configure() called");
+
+        // Assert the deprecated options have been set and are getting overwritten
+        // Assert.AreEqual("Old configure got called", cliOptions.Auth);
```
bruno-garcia, line 13: Not asserting this? should we?

It's important to verify we're not breaking people who are upgrading btw. 

Without snapshot testing we're just going from a blank state, so we won't validate that

test/Scripts.Integration.Test/Scripts/RuntimeOptions.cs
```diff
@@ -9,36 +10,8 @@ public override void Configure(SentryUnityOptions options)
...
-            {
-                return options.CrashedLastRun() ? 1 : 0;
-            }
-            return -2;
-        };
-
-        // If an ANR triggers while the smoke test runs, the test would fail because we expect exact order of events.
-        options.DisableAnrIntegration();
+        // Make sure the config is still called and the DSN has been set
+        Assert.IsFalse(string.IsNullOrWhiteSpace(options.Dsn));
```
bruno-garcia, line 14: oh actually now we get an assert here, while before we just wrote some setting to assert later?

test/Scripts.Integration.Test/add-sentry.ps1
```diff
@@ -17,4 +17,4 @@ RunUnityAndExpect $UnityPath "AddSentryPackage" "Sentry Package Installation: SU
 Write-Host "Copying Integration Test Files"
 New-Item -Path "$(GetNewProjectAssetsPath)" -Name "Scripts" -ItemType "directory"
 Copy-Item -Recurse "$IntegrationScriptsPath/Scripts/*" -Destination "$(GetNewProjectAssetsPath)/Scripts/"
-Copy-Item -Recurse "$IntegrationScriptsPath/Editor/BuildTimeOptions.cs" -Destination "$(GetNewProjectAssetsPath)/Editor/"
+Copy-Item -Recurse "$IntegrationScriptsPath/Editor/BuildTimeOptions.cs" -Destination "$(GetNewProjectAssetsPath)/Editor/"
```
bruno-garcia, line 20:
```suggestion
Copy-Item -Recurse "$IntegrationScriptsPath/Editor/BuildTimeOptions.cs" -Destination "$(GetNewProjectAssetsPath)/Editor/"

```

samples/unity-of-bugs/Assets/Scripts/SentryOptionConfiguration.cs
vaind: It would be simpler but we would lose on the ability to catch native exceptions happening at startup. Is that worth it?

vaind (commented):
> The Runtime/BuildTime config got deprecated but will still be called last as keep backwards compatibility. I've updated the config window with a warning label.
> Screenshot 2024-11-15 at 11 18 18

What do you think about only showing the new option on the UI by default and only show the original options (i.e. as in the screenshot), when any of the original options are actually set in the existing config?

samples/unity-of-bugs/Assets/Scripts/SentryOptionConfiguration.cs
bitsandfoxes: I can see a point for: If you've got plugins that might crash the app before the Unity game starts you can opt-in the native first init and still have full coverage. 
If you don't then are those native crashes/exceptions be even actionable for you?

bitsandfoxes:
> What do you think about only showing the new option on the UI by default and only show the original options (i.e. as in the screenshot), when any of the original options are actually set in the existing config?

I thought about that. But what if a user messes up their setup, removes the reference and wants to restore what they had. Then we've built a "smart" UI solution that breaks and blocks them from fixing it themselves. I guess version control could fix that but still..

src/Sentry.Unity.Editor/ConfigurationWindow/OptionsConfigurationTab.cs
bitsandfoxes: > Could we show this message only if it's an upgrade, where they have the old, 2 files thingy. But not for new installs where there are no files yet created? (avoid confusion to new users and don't scare old users?)

Great point. Made it so.

test/Scripts.Integration.Test/Editor/BuildTimeOptions.cs
bitsandfoxes: Those are the old config callbacks for the smoke test. All I want to check there is that those are getting called.
1. Invoke the "new way" to programmatic Configure and set the options for the smoke tests
2. Make sure the "old way" is still getting called after the new one. So we stay compatible and won't break users.

test/Scripts.Integration.Test/Scripts/CliConfiguration.cs
bitsandfoxes: Good catch. This is dead code and actually the other way around. First New than Old.

test/Scripts.Integration.Test/Scripts/RuntimeOptions.cs
bitsandfoxes: This is the same as the `BuildTimeOptions` above. The options used for smoke testing are consolidated in the new `OptionConfiguration` callback and here we're making sure those options are still getting set and the old way is run last.

src/Sentry.Unity/SentryOptionsConfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+                *  │  The configure call made │
+                *  │  for this part ran on    │
+                *  │  your build-machine      │
+                *  └──────────────────────────┘
+                *               │
+                *               │
+                *               ▼
+                *  ┌──────────────────────────┐
+                *  │         Native SDK       │
+                *  │    Android/iOS/macOS)    │
```
bruno-garcia, line 60:
```suggestion
                *  │    (Android/iOS/macOS)   │
```

samples/unity-of-bugs/Assets/Scripts/SentryOptionConfiguration.cs
vaind: > I can see a point for: If you've got plugins that might crash the app before the Unity game starts you can opt-in the native first init and still have full coverage. 

Wouldn't that mean we still need to keep the buildtime native configuration generator (which I understand Bruno was suggesting we could remove)?

> If you don't then are those native crashes/exceptions be even actionable for you?

Even the information that there are startup crashes is valuable IMO. 

---- 
Overall, I wouldn't change this - it does provide value (startup crashes) even though it does make things a little more complicated IF users need to adapt these settings programmatically

bruno-garcia (approved):

src/Sentry.Unity/SentryOptionsConfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+
+                // Works as expected and will enable all debug logging
+                // options.Debug = true;
+
+                // Will NOT work as expected. 
+                // This will run twice.
+                //    1. Once during the build, being baked into the native SDKs
+                //    2. And a second time every time when the game starts
+                // options.Release = ComputeVersion();                
+        #endif
```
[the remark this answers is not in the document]
bitsandfoxes: These are not actual precompile directives but part of the template. Their indentation is correct when written to file.

Issue #1864, opened 2024-10-24
sentry_reinstall_backend assembly exception


flashup:
### Environment

How do you use Sentry?
Sentry self-hosted (22.10.0)

Which version of the SDK?
2.2.2

How did you install the package?
Git-URL

Which version of Unity?
2022.3.41f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Windows

### Steps to Reproduce

1. The error happens on application startup.

#### Callstack
```
System.EntryPointNotFoundException: sentry_reinstall_backend assembly:<unknown assembly> type:<unknown type> member:(null)
  Module "Sentry.Unity.Native.SentryNativeBridge", in ReinstallBackend
  Module "Sentry.Unity.Native.SentryNative", in ReinstallBackend
  Module "Sentry.Unity.SentryInitialization", in ReinstallBackend
```

#### Breadcrumbs
![Image](https://github.com/user-attachments/assets/1b502ed5-0646-4ebc-a28e-d79481293885)

bruno-garcia:
Looks like `sentry.dll` is missing? That P/Invoke call is trying to call a function written in C on our native SDK.
This one:

https://github.com/getsentry/sentry-native/blob/66e5aae0ba472f80086cfa32466f054cffd5ce3d/src/sentry_core.c#L277

Looks like it can't find the DLL

bruno-garcia:
> In the output dir I see the game.exe and the library game_Data/Managed/Sentry.dll

That's the managed (.NET) Sentry dll. We also provide a native DLL (`sentry.dll`). 

Seems to be a known issue:
* https://github.com/getsentry/sentry-unity/issues/1108

Native support for desktop targets IL2CPP so there might be a bug on Linux.


Does this crash the game or something? Or it's just an exception in the terminal/breadcrumb?
If the latter it might be a matter of skipping this call to`sentry_reinstall_backend`

Right now it load automatically:

https://github.com/getsentry/sentry-unity/blob/8c0967fedc553a2bd0f0ab95ece8b21b6c47e9d2/package-dev/Runtime/SentryInitialization.cs#L122-L131


Also relates to:
* https://github.com/getsentry/sentry-unity/issues/782

flashup:
> That's the managed (.NET) Sentry dll. We also provide a native DLL (`sentry.dll`).

Ok, I also have this lib: `game_Data/Plugins/x86_64/sentry.dll`


> Does this crash the game or something?

[...]

PR #1895, opened 2024-11-10
Suppress EntryPointNotFoundException on sentry native

bruno-garcia:
Resolves: https://github.com/getsentry/sentry-unity/issues/1864

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -118,25 +118,6 @@ namespace Sentry.Unity
 #endif
             }
         }
-
-#if SENTRY_NATIVE
-        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
-        public static void ReinstallBackend()
-        {
-            try
-            {
-                // At this point Unity has taken the signal handler and will not invoke our handler. So we register our
-                // backend once more to make sure user-defined data is available in the crash report and the SDK is able
-                // to capture the crash.
-                SentryNative.ReinstallBackend();
-            }
-            catch (EntryPointNotFoundException e)
-            {
-                // See: https://github.com/getsentry/sentry-unity/issues/1864
-                // TODO: if option.Debug: true, debug log this
-            }
-        }
-#endif
     }
 
     public class SentryUnityInfo : ISentryUnityInfo
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -123,10 +123,18 @@ namespace Sentry.Unity
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         public static void ReinstallBackend()
         {
-            // At this point Unity has taken the signal handler and will not invoke our handler. So we register our
-            // backend once more to make sure user-defined data is available in the crash report and the SDK is able
-            // to capture the crash.
-            SentryNative.ReinstallBackend();
+            try
+            {
+                // At this point Unity has taken the signal handler and will not invoke our handler. So we register our
+                // backend once more to make sure user-defined data is available in the crash report and the SDK is able
+                // to capture the crash.
+                SentryNative.ReinstallBackend();
+            }
+            catch (EntryPointNotFoundException e)
+            {
+                // See: https://github.com/getsentry/sentry-unity/issues/1864
+                // TODO: if option.Debug: true, debug log this
+            }
         }
 #endif
     }
--- a/src/Sentry.Unity.Native/SentryNative.cs
+++ b/src/Sentry.Unity.Native/SentryNative.cs
@@ -14,7 +14,7 @@ public static class SentryNative
 {
     private static readonly Dictionary<string, bool> PerDirectoryCrashInfo = new();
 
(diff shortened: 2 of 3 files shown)


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -123,10 +123,18 @@ public static void Init()
...
+            {
+                // At this point Unity has taken the signal handler and will not invoke our handler. So we register our
+                // backend once more to make sure user-defined data is available in the crash report and the SDK is able
+                // to capture the crash.
+                SentryNative.ReinstallBackend();
+            }
+            catch (EntryPointNotFoundException e)
+            {
+                // See: https://github.com/getsentry/sentry-unity/issues/1864
+                // TODO: if option.Debug: true, debug log this
```
bruno-garcia, line 136: @bitsandfoxes is there a static logger I can try to grab to do this?

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -80,7 +83,26 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
...
+        try
+        {
+            // At this point Unity has taken the signal handler and will not invoke our handler. So we register our
+            // backend once more to make sure user-defined data is available in the crash report and the SDK is able
+            // to capture the crash.
+            ReinstallBackendFunction?.Invoke();
+        }
+        catch (EntryPointNotFoundException e)
+        {
+            SentrySdk.CaptureException(new Exception("The SDK failed to reinstall the backend and won't capture native errors.", e));
```
bitsandfoxes, line 105: I'm not a huge fan of capturing exceptions the SDK throws itself but this is during initialization, making sure that the lack of observability of native crashes is known.

bruno-garcia: I'm not sure this is the best route. What would the user do with this exception?
Also, they'd be paying for these events.

Debug log is the way to go imo

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -80,7 +83,26 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
             }
         }
         options.CrashedLastRun = () => crashedLastRun;
+
+        // The backend will only be reinstalled if the native SDK has been initialized successfully.
+        // We're handing this off to be called in `BeforeSceneLoad` instead of `SubsystemRegistration` as it's too soon
+        // and the SignalHandler would still get overwritten.
+        ReinstallBackendFunction = SentryNativeBridge.ReinstallBackend;
```
bruno-garcia, line 90: This seems hacky: do we change the method in any way? We're just signaling from this part of the case to another. I think a bool is more clear and has other benefits (as in not having the action)

src/Sentry.Unity.Native/SentryNative.cs
bitsandfoxes: How would logging help? I think we have to make sure users can be certain the SDK is functioning correctly.
The scenario I've got in my head and would like to cover:
1. User ships their game but the native SDK fails to reinstall.
2. Crash free rate is 100%
3. Players still report crashes but the user has no idea why Sentry is not reporting any
4. What now? Enable `debug` and ask for logs?

We're doing the same thing to cover the native SDK's `Init` here
https://github.com/getsentry/sentry-unity/blob/8c0967fedc553a2bd0f0ab95ece8b21b6c47e9d2/package-dev/Runtime/SentryInitialization.cs#L77-L92

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -13,6 +14,8 @@ public static class SentryNative
 {
     private static readonly Dictionary<string, bool> PerDirectoryCrashInfo = new();
 
+    private static bool ShouldReinstallBackend;
```
bruno-garcia, line 17:
```suggestion
    private static bool _shouldReinstallBackend;
```

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -80,7 +83,31 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
             }
         }
         options.CrashedLastRun = () => crashedLastRun;
+
+        ShouldReinstallBackend = true;
```
bruno-garcia, line 87:
```suggestion
        _shouldReinstallBackend = true;
```

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -80,7 +83,31 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
...
     }
 
-    public static void ReinstallBackend() => SentryNativeBridge.ReinstallBackend();
+    // We're calling this in `BeforeSceneLoad` instead of `SubsystemRegistration` as it's too soon and the
+    // SignalHandler would still get overwritten.
+    [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
+    private static void ReinstallBackend()
+    {
+        // The backend should only be reinstalled if the native SDK has been initialized successfully.
+        if (!ShouldReinstallBackend)
```
bruno-garcia, line 96:
```suggestion
        if (!_shouldReinstallBackend)
```

Issue #1727, opened 2024-07-16
Using Unity as a Library does not report crashes/fatals


Esildor:
### Environment

How do you use Sentry?
Sentry SaaS

Which version of the SDK?
2.1.0

How did you install the package?
Git-URL

Which version of Unity?
6000.0.4f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Android

### Steps to Reproduce
1. Download and open the [Unity as a Library (UaaL) example](https://github.com/Unity-Technologies/uaal-example).
2. Install Sentry as a package.
3. Run the setup wizard to ensure the Sentry plugin is activated.
4. Ensure that **Android Native Support** is enabled.
5. Add this script [this script](https://github.com/user-attachments/files/16240283/CrashSimulator.cs.txt) to the project. This will add buttons to simulate various crashes via `UnityEngine.Diagnostics.Utils.ForceCrash()`. To the top left of the screen.
6. Continue setting up the UaaL project per the instructions in [Unity's Github docs](https://github.com/Unity-Technologies/uaal-example/blob/master/docs/android.md)
7. Build and run on a device
8. Press any of the `ForcedCrashCategory` buttons, for example: _Engine Crash - FatalError_, _Engine Crash - AccessViolation_, etc.

### Expected Result
Unity engine-level crashes/fatal issues are reported to Sentry.

### Actual Result
No fatal crash is reported. It's also worth noting: Performing these steps in a _non-UaaL_ project reports fatal level crashes to Sentry as expected, see the image below.

### Any logs or screenshots
![image](https://github.com/user-attachments/assets/36bc1ae1-405f-4965-8ecd-fec266f61861)

bitsandfoxes:
Thanks @Esildor for reaching out! Tbh I've never tried the UaaL sample. I'll get back to you after giving this a try!

bitsandfoxes:
This has not been forgotten. I'm working my way through the guide.

Esildor:
Thanks for the update 💖

bitsandfoxes:
We're working on, well, making that work. There are a couple of questions that need to be answered first:
- How (if) the Unity SDK should sync with the surrounding native SDK
- Do those SDK share a DSN? Should the Unity SDK just send events one way and not care?
- Should the scope be de-synced during the games shutdown. I.e. The native app is some sort of launcher and there are a bunch of Unity games

bitsandfoxes:
[...]

PR #1897, opened 2024-11-11
fix: Only close the native SDKs if native support has been enabled

bitsandfoxes:
Relates to: https://github.com/getsentry/sentry-unity/issues/1727

When having the game run as a library, closing the game closes the SDK. This causes the surrounding app's native SDK to close as well.
Because the SDK can't determine the setup it is used in, it should only close the native SDKs if it is sure it was the one enabling it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- The SDK no longer closes the underlying native SDK during the games shutdown if native support has not been enabled. This enables the SDK to support and capture errors in case of building the game as a library. ([#1897](https://github.com/getsentry/sentry-unity/pull/1897))
+
 ### Dependencies
 
 - Bump .NET SDK from v4.12.1 to v4.13.0 ([#1879](https://github.com/getsentry/sentry-unity/pull/1879), [#1885](https://github.com/getsentry/sentry-unity/pull/1885))
--- a/src/Sentry.Unity.Android/SentryNativeAndroid.cs
+++ b/src/Sentry.Unity.Android/SentryNativeAndroid.cs
@@ -1,5 +1,7 @@
 using System;
 using Sentry.Extensibility;
+using Sentry.Unity.Integrations;
+using UnityEngine;
 using UnityEngine.Analytics;
 
 namespace Sentry.Unity.Android;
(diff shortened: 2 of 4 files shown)


src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -1,5 +1,7 @@
 using System;
 using Sentry.Extensibility;
+using Sentry.Unity.Integrations;
+using UnityEngine;
```
bruno-garcia, line 4: Are we using this?
```suggestion
```

src/Sentry.Unity.iOS/SentryNativeCocoa.cs
```diff
@@ -81,9 +81,17 @@ internal static void Configure(SentryUnityOptions options, ISentryUnityInfo sent
...
+
+    internal static void Close(SentryUnityOptions options, ISentryUnityInfo sentryUnityInfo, RuntimePlatform platform)
     {
-        logger?.LogDebug("Closing the sentry-cocoa SDK");
+        if (!sentryUnityInfo.IsNativeSupportEnabled(options, platform))
+        {
+            return;
+        }
+
+        options.DiagnosticLogger?.LogDebug("Closing the sentry-cocoa SDK");
```
bruno-garcia, line 94:
```suggestion
            options.DiagnosticLogger?.LogDebug("Native is not enable so bypassing native close");
            return;
        }

        options.DiagnosticLogger?.LogDebug("Closing the sentry-cocoa SDK");
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- The SDK no longer closes the underlying native SDK during the games shutdown if native support has not been enabled. This enables the SDK to support and capture errors in case of building the game as a library. ([#1897](https://github.com/getsentry/sentry-unity/pull/1897))
```
bruno-garcia, line 7:
```suggestion
- The SDK no longer closes the underlying native SDK during the games shutdown if native support has not been enabled. This allows the SDK to support and capture errors in case of building the game as a library on a mobile (Android or iOS) game. ([#1897](https://github.com/getsentry/sentry-unity/pull/1897))
```

bruno-garcia (approved):
some nitpicks and suggestions but lgtm

src/Sentry.Unity.Android/SentryNativeAndroid.cs
bitsandfoxes: We need it for the `RuntimePlatform`

PR #1898, opened 2024-11-12
fix: Suppress and log known exceptions during native initialization

bitsandfoxes:
The SDK will no longer attempt to reinstall the backend without native support enabled and without successful initialization.
It'll also capture `EntryPointNotFoundException` during reinstallation.

### DIFF
--- a/src/Sentry.Unity.Native/SentryNative.cs
+++ b/src/Sentry.Unity.Native/SentryNative.cs
@@ -14,7 +14,7 @@ public static class SentryNative
 {
     private static readonly Dictionary<string, bool> PerDirectoryCrashInfo = new();
 
-    private static Action? ReinstallBackendFunction;
+    private static bool ShouldReinstallBackend;
 
     /// <summary>
     /// Configures the native SDK.
@@ -84,23 +84,28 @@ public static class SentryNative
         }
         options.CrashedLastRun = () => crashedLastRun;
 
-        // The backend will only be reinstalled if the native SDK has been initialized successfully.
-        // We're handing this off to be called in `BeforeSceneLoad` instead of `SubsystemRegistration` as it's too soon
-        // and the SignalHandler would still get overwritten.
-        ReinstallBackendFunction = SentryNativeBridge.ReinstallBackend;
+        ShouldReinstallBackend = true;
     }
 
+    // We're calling this in `BeforeSceneLoad` instead of `SubsystemRegistration` as it's too soon and the
+    // SignalHandler would still get overwritten.
     [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
-    public static void ReinstallBackend()
+    private static void ReinstallBackend()
     {
+        // The backend should only be reinstalled if the native SDK has been initialized successfully.
+        if (!ShouldReinstallBackend)
+        {
+            return;
+        }
+
         try
         {
             // At this point Unity has taken the signal handler and will not invoke our handler. So we register our
             // backend once more to make sure user-defined data is available in the crash report and the SDK is able
             // to capture the crash.
-            ReinstallBackendFunction?.Invoke();
+            SentryNativeBridge.ReinstallBackend();
         }
-        catch (Exception e)
+        catch (EntryPointNotFoundException e)
         {
             SentrySdk.CaptureException(new Exception("The SDK failed to reinstall the backend and won't capture native errors.", e));
         }
--- a/src/Sentry.Unity.Native/SentryNative.cs
+++ b/src/Sentry.Unity.Native/SentryNative.cs
@@ -111,7 +111,7 @@ public static class SentryNative
         }
         catch (EntryPointNotFoundException e)
         {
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -75,12 +81,37 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
...
         }
         options.CrashedLastRun = () => crashedLastRun;
+
+        ShouldReinstallBackend = true;
     }
 
-    public static void ReinstallBackend() => SentryNativeBridge.ReinstallBackend();
+    // We're calling this in `BeforeSceneLoad` instead of `SubsystemRegistration` as it's too soon and the
+    // SignalHandler would still get overwritten.
+    [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
```
bruno-garcia, line 95: We could have had this in here all along? :)
I dont' recall why we left on `SentryInitialization` originally, do you recall?

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -75,12 +81,37 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
...
+        try
+        {
+            // At this point Unity has taken the signal handler and will not invoke our handler. So we register our
+            // backend once more to make sure user-defined data is available in the crash report and the SDK is able
+            // to capture the crash.
+            SentryNativeBridge.ReinstallBackend();
+        }
+        catch (EntryPointNotFoundException e)
+        {
+            Logger?.LogError(e, "The SDK failed to reinstall the backend - native crashes are not captured.");
```
bruno-garcia, line 114: How about this: not sure anyone knows what reinstalling the backend even means.

```suggestion
            Logger?.LogError(e, "Native dependency not found. Did you delete sentry.dll or move files around?");
```

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -75,12 +81,37 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
...
-    public static void ReinstallBackend() => SentryNativeBridge.ReinstallBackend();
+    // We're calling this in `BeforeSceneLoad` instead of `SubsystemRegistration` as it's too soon and the
+    // SignalHandler would still get overwritten.
+    [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
+    private static void ReinstallBackend()
+    {
+        // The backend should only be reinstalled if the native SDK has been initialized successfully.
+        if (!ShouldReinstallBackend)
+        {
+            Logger?.LogDebug("Skipping reinstalling the native backend as the initialization seems to have failed.");
```
bruno-garcia, line 101: We avoid reinstalling it if fails below so not sure I get this message.

The flag right now is flipped if native support is disabled, right? So there could be different reasons why we disabled it. 

```suggestion
            Logger?.LogWarning("Skipping reinstalling the native backend.");
```

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -75,12 +81,37 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
...
+            return;
+        }
```
bruno-garcia, line 103: how about we also log if we do this, since I suspect we might want to debug this:

```suggestion
        } else {
            Logger?.LogDebug("Reinstalling the native backend to make sure we capture native crashes.");
        }
```

bruno-garcia (commented):
I actually though the flag was "UseNativeBackend" which we'll flip to false if native support was disabled, but also, on the catch block for the MissingEntryPoint.

And we can check that flag before doing any native calls. Or we don't need this because no native call is done anyway because we don't add the scope observers?

So the only thing left is making sure the BeforeScene thing doesn't run, hence the boolean you used in this PR? If that's the case there's no need to write `LogDebug(
"something failed")` because nothing failed, just no backend was selected.

And no need for flipping any flag on the catch block either if no where else after this we'll need to know if it worked or not. (We could disable sessions if we're afraid of reporting 100% crash free rate, so the user will at least see a drop in sessions if it's a widespread error)?

src/Sentry.Unity.Native/SentryNative.cs
bitsandfoxes: The intention was to have everything annotated `RuntimeInitializeOnLoadMethod` in the `SentryInitialization.cs` so users could overwrite/remove it. But it should have belonged in `SentryNative` all along.

bitsandfoxes:
>  Or we don't need this because no native call is done anyway because we don't add the scope observers?

Exactly, if the init failed we don't add the observers and we won't reinstall the backend.

> So the only thing left is making sure the BeforeScene thing doesn't run, hence the boolean you used in this PR? If that's the case there's no need to write LogDebug( "something failed") because nothing failed, just no backend was selected.

Yes. Fair point. Updated the logs! Thanks!

PR #1900, opened 2024-11-13
fix: Logging

bitsandfoxes:
#skip-changelog

PR #1901, opened 2024-11-15
fix: Logger argument order

bitsandfoxes:
Reverts https://github.com/getsentry/sentry-unity/pull/1900

I had https://github.com/getsentry/sentry-dotnet/pull/2715 and https://github.com/getsentry/sentry-dotnet/pull/3630 wrong in my head and assumed the order was the issue.

#skip-changelog

### DIFF
--- a/src/Sentry.Unity.Native/SentryNative.cs
+++ b/src/Sentry.Unity.Native/SentryNative.cs
@@ -44,7 +44,7 @@ public static class SentryNative
         }
         catch (Exception e)
         {
-            Logger?.LogError("Sentry native initialization failed - native crashes are not captured.", e);
+            Logger?.LogError(e, "Sentry native initialization failed - native crashes are not captured.");
             return;
         }
 
@@ -111,7 +111,7 @@ public static class SentryNative
         }
         catch (EntryPointNotFoundException e)
         {
-            Logger?.LogError("Native dependency not found. Did you delete sentry.dll or move files around?", e);
+            Logger?.LogError(e, "Native dependency not found. Did you delete sentry.dll or move files around?");
         }
     }
 }


bruno-garcia (approved):
we really need an analyzer for this.
It's the same logic as `string.Format` so there's for sure analyzers out there that can do this, so it would be a copy of it.

We got an analyzer in the dotnet repo (and the IDiagnosticLogger interface is also there): https://github.com/getsentry/sentry-dotnet/pull/3725

The specialist is @RRode who might be interested in this :)

bruno-garcia:
"you missed a spot":

image

RRode:
> we really need an analyzer for this. It's the same logic as `string.Format` so there's for sure analyzers out there that can do this, so it would be a copy of it.

@bruno-garcia  I also checked, but couldn't find anything for ILogger or in your case IDiagnosticLogger

> We got an analyzer in the dotnet repo (and the IDiagnosticLogger interface is also there): [getsentry/sentry-dotnet#3725](https://github.com/getsentry/sentry-dotnet/pull/3725)
> 
> The specialist is @RRode who might be interested in this :)

Can't promise anything, but will give it a try. If I understand the issue, you would want to prevent wrong method overload usage right? The question here is more what makes sense to allow. IMO it never makes sense to provide an exception as an string format param / args, since the default ToString doesn't provide much of a logging value. Also providing exception properties under params seems somewhat off, since the exception overload should log all relevant exception information. So is it fine to raise a warning if an exception is found under params?

bitsandfoxes:
> So is it fine to raise a warning if an exception is found under params?

Ideally, it'd be a bit more "in your face" than just a warning.

bruno-garcia:
We have warnings as errors, so that would be fine

Issue #1918, opened 2024-11-23
NullReferenceException with ViewHierarchy


bruno-garcia:
[Came up on Discord](https://discordapp.com/channels/621778831602221064/621778831602221072/1309967617519517847)

[Based on the error message and logic aroud](https://github.com/getsentry/sentry-dotnet/blob/3832bcc15aab98ad642470c52515ef144304ab96/src/Sentry/Protocol/Envelopes/Envelope.cs#L273), the event shouldn't be lost. Only the ViewHierarchy would be:

Based on the stack trace, something is `null` here:

https://github.com/getsentry/sentry-unity/blob/bac420f65b119bab0cf64cf6820a7ce1f183c811/src/Sentry.Unity/UnityViewHierarchyAttachmentContent.cs#L82

```
Sentry: (Error) Failed to add attachment: view-hierarchy.json. System.NullReferenceException: Object reference not set to an instance of an object
  at Sentry.Unity.UnityViewHierarchyAttachmentContent+<>c.<CreateNode>b__5_0 (UnityEngine.Component e) [0x00000] in /sentry-unity/src/Sentry.Unity/UnityViewHierarchyAttachmentContent.cs:82 
  at System.Linq.Enumerable+SelectListIterator`2[TSource,TResult].ToList () [0x0002a] in <2a397996daff412691ae4781443e8e88>:0 
  at System.Linq.Enumerable.ToList[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x0001f] in <2a397996daff412691ae4781443e8e88>:0 
  at Sentry.Unity.UnityViewHierarchyAttachmentContent.CreateNode (System.Int32 remainingDepth, System.Int32 maxChildCount, Sentry.ViewHierarchyNode parentNode, UnityEngine.Transform transform) [0x0000e] in /sentry-unity/src/Sentry.Unity/UnityViewHierarchyAttachmentContent.cs:75 
  at Sentry.Unity.UnityViewHierarchyAttachmentContent.CreateNode (System.Int32 remainingDepth, System.Int32 maxChildCount, Sentry.ViewHierarchyNode parentNode, UnityEngine.Transform transform) [0x000f2] in /sentry-unity/src/Sentry.Unity/UnityViewHierarchyAttachmentContent.cs:96 
[...]

PR #1919, opened 2024-11-23
fix: null ref in view hierarchy

bruno-garcia:
Resolves: https://github.com/getsentry/sentry-unity/issues/1918

Skips possible `null` components before trying to get their type info.

Not sure how to test this because it depends on a Unity API `GetComponents`

Also: replaced the LINQ call with a foreach and allocated a new List with the known size to avoid resizes.

### DIFF
--- a/src/Sentry.Unity/UnityViewHierarchyAttachmentContent.cs
+++ b/src/Sentry.Unity/UnityViewHierarchyAttachmentContent.cs
@@ -72,6 +72,14 @@ internal class UnityViewHierarchyAttachmentContent : IAttachmentContent
     {
         var components = new List<Component>();
         transform.GetComponents(components);
+        var extras = new List<string>(components.Count);
+        foreach (var component in components)
+        {
+            if (component?.GetType().Name is { } componentName)
+            {
+                extras.Add(componentName);
+            }
+        }
         var node = new UnityViewHierarchyNode(transform.name)
         {
             Tag = transform.tag,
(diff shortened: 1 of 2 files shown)


src/Sentry.Unity/UnityViewHierarchyAttachmentContent.cs
```diff
@@ -72,14 +72,22 @@ internal void CreateNode(int remainingDepth, int maxChildCount, ViewHierarchyNod
     {
         var components = new List<Component>();
         transform.GetComponents(components);
+        var extras = new List<string>(components.Count);
+        foreach (var component in components)
+        {
+            if (component?.GetType().Name is { } componentName)
```
bruno-garcia, line 78: The stack trace on https://github.com/getsentry/sentry-unity/issues/1918 led me to believe `GetComponents` is adding `null` to entries to the List it receives as argument.

bruno-garcia:
CI is showing green here but only 5 checks show. I had to cancel CI due to #1920

bitsandfoxes (approved):
🔥 🚒

bitsandfoxes:
> CI is showing green here but only 5 checks show.

CI will skip if the only changes are made to `.txt` or `.md` - i.e. the changelog.
https://github.com/getsentry/sentry-unity/blob/89c7acf18d75387f04f4b805487b51f6bb00e6b8/.github/workflows/ci.yml#L3-L8

PR #1921, opened 2024-11-26
chore: Remove `-mno-thumb` from Xcode projects for unity-of-bugs

bitsandfoxes:
### Why?

With versions 2019 and 2020 Unity exports with a `-mno-thumb` flag by default. With the current version of Xcode (16.1) building those projects fails with `"unsupported option '-mno-thumb' for target"`. Those versions of Unity are also out-of-support and won't likely receive an update.

### How this fixes it

I've added added a `PostProcessBuild` that removes the flag. This allows us to still build for iOS on Unity 2019 and 2020. This script resides inside of `Assets/Editor/` and is excluded from being packaged. Endusers will not have his script as part of the package.

#skip-changelog


scripts/pack.ps1
```diff
@@ -23,7 +23,6 @@ Copy-Item "LICENSE.md" -Destination "package-release/LICENSE.md"
 
 # Copy samples
 Copy-Item "samples/unity-of-bugs/Assets/Scenes*" -Destination "package-release/Samples~/unity-of-bugs/" -Recurse
-Copy-Item "samples/unity-of-bugs/Assets/Editor*" -Destination "package-release/Samples~/unity-of-bugs/" -Recurse
```
bitsandfoxes, line 26: We had nothing there in the first place.

samples/unity-of-bugs/Assets/Editor/XcodeProjectUpdater.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        [PostProcessBuild]
+        public static void OnPostProcessBuild(BuildTarget target, string pathToBuildProject)
+        {
+            if (target != BuildTarget.iOS)
+            {
+                return;
+            }
+
+            Debug.Log("XcodeUpdater.OnPostProcessBuild started.");
```
bruno-garcia, line 25: This logger doesn't let us opt out or set level right? are we not using our logger here why?

bitsandfoxes: This is a script that lives outside the SDK and is not part of the shipped package.
It's is purely there to help us run builds when targeting iOS on old Unity versions.

Issue #1907, opened 2024-11-19
[Unity] Update initialization behaviour on `mobile`


bitsandfoxes:
# Programmatic Configuration and Native Support on `Mobile`

## Background

The SDKs current approach let's the native SDKs self-initialize when targeting `mobile`. This approach guarantees the SDK to capture all errors (even Unity engine crashes) but it also creates a lot of friction for users during onboarding and setup.
Having to explain the different layers is a non-trivial task. 

Example:
- A user wants to set the environment:
  For that they have to implement both, the Runtime and the BuildTime configuration. Even with the unified approach of having only one configure callback, they are still limited to setting a static `string` and are not able to compute the value at runtime. That limitation has to somehow be explained again.

## Current State

The Unity SDK currently handles native support differently across platforms:

### Mobile Platforms (Android, iOS)
- **Current Implementation**: Native SDKs self-initialize outside Unity runtime
- **Mechanism**: Configured through Gradle (Android) and Xcode (iOS) project modifications
- **Advantages**:
  - Can capture Unity crashes
- **Challenges**:
  - Configuration gets "baked" into the project at build time
  - Complex configuration layers confuse users.
  - Inconsistent runtime/build time configuration patterns even with the unified configure callback

### Desktop Platforms (Windows/Linux/macOS)
- **Current Implementation**: Native SDKs initialize within Unity runtime
- **Behavior**: Uses Configure callback consistently across C# Unity SDK and native layers. The options at the end of `Configure` are the ones that initialize the native layer

### 1. Unified Configuration API (mostly done already)
- Replace separate Runtime/BuildTime configuration with single `Configure` method
  - Issue #1887: Separate CLI options from build time configure
  - Issue #1888: Implement unified scriptable object for configuration callback
  - Deprecate existing Runtime/BuildTime configuration callbacks

#### iOS Implementation
- Framework handling:
  - ~~Instead of handling filecopy ourselves we can mark `.framework` for targeting iOS.~~
  - ~~Unity will handle that for us since we'll never know whether we'll need it. The SDK can and will be initialized during runtime.~~
  - We will have to rely on the already built Xcode setup functionality to copy the framework and bridge. Unty does not recognize `.xcframework` on versions 2020 and older.
- Initialization:
  - Use native bridge for SDK initialization. We've already got a working example for macOS.
  - We can rely on the Cocoa SDK's `isEnabled` property to avoid reinitialization.

#### Android Implementation
- Package handling:
[...]

PR #1865, opened 2024-10-27
Update android-smoke-test.yml

lucas-zimerman:
use older  macos

PR #1924, opened 2024-11-28
feat: Managed layer initializes native SDK on Android

bitsandfoxes:
# Android: Android SDK Initialization

## Overview
Implements the late initialization of the Android SDK. This allows users to apply code changes to the options and for those options to apply to the native layer. Related issue: [sentry-unity#1907](https://github.com/getsentry/sentry-unity/issues/1907)

## Implementation Details
The SDK now always includes the Android SDK in the generated Gradle project with the following changes:

1. **New Native Support Option**
  - Added an enum `AndroidNativeInitializationType` to the SentryOptions. This allows users to switch between `Runtime` and `BuildTime` initialization and defaults to `Runtime`.
  - The `AndroidNativeSupportEnabled` controls whether the SDK should add native support. This helps i.e. in the case of UaaL where users want to completely disabled the SDK from modifying the Gradle project.

2. **Build Process Changes**
  - Adding the native options to the manifest is now controlled by `AndroidNativeInitializationType` enum
    - With type `Runtime` the SDK will add `auto-init = false`.
    - With type `BuildTime` it will add the options and let the Android SDK autoinitialize.
   
3. **Runtime Behavior**
  - `SentryNativeAndroid.Configure` checks whether the native SDK was already initialized
  - Both early and late initialization are going through the same code paths: i.e. setting up scope observer. 

4. **JNI Executor Behavior**
  - The JNI Executor defaults to timeout after 16ms as to not drop more than a frame
  - Android SDK initialization takes about 25ms on a current device. So the timeout is now overwriteable when calling into the executor.

## Pending Tasks
- [ ] `RuntimeInitialization` is default. This means we're missing testing `BuildTime` in CI right now.
- [ ] "Init Native First" flow
- [x]  "Init Native Late" flow
- Note: Requires duplicate mobile smoke testing (2x build, compile, test)


src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -40,6 +42,55 @@ internal class SentryJava : ISentryJava
...
 
+    public void Init(IJniExecutor jniExecutor, SentryUnityOptions options)
+    {
+        jniExecutor.Run(() =>
+        {
+            using var sentry = new AndroidJavaClass("io.sentry.android.core.SentryAndroid");
+            using var context = new AndroidJavaClass("com.unity3d.player.UnityPlayer")
+                .GetStatic<AndroidJavaObject>("currentActivity");
+
+            sentry.CallStatic("init", context, new OptionsConfiguration(androidOptions =>
```
bruno-garcia, line 53: Can this throw? I imagine do this in a `try` `catch` and just degrade to C# only support is a good approach.
If it fails you can flip that flag that native support is off, and avoid the P/Invoke calls

src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -40,6 +42,55 @@ internal class SentryJava : ISentryJava
...
+    }
+
+    // Update the callback class to match the Java interface name
+    internal class OptionsConfiguration : AndroidJavaProxy
+    {
+        private readonly Action<AndroidJavaObject> _callback;
+        private readonly IDiagnosticLogger? _logger;
+
+        public OptionsConfiguration(Action<AndroidJavaObject> callback, IDiagnosticLogger? logger)
+            : base("io.sentry.Sentry$OptionsConfiguration")
```
bruno-garcia, line 70: Any concerns that proguard can mess with this? I imagine not, because we have proguard rules. Assuming those get passed in to the final build.
I imagine this is already a concern to all the other calls we do into Java

src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -40,6 +42,55 @@ internal class SentryJava : ISentryJava
...
+            {
+                if (methodName != "configure" || args.Length != 1)
+                {
+                    throw new Exception($"Invalid invocation: {methodName}({args.Length} args)");
+                }
+                _callback(args[0]);
+            }
+            catch (Exception e)
+            {
+                _logger?.LogError(e, "Error in SentryJava.OptionsConfiguration: {0}");
```
bruno-garcia, line 88: Is the `{0}` correct here? If we're passing just the exception as the first arg, does that take a positioned arg? I thought it was only stuff that comes after the string argument

src/Sentry.Unity.Android/SentryJava.cs
bitsandfoxes: Yes. I've improved on this in https://github.com/getsentry/sentry-unity/pull/1927

src/Sentry.Unity.Android/SentryJava.cs
bitsandfoxes: I don't think so. We're using the very same mechanism for the scope sync so far tho.

src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -40,6 +43,88 @@ internal class SentryJava : ISentryJava
...
+                androidOptions.Call("setMaxCacheItems", options.MaxCacheItems);
+                androidOptions.Call("setSendDefaultPii", options.SendDefaultPii);
+                // Note: doesn't work - produces a blank (white) screenshot
+                androidOptions.Call("setAttachScreenshot", false);
+                androidOptions.Call("setEnableNdk", options.NdkIntegrationEnabled);
+                androidOptions.Call("setEnableScopeSync", options.NdkScopeSyncEnabled);
+                androidOptions.Call("setEnableAutoSessionTracking", false);
+                androidOptions.Call("setEnableActivityLifecycleBreadcrumbs", false);
+                androidOptions.Call("setAnrEnabled", false);
+                androidOptions.Call("setEnableScopePersistence", false);
```
bruno-garcia, line 90: Consider spliting it more clearly like this:

```suggestion
                androidOptions.Call("setEnableNdk", options.NdkIntegrationEnabled);
                androidOptions.Call("setEnableScopeSync", options.NdkScopeSyncEnabled);

                // Options that are not to be set by the user

                // Note: doesn't work - produces a blank (white) screenshot
                androidOptions.Call("setAttachScreenshot", false);
                androidOptions.Call("setEnableAutoSessionTracking", false);
                androidOptions.Call("setEnableActivityLifecycleBreadcrumbs", false);
                androidOptions.Call("setAnrEnabled", false);
                androidOptions.Call("setEnableScopePersistence", false);
```

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -160,6 +160,12 @@ public sealed class SentryUnityOptions : SentryOptions
     /// </summary>
     public bool AndroidNativeSupportEnabled { get; set; } = true;
 
+    /// <summary>
+    /// Whether the SDK should initialize the native SDK before the Unity game. This bakes the options at build-time into
+    /// the generated Gradle project
+    /// </summary>
```
bruno-garcia, line 166: This option need to be documented. 

An idea: the paragraph explaining what "baking into" means, could be a a paragraph that u make an 'include' so u can add on the iOS as well as Android option (since there are two).

```suggestion
    /// </summary>
    /// <seealso href="https://docs.sentry.io/platforms/unity/configuration/options/#android-native-initialization-type"
```

bruno-garcia (commented):
CI seems not to be running

src/Sentry.Unity/SentryUnityOptions.cs
bitsandfoxes: I've added additional remarks to the enum itself and to the `inittype` options. I'll add the link to our docs once they are updated.

bruno-garcia:
image

Smoke Test still unreliable?

modules/sentry-native
bruno-garcia, line 1: Is this related?

src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -40,6 +43,93 @@ internal class SentryJava : ISentryJava
...
+
+    public bool Init(IJniExecutor jniExecutor, SentryUnityOptions options)
+    {
+        jniExecutor.Run(() =>
+        {
+            using var sentry = new AndroidJavaClass("io.sentry.android.core.SentryAndroid");
+            using var context = new AndroidJavaClass("com.unity3d.player.UnityPlayer")
+                .GetStatic<AndroidJavaObject>("currentActivity");
+
+            sentry.CallStatic("init", context, new AndroidOptionsConfiguration(androidOptions =>
```
bruno-garcia, line 63: Curious if this affects the startup time of a game in any significant way. Since now we're doing this via P/Invoke what used to be direct

src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -40,6 +43,93 @@ internal class SentryJava : ISentryJava
...
+                if (methodName != "configure" || args.Length != 1)
+                {
+                    throw new Exception($"Invalid invocation: {methodName}({args.Length} args)");
+                }
+
+                _callback(args[0]);
+            }
+            catch (Exception e)
+            {
+                _logger?.LogError(e, "Error in SentryJava.OptionsConfiguration.");
```
bruno-garcia, line 127:
```suggestion
                _logger?.LogError(e, "Error invoking {0} ’.", methodName);
```

src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -39,6 +39,16 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
 
         JniExecutor ??= new JniExecutor();
 
+        if (SentryJava.IsEnabled(JniExecutor))
+        {
+            options.DiagnosticLogger?.LogDebug("The native SDK is already initialized");
```
bruno-garcia, line 44: Keep terminology consistent (see log msg below)

```suggestion
            options.DiagnosticLogger?.LogDebug("The Android Native SDK is already initialized");
```

src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -98,6 +108,8 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
                 options.DiagnosticLogger?.LogDebug("Failed to create new 'Default User ID'.");
             }
         }
+
+        options.DiagnosticLogger?.LogInfo("Successfully configured the native SDK");
```
bruno-garcia, line 112:
```suggestion
        options.DiagnosticLogger?.LogInfo("Successfully configured the Android Native SDK");
```

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -150,16 +150,34 @@ public sealed class SentryUnityOptions : SentryOptions
...
     /// <summary>
-    /// Whether the SDK should initialize the native SDK before the Unity game. This bakes the options at build-time into
-    /// the generated Xcode project
+    /// Whether the SDK should initialize the native SDK before the game starts. This bakes the options at build-time into
+    /// the generated Xcode project. Modifying the options at runtime will not affect the options used to initialize
+    /// the native SDK.
     /// </summary>
+    /// <remarks>
+    /// When set to <see cref="NativeInitializationType.BuildTime"/>, the options are written and hardcoded into the
+    /// Gradle project during the build process. This means that the options cannot be changed at runtime, as they are
```
bruno-garcia, line 159: Gradle on iOS?

test/Sentry.Unity.Android.Tests/SentryNativeAndroidTests.cs
```diff
@@ -81,28 +92,68 @@ public void Configure_IL2CPP_ReInitializesNativeBackend(bool il2cpp, bool expect
...
+
+        Assert.IsTrue(_logger.Logs.Any(log =>
+            log.logLevel == SentryLevel.Debug &&
+            log.message.Contains("The native SDK is already initialized")));
+
+        Assert.NotNull(_options.ScopeObserver);
+    }
+
+    [Test]
+    public void Configure_NativeInitFails_LogsErrorAndReturns()
```
bruno-garcia, line 146: There was so many changes done above. Are these 3 tests covering all of those?

Some values should be default false, etc.

bruno-garcia (commented):
I see 5 new tests for quite a significant change. I'm concerned there are lot of untested use cases. Isn't that the case?

src/Sentry.Unity.Android/SentryJava.cs
bitsandfoxes: From local testing, `Init` takes about 25ms. Based on what I've got from the Android SDK itself, self-init takes around that time dring self-initialization as well.
So we just moved the time from "before Unity starts" to "while Unity starts" and the P/Invoke does not add any significant overhead to initialization.

test/Sentry.Unity.Android.Tests/SentryNativeAndroidTests.cs
bitsandfoxes: The only behavioral change that we did was introduce `Init` to the `Configure`. The only real thing to check is whether `Init` gets called and validate the fallbacks.

The rest is getting covered by our integration tests. I'll fix those up next as part of https://github.com/getsentry/sentry-unity/pull/1865

src/Sentry.Unity.Android/JniExecutor.cs
```diff
@@ -50,74 +61,110 @@ private void DoWork()
...
     {
+        if (_isDisposed)
+        {
+            return;
+        }
+
+        _isDisposed = true;
+
         _shutdownSource.Cancel();
+        _workerThread?.Join(100);
```
bruno-garcia, line 167: It can throw for:

Exceptions
[ThreadStateException](https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadstateexception?view=net-9.0)
The caller attempted to join a thread that is in the [Unstarted](https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadstate?view=net-9.0#system-threading-threadstate-unstarted) state.

[ThreadInterruptedException](https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadinterruptedexception?view=net-9.0)
The thread is interrupted while waiting.


We should catch this here and suppress it

bruno-garcia (approved):
One issue left which is the `thread.Join` on Dispose which can throw.

Other than that, lgtm

src/Sentry.Unity.Android/JniExecutor.cs
bruno-garcia: Also worth reading the remarks of these APIs.

Curious why we use Thread directly instead of the Task abstraction. It's by default a background thread and it's much better to deal with cancellation, error handling etc (something for a potential future task, to change this)

vaind: > Curious why we use Thread directly instead of the Task abstraction.

The executor needs to run all the calls on a single thread. If we converted this to a task, we'd still need to ensure that so no async/await, etc. In that case, I only see risks that we do something wrong here now or in the future version of the code which would lead to the task being rescheduled to a different thread in the pool.

PR #1856, opened 2024-10-21
chore: Updated Android API levels for smoke tests

bitsandfoxes:
#skip-changelog

PR #1936, opened 2024-12-16
fix: Android device tests

bitsandfoxes:
This PR gets our Android device tests back in working order.

#### Changes to `android-smoke-test.yml`
Simplified our emulator setup by removing the extra configuration steps. We had some complexity trying to speed things up with snapshots, but since we weren't using them anyway, I stripped it out. Tests now run in 2-5 minutes which seems pretty reasonable.

#### Changes to `smoke-test-android.ps1`
Gave this script a good cleanup! Removed a bunch of old workarounds we don't need anymore since the emulator config handles these now (No need to swipe away popups if those are disabled ;) ). Added retry logic for app launches since they can be a bit flaky, but kept it within the same test run. We could probably remove the `android-smoke-test-wrapper` entirely, but let's see how this works in production first. Also cleaned up the logs in CI to make them actually readable 😅.

#### Changes to `SmokeTester.cs`
Added a "Smoke Test finished" message that we can watch for in the emulator. This gives us a reliable way to know when tests are actually done instead of letting it time out and then check the logs for `pass`/`fail`.

#### Removed `android-smoke-test-wrapper.yml`
Since the retry logic is now part of the actual smoke test there is no longer a need to rerun the entire job.

#### Changes to `ci.yml`
Updated our Android API targets (this replaces PR #1856)

#skip-changelog


.github/workflows/android-smoke-test.yml
```diff
@@ -7,19 +7,16 @@ on:
...
+      status:
+        description: "Smoke test status"
+        value: ${{ jobs.run.outputs.status }}
 
 jobs:
   run:
-    name: try-${{ inputs.try }}
-    runs-on: macos-latest
+    name: Android Smoke Test
+    runs-on: ubuntu-latest
```
bruno-garcia, line 19:
```suggestion
    runs-on: ubuntu-latest-4-cores
```

bruno-garcia: Worth trying this, see if things get faster.
Might just break because the VM images are not the same

bruno-garcia (approved):

.github/workflows/android-smoke-test.yml
```diff
@@ -64,29 +68,28 @@ jobs:
           "target=$target" >> $env:GITHUB_OUTPUT
           "api-level=$apiLevel" >> $env:GITHUB_OUTPUT
           "label=$($label ?? $apiLevel)" >> $env:GITHUB_OUTPUT
-
-      - name: Android API ${{ steps.config.outputs.label }} emulator setup + Smoke test
-        uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b # pin@v2.28.0
+      
+      - name: Run Android Smoke Tests
+        uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b # pin@v233.0
```
vaind, line 73:
```suggestion
        uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b # pin@v2.33.0
```

.github/workflows/ci.yml
```diff
@@ -463,26 +463,15 @@ jobs:
...
-    uses: ./.github/workflows/android-smoke-test-wrapper.yml
+    uses: ./.github/workflows/android-smoke-test.yml
     with:
       unity-version: ${{ matrix.unity-version }}
       api-level: ${{ matrix.api-level }}
     strategy:
       fail-fast: false
       matrix:
-        api-level: [27, 28, 29, 30, 31] # last updated October 2022
+        api-level: [27, 31, 34] # last updated Dezember 2024
```
vaind, line 473: 🇩🇪 
```suggestion
        api-level: [27, 31, 34] # last updated December 2024
```

.github/workflows/android-smoke-test.yml
```diff
@@ -64,29 +68,28 @@ jobs:
...
+            pwsh ./scripts/smoke-test-android.ps1 -IsIntegrationTest -WarnIfFlaky
 
       - name: Upload artifacts on failure
         if: ${{ failure() }}
         uses: actions/upload-artifact@v4
         with:
           name: testapp-android-logs-${{ inputs.api-level }}-${{ inputs.unity-version }}-${{ inputs.try }}
           path: ${{ env.ARTIFACTS_PATH }}
-          retention-days: 14
+          retention-days: 14
```
vaind, line 95:
```suggestion
          retention-days: 14

```

vaind (approved):
LGTM other than some lints. As long as CI is happy, I'm happy

Issue #1954, opened 2025-01-09
ANR improvements - Mechanism and docs


bruno-garcia:
Today doesn't seem like we set any Mechanism:
https://github.com/getsentry/sentry-unity/blob/bee77f39ad1137cb21082721d1a2b5fa9fc6f83d/src/Sentry.Unity/Integrations/AnrIntegration.cs#L81

The .NET SDK has a function to help set it:
https://github.com/getsentry/sentry-dotnet/blob/c2c737b37882ae5a9a34e17de9386eb620f00ba4/src/Sentry/SentryExceptionExtensions.cs#L35

These are used by customers to filter things out in Sentry. As well as a clear way to see where things came from when looking at an event.

Additionally, ANR tends to mean that the app crashed because the UI thread is blocked. (this can also be added to the mechanism description)
So might be worth clarifying that this is not the case with Unity.

Mobile: Also worth clarifying in the docs that this detection is not the same as the Android and iOS, and how they differ. And that they automatic instrumentation disables the native detection because they don't work on Unity.

Angelodaniel:
I see @bitsandfoxes 's PR which is merged, is this issue still open? Or is now the functionality available but just not the docs? So it's partly done? @bruno-garcia

bruno-garcia:
There's no link to the docs from the description though, not sure @bitsandfoxes wants to follow up:
https://github.com/getsentry/sentry-unity/pull/1955/files

![Image](https://github.com/user-attachments/assets/fd1822e4-f787-48e0-a8d1-850de2cfc0b9)

Docs go in detail about this: https://docs.sentry.io/platforms/unity/configuration/app-not-responding/

Including listing current limitations like no stack trace of the UI thread

Angelodaniel:
Just to confirm, we don't mention this in the Docs but it's indeed being captured now.

PR #1955, opened 2025-01-09
feat: Add mechanism to ANR event

bitsandfoxes:
Addresses the improvement part of https://github.com/getsentry/sentry-unity/issues/1954

In line with Java and the [V1 ANR watchdog](https://github.com/getsentry/sentry-java/blob/bfbaf37fae93cd88c3c1cd570f6296af1bf064be/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java#L144) (documented [here](https://arc.net/l/quote/zzpsagea)), the SDK will now add the `ANR` mechanism when capturing an ANR event.

Relates to:

* https://github.com/getsentry/sentry-unity/issues/1961

### DIFF
--- a/src/Sentry.Unity/Integrations/AnrIntegration.cs
+++ b/src/Sentry.Unity/Integrations/AnrIntegration.cs
@@ -78,7 +78,10 @@ internal abstract class AnrWatchDog
         {
             var message = $"Application not responding for at least {DetectionTimeoutMs} ms.";
             Logger?.LogInfo("Detected an ANR event: {0}", message);
-            OnApplicationNotResponding?.Invoke(this, new ApplicationNotRespondingException(message));
+
+            var exception = new ApplicationNotRespondingException(message);
+            exception.SetSentryMechanism("ANR", handled: false);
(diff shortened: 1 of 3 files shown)


src/Sentry.Unity/Integrations/AnrIntegration.cs
```diff
@@ -78,7 +78,10 @@ protected void Report()
         {
             var message = $"Application not responding for at least {DetectionTimeoutMs} ms.";
             Logger?.LogInfo("Detected an ANR event: {0}", message);
-            OnApplicationNotResponding?.Invoke(this, new ApplicationNotRespondingException(message));
+
+            var exception = new ApplicationNotRespondingException(message);
+            exception.SetSentryMechanism("ANR", handled: false);
```
bruno-garcia, line 83: Could benefit some description. E.g: "doesn't crash the app"
I suggested that here: https://github.com/getsentry/sentry-unity/issues/1954

bruno-garcia (commented):
Missing tests

bruno-garcia (approved):
nice! Thanks Stefan

PR #1956, opened 2025-01-10
fix: CI - Set correct Java version

bitsandfoxes:
The default Java version is different on different runner images. An update to `-latest` causes Unity 2019 builds to fail.

With this, we explicitly set the correct Java version depending on the targeted Unity version, making this more resilient to future changes.
I've also added timeouts to the smoketests themselves. They tend to either fail fast or run indefinitely. This gets rid of the latter.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -493,28 +493,7 @@ jobs:
 
       - name: Setup Java for Unity
         if: ${{ matrix.platform == 'Android' }}
-        run: |
-          switch -Regex ("${{ matrix.unity-version }}")
-          {
-              "2019" {
-                  $javaHome = $env:JAVA_HOME_8_X64
-                  Write-Host "Using Java 8 for Unity 2019"
-              }
-              "2022" {
-                $javaHome = "$env:JAVA_HOME_11_X64"
-                Write-Host "Using Java 11 for Unity 2022"
-              }
-              "6000" {
-                  $javaHome = "$env:JAVA_HOME_17_X64"
-                  Write-Host "Using Java 17 for Unity 6"
-              }
-              default {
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -491,6 +491,31 @@ jobs:
...
+        if: ${{ matrix.platform == 'Android' }}
+        run: |
+          switch -Regex ("${{ matrix.unity-version }}")
+          {
+              "2019" {
+                  $javaHome = $env:JAVA_HOME_8_X64
+                  Write-Host "Using Java 8 for Unity 2019"
+              }
+              "2022" {
+                $javaHome = "$env:JAVA_HOME_11_X64"
```
bruno-garcia, line 504: Do these need to be installed by us before hand or they are part of the build agent's image?
If something is missing (one of these paths is empty/undefined) how does the error manifest? Do we have some sane way to immediately stop saying "Didn't find X, we need that to work"

.github/workflows/ci.yml
```diff
@@ -500,23 +525,18 @@ jobs:
...
-        if: ${{ matrix.platform == 'Android' && matrix.unity-version == '6000' }}
-        uses: actions/setup-java@v3
-        with:
-          java-version: '17'
-          distribution: 'temurin'
   
       - name: Android smoke test
         if: ${{ matrix.platform == 'Android' }}
         run: ./scripts/smoke-test-android.ps1 Build -IsIntegrationTest -UnityVersion "${{ matrix.unity-version }}"
+        timeout-minutes: 10
```
bruno-garcia, line 532: Nice

bruno-garcia (approved):

PR #1957, opened 2025-01-10
fix: Update to Unity `2022.3.55f1`

bitsandfoxes:
The current LTS of Unity requires a bump to `ndk: "r23b"`.
This PR contains workarounds for an [ongoing issue](https://discussions.unity.com/t/gradle-build-issues-for-android-api-sdk-35-in-unity-2022-3lts/1502187/10) with the current 2022 LTS. This requires us to fix an additional absolut path set in the gradle properties and to pin the targeted Android API target.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -463,7 +463,7 @@ jobs:
           - unity-version: "2019"
             ndk: "r19"
           - unity-version: "2022"
-            ndk: "r21d"
+            ndk: "r23b"
           - unity-version: "6000"
             ndk: "r23b"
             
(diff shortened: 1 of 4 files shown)


bruno-garcia (approved):
I worry about how many different versions and checks are spread throughout the repo to get this to work. Is there a world we can keep a map of versions in a single file for example, that we parse and then things run on the correct version?

assigned @vaind for review here since he touched a lot of this stuff too

vaind (approved):
Looks reasonable to me

bitsandfoxes:
> I worry about how many different versions and checks are spread throughout the repo to get this to work. Is there a world we can keep a map of versions in a single file for example, that we parse and then things run on the correct version?

I get that and parts of it make me nervous too. But we've got the individual changes and checks bundled in their respective place, i.e. 
- `modify-gradle-project.ps1` contains all the changes to the exported gradle project to make it build in CI.
- `Builder.cs` contains all the changes necessary to properly create an artifact in CI.

PR #1958, opened 2025-01-10
fix: CI Android device testing

bitsandfoxes:
When running on `ubuntu-latest` starting the emulator would result in 
```
  /usr/bin/sh: 1: cannot create /home/runner/.android/avd/test.avd/config.ini: Directory nonexistent
```
Creating the directories beforehand seems to resolve this.

The most relevant change in this PR is the fix `return 1` -> `exit 1`. Since this is happening on the "root" scope this prematurely exists the script without actually setting the exit code. This causes CI to turn green even tho the tests failed.

#skip-changelog

### DIFF
--- a/.github/workflows/android-smoke-test.yml
+++ b/.github/workflows/android-smoke-test.yml
@@ -43,6 +43,7 @@ jobs:
           sudo udevadm control --reload-rules
           sudo udevadm trigger --name-match=kvm
       
+      # Make sure that the required directories and .cfg do exists. Workaround to keep ADV happy on `ubuntu-latest`.
       - name: Setup Android directories
         run: |
           mkdir -p $HOME/.android
--- a/.github/workflows/android-smoke-test.yml
+++ b/.github/workflows/android-smoke-test.yml
@@ -16,7 +16,7 @@ on:
 jobs:
   run:
     name: Android Smoke Test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     env:
       ARTIFACTS_PATH: samples/IntegrationTest/test-artifacts/
       HOMEBREW_NO_INSTALL_CLEANUP: 1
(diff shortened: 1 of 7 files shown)


.github/workflows/android-smoke-test.yml
```diff
@@ -42,43 +42,33 @@ jobs:
...
-            Write-Output "Current API is $apiLevel"
-          }
-          # Setup Arch and Target
-          $target = $apiLevel -ge 30 ? 'google_apis' : 'default'
-          Write-Output "Current Target is $target"
-          "target=$target" >> $env:GITHUB_OUTPUT
-          "api-level=$apiLevel" >> $env:GITHUB_OUTPUT
-          "label=$($label ?? $apiLevel)" >> $env:GITHUB_OUTPUT
       
+      - name: Setup Android directories
```
bitsandfoxes, line 46:
```suggestion
      # Make sure that the required directories and .cfg do exists. Workaround to keep ADV happy on `ubuntu-latest`.
      - name: Setup Android directories
```

.github/workflows/android-smoke-test.yml
```diff
@@ -42,43 +42,34 @@ jobs:
           echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
           sudo udevadm control --reload-rules
           sudo udevadm trigger --name-match=kvm
-          
-      # outputs variables: api-level, label, target
-      - name: Configure Android Settings
```
bitsandfoxes, line 47: Since we're bumping the target APIs we no longer need this step.

.github/workflows/android-smoke-test.yml
```diff
@@ -16,7 +16,7 @@ on:
 jobs:
   run:
     name: Android Smoke Test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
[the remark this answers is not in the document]
bitsandfoxes: True dat. Instead of pinning we're not creating the working-directories manually.

vaind (approved):

PR #1959, opened 2025-01-10
fix: Logging during `option creation`, `Configure` and `Init`

bitsandfoxes:
Previously, we've setup logging before calling the user's `Configure` callback. But by doing so, we're ignoring changes to the options until the SDK itself initializes. That way, if the `ScriptableUnityOptions` would not have `debug` enabled, we'd miss out on all debug logs between option creation and initialization - That's native configuration in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- The SDK now  properly sets up logging by respecting the debug settings set during the configure callback. Logs created during the configuration of the native SDKs no longer get lost ([1959]https://github.com/getsentry/sentry-unity/pull/1959)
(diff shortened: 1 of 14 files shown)


bruno-garcia (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- The SDK now  properly sets up logging by respecting the debug settings set during the configure callback. Logs created during the configuration of the native SDKs no longer get lost ([1959]https://github.com/getsentry/sentry-unity/pull/1959)
```
```suggestion
- The SDK now  properly sets up logging by respecting the debug settings set during the configure callback. Logs created during the configuration of the native SDKs no longer get lost ([#1959]https://github.com/getsentry/sentry-unity/pull/1959)
```

Issue #1951, opened 2025-01-09
`Debug.LogError`s are landing in the `production` environment, despite `SentryOptionsConfigurationObject` settings


wyattPD:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)

We use Sentry SaaS.

Which version of the SDK?

2.4.0

How did you install the package? (Git-URL, Assetstore)
Git-URL

Which version of Unity?
2022.3.37f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Android, iOS

### Steps to Reproduce

1. Create a new `SentryOptionsConfigurationObject` via the Sentry SDK UI (this replaces the Build/Run time configuration objects)
2. This object sets the `Environment` field to `pd_dev` or `pd_prod` depending on the context of the build. We do _not_ use the word `production` which is Sentry's default bucket if we left this blank.
4. This object also logs what environment it's in via a `Debug.Log` eg: `"Hello I am in the [dev] environment"`
5. Add a button in a scene that runs `Debug.LogError("Hello, i am error!")`
6. Add a button in the scene that runs `UnityEngine.Diagnostics.Utils.ForceCrash(ForcedCrashCategory.Abort)` (this will force the app to crash)
7. Build the game for iOS or Android.
8. Run the game on device, press each button
### Expected Result
At build-time:
- I see the message telling me what environment I'm in. (eg: "Hello I am the [prod] environment")

At runtime, on device:
- I see the same log message (I'm using SRDebugger to see my console logs on-device).
- Any Debug.LogError, Exception, or Crash should show up under the `pd_prod` / `pd_dev` environment.

### Actual Result

At build-time:

- ✅ I see the message telling me what environment I'm in. (eg: "Hello I am the [prod] environment")

At runtime, on device:

- 🟨 I do not see the log messages telling me what environment I'm in. Implying the `SentryOptionsConfigurationObject` did not execute at runtime.
- ✅ **When the app crashes**: The correct thing happens, the crash report shows up under the `pd_dev` / `pd_prod` environment corresponding to how the project was built.
- 🔴 **When the app errors** (either by throwing an exception or by `Debug.LogError`) the log message shows up in `production` -- as if I didn't set an environment at all!

### Known workaround

[...]

PR #1964, opened 2025-01-13
fix: iOS native bridge creation with parameters

ysle0:
Fixes an issue reported in https://github.com/getsentry/sentry-unity/issues/1951#issuecomment-2581075951

The original code that starts the iOS native bridge caused
compile error by XCode due to the retaining excessive ownership which can lead to a memory leak.

error name during the compilation: 
"PerformSelector names a selector which retains the object".

this fix with PR was already been created by @bitsandfoxes 
but thanks to him I could create my own PR for the first time.

I really appreciate the community.


bitsandfoxes:
Thanks for contributing! ❤️

bitsandfoxes (approved):

Issue #1608, opened 2024-04-05
Split logging integration from error capturing


bitsandfoxes:
We switched from hooking into [`Application.logMessageReceived`](https://docs.unity3d.com/ScriptReference/Application-logMessageReceived.html) to setting ourselves up as a loghandler via [`Debug.unitylogger.logHandler = this;`](https://github.com/getsentry/sentry-unity/blob/87c17f9746a59ef585d517676e3e557444021fa2/src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs#L49C13-L49C49)
This also means that we have to [format](https://github.com/getsentry/sentry-unity/blob/87c17f9746a59ef585d517676e3e557444021fa2/src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs#L132C58-L132C71) the message ourselves which has a non-insignificant impact on performance

bitsandfoxes:
We switched from `Application.logMessageReceived` to `Debug.unitylogger.logHandler` to get access to the actual exception object to hand that to the IL2CPP backend to provide line number support for caught and uncaught exceptions. We previously had our own string parsing functionality to turn the `string stacktrace` into a `SentryStacktrace`. 

Maybe we can split this into two? Keep relying on the `unitylogger` for actual exceptions but set up the `Application.logMessageReceived` so that messages at least get a stacktrace (even without linenumbers for now). What do you think @munkiki7?

munkiki7:
@bitsandfoxes , if Debug.unitylogger.logHandler is only actually useful for getting the exception, and is heavier in performance, then sure, we can only use it in that one place.

WeiLingQiang:
Is there any conclusion to this discussion? Can we use the method of monitoring `Application.logMessageReceived` instead of using `Debug.unitylogger.logHandler = this;`

Because there will be an extra line of useless information in the printed stack information: `Sentry.Unity.Integrations.UnityLogHandlerIntegration:LogFormat(LogType, Object, String, Object[])`
This is a very bad experience

bitsandfoxes:
Not yet. But I'd really like to see this happen.

bitsandfoxes:
We've not forgotten about this and have it on the roadmap.

PR #1965, opened 2025-01-13
fix: Stacktraces for `Debug.LogError` events

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/1608

This PR restores the functionality of capturing `Debug.LogError` and provides stacktraces in the resulting event.

### Before
Screenshot 2025-01-22 at 11 43 07

### After
Screenshot 2025-01-22 at 12 54 51

### Implementation Details

In this PR, we split the logging integration into two components:
- `UnityLogHandlerIntegration`
- `UnityApplicationLoggingIntegration`

The `UnityLogHandlerIntegration` hooks into the actual log handler and intercepts all logs. This allows us to access the exception object directly in case of `Debug.LogException`, but comes with notable limitations:
- We must call `string.Format` for each log captured as a breadcrumb
- We lose access to the stacktrace string that was previously available through `Application.OnMessageReceived`

To address these limitations, we now implement both approaches:
1. Capturing exceptions via `UnityLogHandlerIntegration`
2. Creating breadcrumbs for logs via `UnityApplicationLoggingIntegration`
3. Capturing `LogErrors` via `UnityApplicationLoggingIntegration`

We create a synthetic exception that parses the stringified stacktrace during capture and generates an event from it.

### Known Limitations

Currently, we cannot provide line numbers for these issues. The [`IL2CPPEventProcessor`](https://github.com/getsentry/sentry-unity/blob/main/src/Sentry.Unity/Il2CppEventProcessor.cs) requires an actually thrown exception to retrieve the necessary images and instruction addresses. In the case of `Debug.LogError`, no such exception exists.


src/Sentry.Unity/Il2CppEventProcessor.cs
```diff
@@ -30,6 +31,13 @@ public void Process(Exception incomingException, SentryEvent sentryEvent)
     {
         Options.DiagnosticLogger?.LogDebug("Running Unity IL2CPP event exception processor on: Event {0}", sentryEvent.EventId);
 
+        // UnityLogException is a synthetic exception created by the LoggingIntegration by parsing the stacktrace provided
+        // to the SDK as a string. It therefor lacks the necessary data to fetch the native stacktrace and go from there
```
vaind, line 35:
```suggestion
        // to the SDK as a string. It therefore lacks the necessary data to fetch the native stacktrace and go from there
```

src/Sentry.Unity/Integrations/UnityErrorLogException.cs
```diff
@@ -0,0 +1,170 @@
...
+                {
+                    frame.ConfigureAppFrame(_options);
+                }
+                frames.Add(frame);
+            }
+
+            return frames;
+        }
+
+        private static SentryStackFrame ParseStackFrame(string stackFrameLine)
```
vaind, line 94: Personally I'd have preferred regex as that would have been easier to read to _me_. But AFAICT, the manual parsing seems correct

src/Sentry.Unity/Integrations/UnityErrorLogException.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            Function = functionName,
+            FileName = null,
+            AbsolutePath = null,
+            LineNumber = null
+        };
+
+        // https://github.com/getsentry/sentry-unity/issues/103
+        private static string StripZeroes(string filename)
+            => filename.Equals("<00000000000000000000000000000000>", StringComparison.OrdinalIgnoreCase)
```
vaind, line 152:
```suggestion
            => filename.Replace("0", "").Equals("<>", StringComparison.OrdinalIgnoreCase)
```

vaind (approved):

src/Sentry.Unity/Integrations/UnityErrorLogException.cs
```diff
@@ -0,0 +1,170 @@
...
+
+                return new SentryStackFrame
+                {
+                    FileName = TryResolveFileNameForMono(filenameWithoutZeroes),
+                    AbsolutePath = filenameWithoutZeroes,
+                    Function = functionName,
+                    LineNumber = lineNo == -1 ? null : lineNo
+                };
+            }
+            catch (Exception)
```
bruno-garcia, line 124: concerning that we catch all and don't have any indication of what went wrong. debug logging at least?

PR #1969, opened 2025-01-15
fix: Flaky iOS smoke tests

bitsandfoxes:
During build we would get warnings of `Too Many Simulators` and then flake out by hitting timeouts. Since we're only setting a lower OS version limit we can specify this destination.

Similarly to the Android test we're also now wrapping the device logs in a group and always output them to have them available (in the event of a timeout or somesuch).

#skip-changelog


.github/workflows/ci.yml
```diff
@@ -11,6 +11,10 @@ env:
   LOWEST_SUPPORTED_UNITY_VERSION: 2019
   DOTNET_CLI_TELEMETRY_OPTOUT: 1
   DOTNET_NOLOGO: 1
+  IOS_MIN_VERSION: "16.1"
+
+variables:
+  IOS_MIN_VERSION: "16.1"
```
bitsandfoxes, line 17:
```suggestion
```

vaind (approved):

PR #1997, opened 2025-01-29
feat: Propagate `traceId` to the Android SDK

bitsandfoxes:
Proof of concept. Java errors share the same `trace ID` as the events coming from the C# layer. Native errors lack the trace context.

Screenshot 2025-01-29 at 20 40 06


src/Sentry.Unity.Android/SentryNativeAndroid.cs
```diff
@@ -132,6 +132,24 @@ public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentry
         options.DiagnosticLogger?.LogInfo("Successfully configured the Android SDK");
     }
 
+    public static void SetTraceId(SentryUnityOptions options)
+    {
+        options.DiagnosticLogger?.LogInfo("Setting Trace ID");
+
+        var traceId = SentrySdk.GetTraceHeader()?.TraceId;
```
bruno-garcia, line 139: Ideally we get a hook from the .NET SDK so when a new trace-id is set there, we are able to pass that down to native.

Similar to `ScopeObserver`. Perhaps it should live there actually.

Because we don't want a single trace id for the lifetime of the game ideally. It should recycle "at some point". Perhaps when a scene loads? We could also reset if session ends/restarts.

src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
```diff
@@ -483,6 +484,9 @@ internal void SetNdkEnabled(bool enableNdk)
     internal void SetNdkScopeSync(bool enableNdkScopeSync)
         => SetMetaData($"{SentryPrefix}.ndk.scope-sync.enable", enableNdkScopeSync.ToString());
 
+    public void SetAutoTraceIdGeneration(bool enableAutoTraceIdGeneration)
```
Flash0ver, line 487: nitpick: Could this method be `internal` for consistency?

Flash0ver (approved):

PR #2009, opened 2025-02-08
Feat: Adds an option to disable capturing Debug.LogError as events

TheCrossBoy:
Currently, the SDK captures any logs sent with `Debug.LogError` as an event. There are some ways to mitigate this, such as by using a `BeforeSend` callback, but the filtering risks over (or under) filtration, and isn't completely clear on how to accomplish this. 

Since there are already options to enable/disable breadcrumb capturing, I've added one to enable/disable the functionality for events too.


bitsandfoxes (approved):
Thanks for the contribution! 🔥

PR #2011, opened 2025-02-11
Add DontDestroyOnLoad gameobjects to the viewhierarchy

TheCrossBoy:
Unable to build and see this locally inside of sentry, but the code to get root GOs works outside of it.


bitsandfoxes:
Hey, thanks again for putting something together to look at. It does in fact work! Crazy enough. But instead of putting it in the SDK I'd much rather provide a callback for users to do that themselves.

PR #2015, opened 2025-02-13
Add support for finding Unity in the Hub alternate install path

shana:
I currently can't build this because my Unity installations are in a custom install path managed by the Unity Hub, so I added some logic for supporting that setting in the `FindUnity` target.

Users can set a custom install path for Unity installations in the Unity Hub, and that information is recorded in a json file in the hub config folder. Older versions of the Hub also recorded a preferred version of Unity to be used, which can be used as a fallback if the requested version isn't available.

This adds support for finding Unity installations in this custom location, and for picking the latest (or preferred, if available) version of Unity that's installed, if the requested one isn't found. The error is changed to output all of the calculated paths that were scanned, instead of hardcoding paths per-platform, as some of these paths can't be known until build time.

Note: I have not run this code outside of Windows yet. This is a variation of what I do in https://github.com/spoiledcat/git-for-unity/blob/main/common/unityreferences.targets, which I know works on all platforms, but I haven't tested on Linux in a while. I guess CI will let me know soon :)
Update: ah, looks like Linux is unhappy. Fixing...
Update 2: Fixed now, I see you're relying on a custom env var set in the docker container to find unity which I missed on the rewrite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## Unreleased
 
-- Add support for finding Unity in the Hub alternate install path ([#2015](https://github.com/getsentry/sentry-unity/pull/2015))
-
 ### Features
 
 - The `Ignore CLI Errors` checkbox in the Debug Symbols tab now applies to all supported platforms. ([#2008](https://github.com/getsentry/sentry-unity/pull/2008))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add support for finding Unity in the Hub alternate install path ([#2015](https://github.com/getsentry/sentry-unity/pull/2015))
+
 ### Features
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add support for finding Unity in the Hub alternate install path ([#2015](https://github.com/getsentry/sentry-unity/pull/2015))
+
```
bitsandfoxes, line 6:
```suggestion
```
The changelog strictly contains user facing changes. Any repo-maintence or utility/CI changes can be annotated with `#skip-changelog` in the issue description to silence `Danger`.

bitsandfoxes (approved):
That's neat! Thanks!

CHANGELOG.md
shana: Ah, gotcha, thanks!

PR #2021, opened 2025-02-14
Add debug symbol uploads for Win32 and MacOS Mono builds

TheCrossBoy:
Moves the debug symbols upload into a `finally` so it's done regardless of what happens with the native support handling.

Also adds cases to upload win32 debug files and the `*.pdb` files from MacOS Mono builds.


bitsandfoxes (approved):

bitsandfoxes:
Thanks for contributing! ❤️

PR #2028, opened 2025-02-20
fix: Updated breadcrumb type for scene changes

bitsandfoxes:
This helps (visually) to distinguish the type of breadcrumbs.

### DIFF
--- a/src/Sentry.Unity/Integrations/SceneManagerIntegration.cs
+++ b/src/Sentry.Unity/Integrations/SceneManagerIntegration.cs
@@ -29,7 +29,8 @@ internal class SceneManagerIntegration : ISdkIntegration
 
             hub.AddBreadcrumb(
                 message: $"Scene '{scene.Name}' was loaded",
-                category: "scene.loaded");
+                category: "scene.loaded",
+                type: "navigation");
         }
 
         void SceneManagerOnSceneUnloaded(SceneAdapter scene)
@@ -42,7 +43,8 @@ internal class SceneManagerIntegration : ISdkIntegration
 
             hub.AddBreadcrumb(
                 message: $"Scene '{scene.Name}' was unloaded",
-                category: "scene.unloaded");
+                category: "scene.unloaded",
+                type: "navigation");
         }
 
         void SceneManagerOnActiveSceneChanged(SceneAdapter fromScene, SceneAdapter toScene)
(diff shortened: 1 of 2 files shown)


src/Sentry.Unity/Integrations/SceneManagerIntegration.cs
```diff
@@ -29,7 +29,8 @@ void OnSceneManagerOnSceneLoaded(SceneAdapter scene, LoadSceneMode mode)
 
             hub.AddBreadcrumb(
                 message: $"Scene '{scene.Name}' was loaded",
-                category: "scene.loaded");
+                category: "scene.loaded",
+                type: "navigation");
```
vaind, line 33: navigation breadcrumbs should also have `data`. From the develop docs:

navigation
A navigation event can be a URL change in a web application, or a UI transition in a mobile or desktop application, etc.
Internally we display breadcrumbs of type default that contain category navigation as a breadcrumb of type navigation.

Its data property has the following sub-properties:

from (Required)

A string representing the original application state / location.

to (Required)

A string representing the new application state / location.

```JSON
{
  "type": "navigation",
  "category": "navigation",
  "timestamp": "2016-04-20T20:55:53.845Z",
  "data": {
    "from": "/login",
    "to": "/dashboard"
  }
}
```

bitsandfoxes: Just thinking out loud here: You can load scenes additionally, how'd that work with the `from`->`to`?

bruno-garcia: could be `current` vs `previous` ? it can be anything that makes sense here

vaind: then, I guess it's not a navigation event

bitsandfoxes: That'd actually make sense. As in: Which scene cause which other scene to load.

bitsandfoxes: Sounds good - doesn't work. The `fromScene` in the `ActiveSceneChanged` call is always `null`. In all Unity versions.

bitsandfoxes:
Closing as `Not Planned`

PR #2029, opened 2025-02-20
fix: Desktop symbol upload

bitsandfoxes:
Followup on https://github.com/getsentry/sentry-unity/pull/2021

In this PR I
- Moved option validation outside of `try-catch` block to bail early
- Symbol upload now happens regardless of native-support settings - to provide line numbers
- Improved logging

#skip-changelog

### DIFF
--- a/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
+++ b/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
@@ -62,7 +62,6 @@ public static class BuildPostProcess
 
         try
         {
-            logger.LogDebug("Adding the crash-handler.");
             AddCrashHandler(logger, target, buildOutputDir, executableName);
         }
         catch (Exception e)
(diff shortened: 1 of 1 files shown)


src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -25,51 +25,55 @@ public static void OnPostProcessBuild(BuildTarget target, string executablePath)
...
+                            "\nSkipping adding crash-handler and uploading debug symbols.", executablePath);
+            return;
+        }
 
-            if (!IsEnabledForPlatform(target, options))
-            {
-                logger.LogDebug("Native support for the current platform is disabled in the configuration.");
-                return;
-            }
+        UploadDebugSymbols(logger, target, buildOutputDir, executableName, options, cliOptions, isMono);
```
bitsandfoxes, line 55: We want to upload the symbols regardless of enabled native support.

PR #2049, opened 2025-03-04
feat: Added `GetLastRunState` to the SentryUnity

bitsandfoxes:
This PR adds a method `GetLastRunState` to the SentryUnity API to allow users to access whether the SDK detects a crash in the game's previous run.

Sample-Usage:
```
public class CrashDetector : MonoBehaviour
{
    void Start()
    {
        // This assumes that the SDK is already initialized. 
        // The SDK has already set up the native scope sync - it will return `Unknown` otherwise
        var lastRunState = SentryUnity.CrashedLastRun();

        switch (lastRunState)
        {
            case SentryUnity.LastRunState.Crashed:
                Debug.Log("The application crashed during the previous run");
                // Perform recovery actions or show user message.
                break;
                
            case SentryUnity.LastRunState.DidNotCrash:
                Debug.Log("The previous run terminated normally");
                break;
                
            case SentryUnity.LastRunState.Unknown:
                // The SDK was either not initialized or does not have access to the native layer
                Debug.Log("Could not determine the state of the previous run");
                break;
        }
    }
}
```

### DIFF
--- a/src/Sentry.Unity/SentryUnitySDK.cs
+++ b/src/Sentry.Unity/SentryUnitySDK.cs
@@ -1,6 +1,5 @@
 using System;
 using System.IO;
-using System.Runtime.InteropServices.ComTypes;
 using System.Threading.Tasks;
 using Sentry.Extensibility;
 using Sentry.Unity.Integrations;
--- a/src/Sentry.Unity/SentryUnitySDK.cs
+++ b/src/Sentry.Unity/SentryUnitySDK.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Runtime.InteropServices.ComTypes;
 using System.Threading.Tasks;
 using Sentry.Extensibility;
 using Sentry.Unity.Integrations;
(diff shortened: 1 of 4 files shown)


src/Sentry.Unity/SentryUnitySDK.cs
```diff
@@ -99,4 +99,18 @@ public void Close()
...
+        if (_options.CrashedLastRun is null)
+        {
+            _options.DiagnosticLogger?.LogDebug("The SDK does not have a 'CrashedLastRun' set. " +
+                                                "This might be due to a missing or disabled native integration.");
+            return SentryUnity.LastRunState.Unknown;
+        }
+
+        return _options.CrashedLastRun.Invoke()
+            ? SentryUnity.LastRunState.Crashed
+            : SentryUnity.LastRunState.DidNotCrash;
```
bitsandfoxes, line 114: We could cache it on some `_options.InternalCrashedLastRun` when setting up the native SDK scope sync instead so users don't incur the cost of calling into the native layer.

bruno-garcia (approved):

PR #2089, opened 2025-03-31
feat: Propagate `traceId` on platforms supported by `sentry-native`

bitsandfoxes:
Propagating the `traceId` when running on platforms supported by `sentry-native`.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+- When running on Windows or Linux, the SDK now links errors and events originating on different layers (managed, native errors) via `trace ID` ([#2089](https://github.com/getsentry/sentry-unity/pull/2089))
```
Flash0ver, line 7: question: Could we merge both entries into a single entry linking both PRs, or do we prefer to have one entry per PR?

src/Sentry.Unity/NativeUtils/CFunctions.cs
```diff
@@ -148,6 +148,9 @@ internal static void SetValueIfNotNull(sentry_value_t obj, string key, double? v
     [DllImport("sentry")]
     internal static extern void sentry_remove_extra(string key);
 
+    [DllImport("sentry")]
+    internal static extern void sentry_set_trace(string traceId, string parentSpanId);
```
Flash0ver, line 152: question: Do the parameter names match the native function, or could/should we change `parentSpanId` to `spanId` for consistency?

Flash0ver (approved):

src/Sentry.Unity/NativeUtils/CFunctions.cs
bitsandfoxes: Good catch, I'll update the call, thanks!

vaind (approved):

PR #2093, opened 2025-04-02
chore: Update to .NET SDK `5.5.0`

bitsandfoxes:


test/Sentry.Unity.Tests/IntegrationTests.cs
```diff
@@ -210,7 +210,7 @@ public IEnumerator DebugLogException_IsMarkedUnhandled()
     {
         yield return SetupSceneCoroutine("1_BugFarm");
 
-        var expectedMechanism = "\"mechanism\":{\"type\":\"Unity.LogException\",\"handled\":false}}]}";
+        var expectedMechanism = "\"mechanism\":{\"type\":\"Unity.LogException\",\"handled\":false";
```
bitsandfoxes, line 213: The `mechanism` now also contains the `,"data":{"HResult":"magicnumber"}`. So we're trimming the check because that's not what we're testing here.

src/Sentry.Unity/ScopeObserver.cs
```diff
@@ -81,6 +81,9 @@ public void SetUser(SentryUser? user)
         }
     }
 
+    public void SetTrace(SentryId traceId, SpanId parentSpanId)
+    { }
```
Flash0ver, line 85: note: will be implemented in follow-up

Flash0ver (approved):

bruno-garcia (approved):

PR #2098, opened 2025-04-07
fix: Android device crash tests

bitsandfoxes:
## Fixes
I'm not exactly sure when or with which version this changes but the crash log switches between 
```
terminating with uncaught exception of type char const*
``` 
and 
```
terminating due to uncaught exception of type char const*
```
Replacing the `with` with `due to`.

Instead of relying on the exact wording from the logs from somewhere within Unity, we're relying on the ones from `sentry-native` (so we can blame ourselves if this breaks in the future).

## Cleanup

With the separation into workflows in https://github.com/getsentry/sentry-unity/pull/1989 we're also no longer exporting a Gradle project that we're then building in a dedicated step. The `IsIntegrationTest` is a remnant from there.

#skip-changelog.

### DIFF
--- a/scripts/smoke-test-android.ps1
+++ b/scripts/smoke-test-android.ps1
@@ -332,8 +332,8 @@ function RunTest([string] $Name, [string] $SuccessString, [string] $FailureStrin
         $logCache = ProcessNewLogs -newLogs $newLogs -lastLogCount ([ref]$lastLogCount) -logCache $logCache
 
         # The SmokeTester logs "SmokeTester is quitting." in OnApplicationQuit() to reliably inform when tests finish running.
-        # For crash tests, we expect to see a native crash log "terminating with uncaught exception of type char const*".
-        if (($newLogs | Select-String "SmokeTester is quitting.") -or ($newLogs | Select-String "terminating with uncaught exception of type char const*"))
+        # For crash tests, we're checking for `sentry-native` logging "crash has been captured" to reliably inform when tests finished running.
+        if (($newLogs | Select-String "SmokeTester is quitting.") -or ($newLogs | Select-String "crash has been captured"))
         {
             Write-Host "Process finished marker detected. Finish waiting for tests to run."
             $processFinished = $true
--- a/scripts/smoke-test-android.ps1
+++ b/scripts/smoke-test-android.ps1
@@ -66,8 +66,8 @@ else
 $TestActivityName = "$ProcessName/com.unity3d.player.UnityPlayerActivity"
 $FallBackTestActivityName = "$ProcessName/com.unity3d.player.UnityPlayerGameActivity"
 
-$_ArtifactsPath = ((Test-Path env:ARTIFACTS_PATH) ? $env:ARTIFACTS_PATH : "./$BuildDir/../test-artifacts/") `
-    + $(Get-Date -Format "HHmmss")
+$_ArtifactsPath = (Test-Path env:ARTIFACTS_PATH) ? $env:ARTIFACTS_PATH : (Join-Path $BuildDir "../test-artifacts/" $(Get-Date -Format "HHmmss"))
+
 function ArtifactsPath
 {
     if (-not (Test-Path $_ArtifactsPath))
--- a/scripts/smoke-test-android.ps1
+++ b/scripts/smoke-test-android.ps1
@@ -22,47 +22,9 @@ Write-Host "#  |___/_|  |_|\___/|_|\_\___| |_| |___|___/  |_|   #"
 Write-Host "#                                                   #"
 Write-Host "#####################################################"
 
-if ($IsIntegrationTest)
-{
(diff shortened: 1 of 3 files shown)


scripts/smoke-test-android.ps1
```diff
@@ -22,52 +22,14 @@ Write-Host "#  |___/_|  |_|\___/|_|\_\___| |_| |___|___/  |_|   #"
 Write-Host "#                                                   #"
 Write-Host "#####################################################"
 
-if ($IsIntegrationTest)
-{
-    $BuildDir = $(GetNewProjectBuildPath)
-    $ApkFileName = "test.apk"
-    $ProcessName = "com.DefaultCompany.$(GetNewProjectName)"
-
-    if ($Action -eq "Build")
```
bitsandfoxes, line 31: We're no longer building with this script. Building happens in https://github.com/getsentry/sentry-unity/blob/main/.github/workflows/android-smoke-test-build.yml

scripts/smoke-test-android.ps1
```diff
@@ -22,52 +22,14 @@ Write-Host "#  |___/_|  |_|\___/|_|\_\___| |_| |___|___/  |_|   #"
...
-    $ApkFileName = "IL2CPP_Player.apk"
-    $ProcessName = "io.sentry.samples.unityofbugs"
-}
+$BuildDir = $(GetNewProjectBuildPath)
+$ApkFileName = "test.apk"
+$ProcessName = "com.DefaultCompany.$(GetNewProjectName)"
 $TestActivityName = "$ProcessName/com.unity3d.player.UnityPlayerActivity"
 $FallBackTestActivityName = "$ProcessName/com.unity3d.player.UnityPlayerGameActivity"
 
-$_ArtifactsPath = ((Test-Path env:ARTIFACTS_PATH) ? $env:ARTIFACTS_PATH : "./$BuildDir/../test-artifacts/") `
```
bitsandfoxes, line 69: This was mixing relative paths `./` with absolute paths `$BuildDir`. Using `Join-Path` to resolve this.

scripts/smoke-test-android.ps1
```diff
@@ -332,8 +294,8 @@ function RunTest([string] $Name, [string] $SuccessString, [string] $FailureStrin
         $logCache = ProcessNewLogs -newLogs $newLogs -lastLogCount ([ref]$lastLogCount) -logCache $logCache
 
         # The SmokeTester logs "SmokeTester is quitting." in OnApplicationQuit() to reliably inform when tests finish running.
-        # For crash tests, we expect to see a native crash log "terminating with uncaught exception of type char const*".
-        if (($newLogs | Select-String "SmokeTester is quitting.") -or ($newLogs | Select-String "terminating with uncaught exception of type char const*"))
+        # For crash tests, we're checking for `sentry-native` logging "crash has been captured" to reliably inform when tests finished running.
+        if (($newLogs | Select-String "SmokeTester is quitting.") -or ($newLogs | Select-String "crash has been captured"))
```
bitsandfoxes, line 298: Using a log we control (from within `sentry-native`) so it won't change with different versions.

test/Scripts.Integration.Test/integration-test.ps1
bitsandfoxes, line 1: This whole script is a utility script to run the scripts used by CI locally.

test/Scripts.Integration.Test/integration-test.ps1
```diff
@@ -82,18 +82,16 @@ If (-not(Test-Path -Path "$(GetNewProjectPath)"))
...
-    If (("iOS", "Android-Export") -contains $Platform)
+    If ("iOS" -eq $Platform)
     {
-        # Workaround for having `exportAsGoogleAndroidProject` remain `false` in Unity 6 on first build
+        # We're exporting an Xcode project and building that in a separate step.
         ./test/Scripts.Integration.Test/build-project.ps1 -UnityPath "$UnityPath" -UnityVersion $UnityVersion -Platform $Platform
-        Remove-Item -Path $(GetNewProjectBuildPath) -Recurse -Force -Confirm:$false
-
-        ./test/Scripts.Integration.Test/build-project.ps1 -UnityPath "$UnityPath" -UnityVersion $UnityVersion -Platform $Platform
-        & "./scripts/smoke-test-$($Platform -eq 'iOS' ? 'ios' : 'android').ps1" Build -IsIntegrationTest -UnityVersion $UnityVersion
```
bruno-garcia, line 96: we don't need the Android stuff here anymore?

bruno-garcia (approved):

test/Scripts.Integration.Test/integration-test.ps1
bitsandfoxes: That script gluing all the scripts together and letting us "run CI locally".
And no, we're not exporting the Gradle project to manually build it on a different runner. We're relying on the Android Dev environment Unity comes with.

Issue #2099, opened 2025-04-07
End-to-End Integration Tests


bitsandfoxes:
## Problem Statement
The current integration test is limited to running the [SmokeTester](https://github.com/getsentry/sentry-unity/blob/main/test/Scripts.Integration.Test/Scripts/SmokeTester.cs)  and crash tests with a mock server that only confirms whether events were sent, without providing visibility into their content.
This creates significant testing gaps. For example, we cannot verify if scope synchronization is functioning correctly. This limitation is particularly evident when trying to observe trace changes introduced in https://github.com/getsentry/sentry-dotnet/pull/4026. 

There currently is no way to validate events coming from different layers of the application.

## Proposal
Rather than depending on the SmokeTester to intercept .NET SDK send-attempts and validate them on the running client, we should enhance the [`crash-test-server`](https://github.com/getsentry/sentry-unity/blob/main/test/Scripts.Integration.Test/crash-test-server.py) to persist envelopes to disk. When executing the [CrashTest](https://github.com/getsentry/sentry-unity/blob/cac2a683f47f549cbf98551ac4c32a58e504d192/test/Scripts.Integration.Test/common.ps1#L75), we should then validate both the presence and contents of these stored envelopes.

### Implementation steps:
1. Start the fake server
2. Execute the game (SmokeTester remains to trigger errors and send events)
3. Server receives and persists events to disk
4. Verify successful game execution
5. Read persisted envelopes and validate their contents against expected values

### Non-exhaustive list of things to check for:
- Reporting SDKs name (`sentry.cocoa.unity`)
- Synched Scope
  - Tags
  - User
  - Breadcrumbs

bruno-garcia:
> As an alternative, a few other SDKs actually post events to sentry.io and then fetch them through an API. Besides resolving the pain points mentioned above, that would also enable testing symbolication.

React Native does this but note: It is (was?) extremely flakey and slow.
So while I'm in favor of having this (I asked for that to be added back then). I wouldn't rely on that for lots of testing. 
It should be just another layer, to have perhaps a single, e2e test. But I not necessarily having lots of test cases added to it.

bitsandfoxes:
Update: Instead of rolling with our on solution this should follow suit with what was done in [https://github.com/getsentry/sentry-unreal/pull/1127](https://github.com/getsentry/sentry-unreal/pull/1127/files)

bitsandfoxes:
The SDK runs e2e integration tests via app-runner, see
[...]

PR #2107, opened 2025-04-10
fix: Async Android scope sync

bitsandfoxes:
## Problem Statement

The issue manifested in two places: 
1. Customer conversation at the booth with complaints that heavy use of logging is killing their games performance on Android.
2. CI being flaky a lot with logs showing `JNI execution timed out`.

## Analysis

After doing some naive and simple benchmarking, by logging 100 messages within one frame we get to `1.1707766s` spent on that frame.

When looking at the `ScopeObserver` for Android we - i.e. `AddBreadcrumb`

https://github.com/getsentry/sentry-unity/blob/16e21d8a6e7d70ca74ce3083e8eddea78bcc9f79/src/Sentry.Unity.Android/AndroidJavaScopeObserver.cs#L23-L25

see that we end up on the `JniExecutor`

https://github.com/getsentry/sentry-unity/blob/16e21d8a6e7d70ca74ce3083e8eddea78bcc9f79/src/Sentry.Unity.Android/JniExecutor.cs#L128

that makes blocking calls into the JNI with a timeout

https://github.com/getsentry/sentry-unity/blob/16e21d8a6e7d70ca74ce3083e8eddea78bcc9f79/src/Sentry.Unity.Android/JniExecutor.cs#L140

that easily gets hit

https://github.com/getsentry/sentry-unity/blob/16e21d8a6e7d70ca74ce3083e8eddea78bcc9f79/src/Sentry.Unity.Android/JniExecutor.cs#L144

### Considerations

1. Scope sync should not block or slow down the game.
2. A regular log message eating up almost an entire frame to sync is unacceptable.
3. There is a `maxBreadcrumb` limit in the first place. So potentially blocking the game just to have them synched down a layer is incurring all the cost for limited benefit.
4. In case there is a crash happening while the scope is still in process of being synched: They are likely to time out right now anyway. 
5. Limiting the async scope sync to `AddBreadcrumb` for now. This happens under the assumption that any changes to the scope at runtime is triggered by the user and should end up on the native layer.

## Resolution

After running some benchmarks I realized that the whole worker-thread implementation is causing way more overhead than it was doing any good. Instead, the `JniExecutor` is now merged into `SentryJava` since there was nothing left to du but handle `attaching` and `detaching` to the `AndroidJNI`. This has to happen depending on whether we're running on the main-thread or not. Detaching from a still code executing thread will cause a crash.

1. Move everything related to calling into Java into `SentryJava`. - Instead of having the scope observer call the `JniExecutor` on it's own.
2. Properly handle attaching/detaching to the JNI

### Benchmarks

| Runs | Metric | Baseline | Old | New | New vs Old (Improvement) | Baseline vs New |
|------|--------|----------|-----|-----|--------------------------|----------------|
| 1 | Median | 0.01ms | 8.32ms | 0.10ms | -8.22ms (99%) | +0.09ms |
| 10 | Median | 0.01ms | 104.92ms | 0.59ms | -104.33ms (99%) | +0.58ms |
| 50 | Median | 0.06ms | 544.89ms | 2.16ms | -542.73ms (99%) | +2.10ms |
| 100 | Median | 0.08ms | 1095.85ms | 3.13ms | -1092.72ms (99%) | +3.05ms |
| 1000 | Median | 0.62ms | 11079.85ms | 16.83ms | -11063.02ms (99%) | +16.21ms |

Baseline = No sync to the Java SDK
Old = JNI worker thread synchronization
New = This PR

### Follow Ups

- Completeness of scope on native events is currently untested and should be part of https://github.com/getsentry/sentry-unity/issues/2099


src/Sentry.Unity.Android/JniExecutor.cs
```diff
@@ -8,150 +9,157 @@ namespace Sentry.Unity.Android;
...
+    /// <summary>
+    /// Runs a JNI operation that returns a result, waiting for completion with a timeout
+    /// </summary>
+    public TResult? Run<TResult>(Func<TResult?> jniOperation, TimeSpan? timeout = null)
+    {
+        timeout ??= DefaultTimeout;
+        var completionSource = new TaskCompletionSource<object?>();
+        var operation = new JniOperation(jniOperation, completionSource);
+
+        _operationQueue.Enqueue(operation);
```
bruno-garcia, line 110: what's the queue capacity? if it's full, it blocks here too

bitsandfoxes: Fixed.

.github/workflows/ci.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           - target: Linux
             # Build using older Linux version to preserve sdk compatibility with old GLIBC
             # See discussion in https://github.com/getsentry/sentry-unity/issues/1730 for more details
-            host: ubuntu-20.04
+            host: ubuntu-22.04
```
bitsandfoxes, line 37: This is a unrelated and temporary change to unblock CI in this PR.

src/Sentry.Unity.Android/AndroidOptionConfiguration.cs
```diff
@@ -0,0 +1,36 @@
...
+            if (methodName != "configure" || args.Length != 1)
+            {
+                throw new Exception($"Invalid invocation: {methodName}({args.Length} args)");
+            }
+
+            _callback(args[0]);
+        }
+        catch (Exception e)
+        {
+            _logger?.LogError(e, "Error invoking {0} ’.", methodName);
```
bruno-garcia, line 32:
```suggestion
            _logger?.LogError(e, "Error invoking '{0}'.", methodName);
```

src/Sentry.Unity.Android/JniExecutor.cs
```diff
@@ -1,182 +1,58 @@
...
-        _taskEvent = new AutoResetEvent(false);
-        _shutdownSource = new CancellationTokenSource();
-
-        _workerThread = new Thread(DoWork) { IsBackground = true, Name = "SentryJniExecutorThread" };
-        _workerThread.Start();
+        _androidJNI ??= androidJNI ?? AndroidJNIAdapter.Instance;
     }
 
-    private void DoWork()
+    public TResult? Run<TResult>(Func<TResult?> jniOperation, bool? isMainThread = null)
```
bruno-garcia, line 17: is the argument here for testing purposes? I imagine we can't trust the caller on whether it's UI thread or not?

src/Sentry.Unity.Android/JniExecutor.cs
```diff
@@ -1,182 +1,58 @@
...
-            catch (Exception e)
-            {
-                _logger?.LogError(e, "Error during JNI execution.");
-                return default;
-            }
-            finally
-            {
-                _currentTask = null!;
-            }
+            return jniOperation.Invoke();
```
bruno-garcia, line 27: We throw from this, right? e.g; when method doesn't exist? 
Where do we handle those exceptions?

src/Sentry.Unity.Android/JniExecutor.cs
```diff
@@ -1,182 +1,58 @@
...
-            finally
-            {
-                _currentTask = null!;
+                _androidJNI.DetachCurrentThread();
             }
         }
     }
 
-    public void Dispose()
+    public void Run(Action jniOperation, bool? isMainThread = null)
```
bruno-garcia, line 38: same here, for testing?

src/Sentry.Unity.Android/JniExecutor.cs
```diff
@@ -1,182 +1,58 @@
...
+            _androidJNI.AttachCurrentThread();
         }
 
-        _isDisposed = true;
-
-        _shutdownSource.Cancel();
         try
         {
-            _workerThread?.Join(100);
+            jniOperation.Invoke();
```
bruno-garcia, line 48: where do we handle exceptions?

src/Sentry.Unity.Android/ScopeCallback.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            if (methodName != "run" || args.Length != 1)
+            {
+                throw new Exception($"Invalid invocation: {methodName}({args.Length} args)");
+            }
+            _callback(args[0]);
+        }
+        catch (Exception e)
+        {
+            // Adding the Sentry logger tag ensures we don't send this error to Sentry.
```
bruno-garcia, line 33: do we send our own diagnostic logger stuff to Sentry? We shouldn't

bruno-garcia (approved):
Does the device test cover any of this?
are sync'ing scope, and do we verify things got passed down?

bitsandfoxes:
> Does the device test cover any of this? are sync'ing scope, and do we verify things got passed down?

Device tests make sure that the app does not crash. To verify the sync itself we need #2099

src/Sentry.Unity.Android/JniExecutor.cs
bitsandfoxes: Updated by splitting public/interal for testing purposes.

PR #2109, opened 2025-04-11
feat: Serverside integrationtests

bitsandfoxes:
Relates to https://github.com/getsentry/sentry-unity/issues/2099

#skip-changelog


bitsandfoxes:
This will be picked up with https://github.com/getsentry/sentry-unity/issues/2099 and use a broader solution by utilizing https://github.com/getsentry/app-runner.git

PR #2112, opened 2025-04-15
fix: Reporting SDK's name on Apple platforms

bitsandfoxes:
Since we're not building the Cocoa SDK we have to set the name ourselves. When targeting iOS and configuring the Cocoa SDK at build-time, this is already implemented:
https://github.com/getsentry/sentry-unity/blob/3add276b21ee666aedc3ab2264cbf125b4a605bf/src/Sentry.Unity.Editor.iOS/NativeOptions.cs#L24

This was missing when setting `RuntimeInitialization`, or when running on macOS in general.

#skip-changelog


bruno-garcia (approved):

bruno-garcia:
Any other SDK that changed for submodule to dep needs to do this? Like .NET or Unreal?
cc @tustanivsky

tustanivsky:
We're already doing similar thing for Unreal:
https://github.com/getsentry/sentry-unreal/blob/42f1fbf6e506e113ab8df8ed3870c8afd62d6f8d/plugin-dev/Source/Sentry/Private/Apple/AppleSentrySubsystem.cpp#L41

PR #2113, opened 2025-04-16
fix: Bump Ubuntu runner building the SDK for Linux

bitsandfoxes:
#skip-changelog


bitsandfoxes:
Maybe this works, but it's not the best idea as we're going to run into the very same issue once Ubuntu 22 goes the way of the dodo.

PR #2118, opened 2025-04-17
fix: use ubuntu-20.04 container

### LINKED ISSUE #2114: Dockerize building `sentry-native` for Linux
## Problem

Github no longer supports `Ubuntu 20.04` runners, causing us to not be able to release without potentially breaking user's builds when running their game/server on older Ubuntu versions.

## Context

We're currently building `sentry-native` as part of our CI (and package & release process).
For Linux, we've pinned the runner version to `Ubuntu 20.04` which is [no longer supported](https://github.com/actions/runner-images/issues/11101). We were building on the oldest available version to maintain compatibility.

https://github.com/getsentry/sentry-unity/blob/3add276b21ee666aedc3ab2264cbf125b4a605bf/.github/workflows/ci.yml#L34-L37

We're already fixing and installing additional Linux dependencies as part of the build pipeline.

The actual building happens from within a MSBuild target

https://github.com/getsentry/sentry-unity/blob/3add276b21ee666aedc3ab2264cbf125b4a605bf/Directory.Build.targets#L303-L319

## Possible Solutions

Dockerize the build process for `sentry-native` on Linux.

jpnurmi:
Uses `dotnet-install.sh` from https://github.com/dotnet/install-scripts.

#skip-changelog

Close #2114


scripts/build-linux-sdk.sh
bitsandfoxes, line 1: I'm sorry, I did not think this through in my comment. 
Now the build & copy logic moved into a bash-script. How much effort would it be to add dotnet to the docker and leave the rest as-is?

jpnurmi: Not much effort at all because I already experimented with that idea yesterday :wink:

jpnurmi: I pushed the `ubuntu:20.04` + `dotnet-install.sh` variant that feels nicest out of the discussed options. What do you think?

bitsandfoxes: It's so beautiful, it's frightening! Thanks a lot!

bitsandfoxes (approved):
Awesome! Thanks a lot!

PR #2123, opened 2025-04-17
feat: Trace Generation Integration

bitsandfoxes:
This PR adds an integration that is responsible for generating a new trace based 
on already existing mechanisms using in the [`SceneManagerIntegration`](https://github.com/getsentry/sentry-unity/blob/main/src/Sentry.Unity/Integrations/SceneManagerIntegration.cs) and the [`SessionIntegration`](https://github.com/getsentry/sentry-unity/blob/main/src/Sentry.Unity/Integrations/SessionIntegration.cs)
- Active scene changes
- When the app gains focus

In practical terms this translates into the following flow:
1. App starts up
2. SDK auto-initializes
3. `TraceGenerationIntegration` gets registered and starts a new trace
4. The rest of the startup procedure (i.e. subsystem registration, splash-screen e.t.c.)
5. First scene load
6. Active Scene changes - `TraceGenerationIntegration` kicks in and creates a new trace

Logically, this separates the startup from the actual game. During gameplay the trace gets re-generated every time the game gains focus. This can/might/should/could get updated in the future by different lifecycle events. I.e. touches on mobile (like Android does it).


src/Sentry.Unity/Integrations/SessionIntegration.cs
```diff
@@ -19,8 +19,6 @@ public void Register(IHub hub, SentryOptions options)
             return;
         }
 
-        options.DiagnosticLogger?.LogDebug("Registering Session integration.");
```
bitsandfoxes, line 22: The [hub already logs](https://github.com/getsentry/sentry-dotnet/blob/e72ca08d3b0d25cf09690090cd09187c235922b1/src/Sentry/Internal/Hub.cs#L83) which integration get registered. This is redundant.

src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -4,11 +4,16 @@
 
 namespace Sentry.Unity;
 
+internal interface ISentryMonoBehaviour
```
bitsandfoxes, line 7: I'm creating an internal interface to be able to mock this in the tests.

src/Sentry.Unity/SentryUnitySDK.cs
```diff
@@ -51,11 +51,6 @@ private SentryUnitySdk(SentryUnityOptions options)
 
         unitySdk._dotnetSdk = SentrySdk.Init(options);
 
-        // For now, we're creating a new trace after initializing to be able to tie errors and crashes together on all
-        // layers. To be able to regenerate new traces based on some mechanism, this will move into some sort of
-        // integration i.e. scene manager.
-        SentrySdk.SetTrace(SentryId.Create(), SpanId.Create());
-
```
bitsandfoxes, line 58: This is the part that is getting replaced by the integration.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Features
 
+- The trace used to connect errors on different layers of your game now gets regenerated every time the aspp gains focus again, or the active scene changes ([#2123](https://github.com/getsentry/sentry-unity/pull/2123))
```
Flash0ver, line 11: issue: typo
```diff
-..every time the aspp gains focus again..
+..every time the app gains focus again..
```

src/Sentry.Unity/Integrations/TraceGenerationIntegration.cs
```diff
@@ -0,0 +1,36 @@
+using Sentry.Extensibility;
+using Sentry.Integrations;
+
+namespace Sentry.Unity.Integrations;
+
+internal class TraceGenerationIntegration : ISdkIntegration
```
Flash0ver, line 6: nitpick: Per default, I like to make new types `sealed`, especially when implementing interfaces and/or overriding members from base classes

There are - in some cases - also performance benefits that the .NET Runtime can utilize. But I'm uncertain if that has any effect after IL2CPP. And even if there was, `Register` is only called once per lifetime of the process. So I guess, in this case, especially for `internal` types, this is more of a preference/opinion. Just as with all my `nitpick` comments: feel free to ignore it.

test/Sentry.Unity.Tests/TraceGenerationIntegrationTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        sut.Register(_fixture.TestHub, _fixture.SentryOptions);
+
+        // Act
+        _fixture.SentryMonoBehaviour.ResumeApplication();
+
+        // Assert
+        var configureScope = _fixture.TestHub.ConfigureScopeCalls.Single();
+        var scope = new Scope(_fixture.SentryOptions);
+        var initialPropagationContext = scope.PropagationContext;
+        configureScope(scope);
```
Flash0ver, line 70: question: Why do we need to invoke the delegate manually?

Isn't `hub.ConfigureScope` meant to already invoke the `Action<Scope>`? Or is this expected through the `TestHub`?

src/Sentry.Unity/Integrations/TraceGenerationIntegration.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    internal TraceGenerationIntegration(ISentryMonoBehaviour sentryMonoBehaviour, ISceneManager sceneManager)
+    {
+        _sceneManager = sceneManager;
+        _sentryMonoBehaviour = sentryMonoBehaviour;
+    }
+
+    public void Register(IHub hub, SentryOptions options)
+    {
+        hub.ConfigureScope(scope =>
```
Flash0ver, line 22: question: Before this update, I was wondering if we may run into a timing issue when the `ActiveSceneChanged` has already been raised (or is never going to be raised if no scene is ever going to change for the game ... I am unfamiliar with Unity's lifetime events) before we are going to _Register_ this _SDK Integration_.

Is this line covering above scenario, or am I getting something wrong?

test/Sentry.Unity.Tests/TraceGenerationIntegrationTests.cs
bitsandfoxes: The `TestHub` does not invoke the actions. It captures the delegates so we can validate them later.
https://github.com/getsentry/sentry-unity/blob/f0e68a5a9c5d00216d557aac9ee3ad131ce88bc2/test/Sentry.Unity.Tests/Stubs/TestHub.cs#L83

src/Sentry.Unity/Integrations/TraceGenerationIntegration.cs
bitsandfoxes: The `ActiveSceneChanged` might or might not be called many times over the game's lifetime. It is getting called at least once, during startup.

In case of user initializing themselves, we're "missing" out on the first `sceneChange` but then the call to `Register` covers that, making sure that there's a proper trace present.

src/Sentry.Unity/Integrations/TraceGenerationIntegration.cs
```diff
@@ -0,0 +1,39 @@
...
+            scope.SetPropagationContext(new SentryPropagationContext(SentryId.Create(), SpanId.Create())));
```
bruno-garcia, line 23: This constructor feels like a lot of boiler plate. In fact even having to pass a new propagation context feels redundant as it seems it's the main use case we have, just setting a new propagation context.

Could also be solved with an extension method like `ResetPropagationContext()` that calls `SetPropagationContext(new SentryPropagationContext(SentryId.Create(), SpanId.Create())))`

bruno-garcia (approved):

Issue #2057, opened 2025-03-06
When building for iOS, if the "Ignore CLI Errors" option is on, but sentry-cli produces stderr output, the build will fail


TheCrossBoy:
### Environment

How do you use Sentry?
Sentry SaaS

Which version of the SDK?
4.0.1

How did you install the package? (Git-URL, Assetstore)
Git

Which version of Unity?
2022.3.23f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Editor

### Steps to Reproduce

1. Install the Unity SDK and target iOS with native support enabled
2. Enable the "Ignore CLI Errors" option
3. Setup the configuration in a way that the sentry-cli output will print to stderr (in my case, it was by having a valid auth token for an organization that had its name changed, more details in <insert issue>)
4. Build

### Expected Result

The CLI produces an error without causing the build itself to fail.

### Actual Result

The build runs, sentry-cli fails, and the output is something like this:
```
error: API request failed

Caused by:
    sentry reported an error: not found (http status: 404)

Command failed, however, "SENTRY_ALLOW_FAILURE" variable or "allow-failure" flag was set. Exiting with 0 exit code.
Command PhaseScriptExecution emitted errors but did not return a nonzero exit code to indicate failure
```

### Any logs or screenshots
See above.

bruno-garcia:
Thanks for reporting, indeed seems like a bug.

> Command PhaseScriptExecution emitted errors but did not return a nonzero exit code to indicate failure

I wonder if there's something with xcode in considering the step a failure just because `stderr` was written to? It should just respect the fact the CLI returned `0`.

bitsandfoxes:
I gave this a try but no luck. The moment the logs make it back into Xcode, the error registers.

I tried adding `export XCODE_IGNORE_SCRIPT_OUTPUT_ERRORS=1` to the script. That does not seem to do anything.
I tried logging to a file and then simply printing those logs - no luck.
```
export SENTRY_PROPERTIES=sentry.properties
export NSUnbufferedIO=YES
export XCODE_IGNORE_SCRIPT_OUTPUT_ERRORS=1

echo ""Uploading debug symbols and bcsymbolmaps...""
LOG_FILE=""$BUILT_PRODUCTS_DIR/../sentry_upload.log""

(./{0} debug-files upload {1} $BUILT_PRODUCTS_DIR > ""$LOG_FILE"" 2>&1 || true)

[...]

PR #2136, opened 2025-04-30
fix: Ignor CLI errors when targeting `iOS`

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/2057

In my opinion, this flag is supposed to allow users to unblock themselves. I.e. automated build systems somewhere, local dev environment and is accessible via the configure cli options callback. 
Since the user explicitly opted in to ignoring CLI errors, I think it's fine to simply log into a file.

The difference might seem minor but the build no longer errors with 
```
Command PhaseScriptExecution emitted errors but did not return a nonzero exit code to indicate failure
```

Before:
Screenshot 2025-04-30 at 17 51 17

After:
Screenshot 2025-04-30 at 17 50 29

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- When targeting iOS and setting `IgnoreCliErrors = true`, the Xcode build will now succeed even if the symbol upload itself failed. This is aimed to allow users to unblock themselves ([#2136](https://github.com/getsentry/sentry-unity/pull/2136)) 
+
 ## 3.2.0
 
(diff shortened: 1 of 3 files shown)


src/sentry-dotnet
bruno-garcia, line 1: You changed the .NET SDK  version accidently here

bruno-garcia (changes requested):
Please revert the submodule change, other than that :shipit:

src/sentry-dotnet
bitsandfoxes: Thanks for the catch!

bruno-garcia (approved):

Issue #1845, opened 2024-10-14
Releases cause 2 branches to be created (CI runs then twice)


bruno-garcia:
Something odd with our release process. It seems we end up with 2 release branches each time.

`release/version` and just `version`, e.g:

https://github.com/getsentry/sentry-unity/actions?page=3

![Image](https://github.com/user-attachments/assets/06a7955a-6a40-49fa-96f3-7c1f75753253)

PR #2157, opened 2025-05-14
chore: Specify what CI runs on

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/1845

The CI run described in the issue is not a branch but the `tag` created during releasing. Specifying the branches  should automatically exclude tags.

#skip-changelog


vaind (approved):

bruno-garcia:
nice!!

We should make sure this is done right in all other SDKs. cc @tustanivsky @limbonaut

PR #2163, opened 2025-05-16
fix: Disable screenshot capture in Editor

bitsandfoxes:
Capturing screenshot in the Editor is flaky and often a bad user experience when looking through issues. Camera scale inside the Player causes all kinds of artifacts and the benefit of reporting these is minimal. Users typically have the Editor open when running into errors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: "CI"
 on:
   push:
+    branches:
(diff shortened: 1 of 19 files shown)


bruno-garcia:
CI seems completely broken

bruno-garcia (approved):
Approving but please check CI, and add a changelog entry

Issue #2164, opened 2025-05-18
SIGABORT in Android NDK in version 3.2.0


mizunokazumi:
### Environment

Sentry SaaS (sentry.io)
UPM: https://github.com/getsentry/unity.git#3.2.2
Unity: 2022.3.24f1
SIGABRT in Android

### Steps to Reproduce

1. install sentry-unity 3.2.2
2. install Applovin MAX SDK (`"com.applovin.mediation.ads": "8.0.1"`)
3. build and run

Mandatory conditions:
1. sentry-unity > 3.1.0 (In 3.1.0 everything is fine)
2. Native Android support is enable (NDK integration and sync scope — does not affect and can be in any state)
3. il2cpp (no mono)

### Expected Result

no crash

### Actual Result

SIGABRT occurred and the application crashed

### Any logs or screenshots

```
2025/05/18 22:08:42.362 29729 30637 Fatal om.sample.tes runtime.cc:2585] Thread[112,tid=30637,Native,Thread*=0xb400007dcfc87400,peer=0x1652eef0,"AppLovinSdk:Max-Unity-Plugin:shared"] attempting to detach while still running code
2025/05/18 22:08:42.865 29729 30637 Fatal om.sample.tes runtime.cc:708] Runtime aborting...
2025/05/18 22:08:42.865 29729 30637 Fatal om.sample.tes runtime.cc:708] Dumping all threads without mutator lock held
2025/05/18 22:08:42.865 29729 30637 Fatal om.sample.tes runtime.cc:708] All threads:
2025/05/18 22:08:42.865 29729 30637 Fatal om.sample.tes runtime.cc:708] DALVIK THREADS (115):
2025/05/18 22:08:42.865 29729 30637 Fatal om.sample.tes runtime.cc:708] "main" prio=10 tid=1 Native
2025/05/18 22:08:42.865 29729 30637 Fatal om.sample.tes runtime.cc:708]   | group="" sCount=1 ucsCount=0 flags=1 obj=0x71ebbeb0 self=0xb400007eee501800
2025/05/18 22:08:42.865 29729 30637 Fatal om.sample.tes runtime.cc:708]   | sysTid=29729 nice=-10 cgrp=default sched=0/0 handle=0x7eefade4f8
2025/05/18 22:08:42.865 29729 30637 Fatal om.sample.tes runtime.cc:708]   | state=S schedstat=( 1232492799 257392494 1293 ) utm=105 stm=17 core=2 HZ=100
[...]

PR #2165, opened 2025-05-19
fix: Don't overwrite `MainThreadId` 

bitsandfoxes:

## The Issue

Based on https://github.com/getsentry/sentry-unity/issues/2164 the SDK seems to run into a race condition with `IsMainThread()`. 

I might have been able to reproduce it locally, once by:
1. Initializing a "plugin" with `RuntimeInitializeOnLoadMethod` as well.
2. What gets initialized first based on 1. - the SDK or my repro-plugin - is undefined
3. `MainThreadData.CollectData` gets called twice. Once by `AndroidNative.Configure`, once by `SentyUnity.Init`
4. The ScopeObserver is already setup, `IsMainThread` falsely resolves to `false` and `SentryJava` attempts to detach from the JNI

My assumption is that calling multiple calls to `CollectData` assign and overwrite the nullable `MainThreadId` and the non-atomic check
```
public static bool IsMainThread()
    => MainThreadId.HasValue && Thread.CurrentThread.ManagedThreadId == MainThreadId;
```

creates a race condition even if the same integer is getting set.

### The `MainThreadId` should not be overwritten once set

The current best practice for libraries to handle `IsMainThread` is to fetch the current thread ID at a place where it is guaranteed to be the main thread. Those are methods like `Awake`, `Start`, and the `RuntimeInitializeOnLoadMethod`.

On Android, when setting up the ScopeObserver, the SDK already requires `IsMainThread`. Since we know it's getting called through the `SentryInitialization`, we keep collecting it there.

The subsequent call during `Init` does not overwrite it.

### Make `IsMainThread` nullable && `SentryJava` update for `IsMainThread`

In case of user manually setting up or initializing, the `SentryJava` needs to be able to be certain whether it is running on the main thread or not. A false positive and the attempt to detach on the main thread crashes the game.


limbonaut (commented):
I'm not familiar with the sentry-unity codebase, so take it with a grain of salt. I fail to see a fix for the issue. Is making IsMainThreadId nullable somehow does it? I'm not entirely sure how overwriting an integer creates a racing condition in the first place. It's the same integer and writing int should be atomic, no? Read-write can be a racing condition, but in this case, we're writing the same integer value and not computing anything. From the problem statement, it seems that MainThreadId is probably not initialized in time, right? I wonder if it can be initialized even earlier.

src/Sentry.Unity/MainThreadData.cs
```diff
@@ -69,8 +68,16 @@ internal static class MainThreadData
 
     public static DateTimeOffset? StartTime { get; set; }
 
-    public static bool IsMainThread()
-        => MainThreadId.HasValue && Thread.CurrentThread.ManagedThreadId == MainThreadId;
+    public static bool? IsMainThread()
+    {
+        if (MainThreadId.HasValue)
```
limbonaut, line 73: Since it’s an integer that should be set once and visible across threads, I think declaring it as volatile makes sense:
```csharp
internal static volatile int _mainThreadId;
```

If you really need to make it nullable, then this should also work:
```csharp
private static int? _mainThreadId;

internal static int? MainThreadId
{
    get => Volatile.Read(ref _mainThreadId);
    set => Volatile.Write(ref _mainThreadId, value);
}
```

jpnurmi: What about [`Interlocked`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked)? (just learned about it from @Flash0ver yesterday :grin:)

limbonaut: If I understood correctly, the problem could be in thread visibility. Atomic operations should also work, but probably unnecessary.

src/Sentry.Unity/MainThreadData.cs
```diff
@@ -79,7 +86,9 @@ public static void CollectData()
     {
         var sentrySystemInfo = SentrySystemInfo ?? SentrySystemInfoAdapter.Instance;
 
-        MainThreadId = sentrySystemInfo.MainThreadId;
+        // Don't overwrite the MainThreadId if it's already been set
+        MainThreadId ??= sentrySystemInfo.MainThreadId;
```
vaind, line 90: This operation is not atomic. How about using `Interlocked.CompareExchange`, with non-nullable mainThreadId that defaults to -1 in which case `IsMainThread` returns null, or even better, change the return value to an enum with "unknown" as one of the values if you care about that. If it doesn't matter, it can just return false until it's known.

vaind (changes requested):

bitsandfoxes:
This is not the fix for the issue linked.

PR #2178, opened 2025-05-28
ci: use envs in GHAs and pin versions

mdtro:
- Running these workflows is gated pretty well, but this mitigates the potential for a script injection attack by passing the input to an intermediary environment variable first.
  - See https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#example-of-a-script-injection-attack for more details.
- Using version tags or no specifier at all can open us up to dependency attacks. Pinning the SHA is a strong mitigation. 
  - See: https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Download UPM package
-        uses: vaind/download-artifact@e7141b6a94ef28aa3d828b52830cfa1f406a1848
+        uses: vaind/download-artifact@e7141b6a94ef28aa3d828b52830cfa1f406a1848 # v4-with-wait-timeout
         with:
           name: ${{ github.sha }}
           wait-timeout: 3600
@@ -118,7 +118,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Free Disk Space (Ubuntu)
-        uses: jlumbroso/free-disk-space@f68fdb76e2ea636224182cfb7377ff9a1708f9b8
+        uses: jlumbroso/free-disk-space@f68fdb76e2ea636224182cfb7377ff9a1708f9b8 # v1.3.0
         with:
           android: true
           dotnet: false
@@ -149,7 +149,7 @@ jobs:
           docker exec --user root unity apt-get -y -q install mono-devel
 
       - name: Download IntegrationTest project
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
         with:
           name: smoke-test-${{ matrix.unity-version }}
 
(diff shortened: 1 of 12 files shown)


mdtro:
> @mdtro I noticed you updated quite a few variable references from double curly brace to single curly brace e.g `${{ inputs.unity-version }}'` to `"${UNITY_VERSION}"` Is that on purpose?

Yes, it is intentional. With double curly brackets `${{...}}` it is still using GHA's template expansion. Instead, those values are set in the `env: ` clauses at the top of the workflow, job, or particular step -- depending on where they're needed.

Using single brackets `${VARNAME}` means the environment variable is used and the shell itself does the variable expansion.

.github/workflows/smoke-test-create.yml
```diff
@@ -16,22 +16,24 @@ jobs:
...
       with:
         registry: ghcr.io
-        username: ${{ github.actor }} 
-        password: ${{ secrets.GITHUB_TOKEN }} 
+        username: ${{ env.GITHUB_ACTOR }}
+        password: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Start the Unity docker container
-      run: ./scripts/ci-docker.sh '${{ inputs.unity-version }}' 'base' '${{ secrets.UNITY_LICENSE_SERVER_CONFIG }}'
+      run: ./scripts/ci-docker.sh "${{ env.UNITY_VERSION }}" 'base' '${{ secrets.UNITY_LICENSE_SERVER_CONFIG }}'
```
limbonaut, line 36:
```suggestion
      run: ./scripts/ci-docker.sh "${UNITY_VERSION}" 'base' '${{ secrets.UNITY_LICENSE_SERVER_CONFIG }}'
```
Probably, missed this one.

.github/workflows/ci.yml
```diff
@@ -376,7 +376,7 @@ jobs:
         run: ./test/Scripts.Integration.Test/build-project.ps1 -UnityPath "${{ env.UNITY_PATH }}"
```
limbonaut, line 376:
```suggestion
        run: ./test/Scripts.Integration.Test/build-project.ps1 -UnityPath "${env:UNITY_PATH}"
```
I think this one also needs to be changed.

.github/workflows/ci.yml
```diff
@@ -160,7 +160,7 @@ jobs:
       #   run: ./test/Scripts.Integration.Test/build-project.ps1 -UnityPath "${{ env.UNITY_PATH }}" -Platform "${{ matrix.platform }}"
```
limbonaut, line 160:
```suggestion
      #   run: ./test/Scripts.Integration.Test/build-project.ps1 -UnityPath "${env:UNITY_PATH}" -Platform "${{ matrix.platform }}"
```
Just in case someone uncomments this.

PR #2198, opened 2025-06-18
feat: Add support for Nintendo Switch

bitsandfoxes:
We'll need to find a way to deal with statically built sentry-native.

Follow up on fixing the installation instructions in https://github.com/getsentry/sentry-switch/pull/26

Todos:

- [x] Import sentry-switch
- [x] Load static lib
- [ ] `AnalyticsSessionInfo::get_userId` is not valid
- [ ] Log Forwarder throws with: `Unable to load 'msvcrt'`
- [ ] Cannot start network transport because `network_connect_func option is not set`
- [ ] No support for IL2CPP line numbers: stacktrace length does not match
- [ ] Debug symbol upload? There is this `.nss` file
- [ ] Add `capture_message` to the native samples
- [ ]


bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-unity/pull/2503

Issue #2181, opened 2025-05-29
Sentry 3.2.0, 3.2.1 and 3.2.2 cause a crash on certain nvidia platforms


paveldjatlow:
### Environment
Nvidia gefore rtx 3070 + latest drivers

How did you install the package? (Git-URL, Assetstore)
Add package

Which version of Unity?
6000.0.47f1

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?
Build

### Steps to Reproduce

1. Install Sentry 3.2.0
2. Load a scene (can send a test build by request)
3. The game crashes with EXCEPTION_ACCESS_VIOLATION_READ / 0x58
4. Note: game does NOT crash with 3.1.0 and earlier

### Any logs or screenshots

![Image](https://github.com/user-attachments/assets/0aaaf121-2e1b-4825-9df7-9837d422a650)

Captured crash for reference: https://white-ink.sentry.io/issues/43477493/?project=4507232130105424&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=3

By the native debug callstack, that's the name of the crashing dll (sentry.dll):
https://github.com/getsentry/sentry-unity/blob/cc4a46795ad34621bbf870ef3d722cebe62c4dcd/src/Sentry.Unity/NativeUtils/CFunctions.cs#L80

bruno-garcia:
Thanks for reporting this. 

Does it crash in launch?
any specific code that triggers it? thinking how we can repro it in our side.

pdyatlov:
The crash comes from the nvidia driver, I'm not sure if it's related to code at all. I have the game on Steam, I can upload it to a testing branch and give you the password so you can repro it on your machine

pdyatlov:
Just double checked,

1. Calling SceneManager.GetActiveScene().name does NOT trigger a crash
2. Loading another scene does NOT trigger a crash
3. Only loading a scene with specific objects (code?) causes it

bruno-garcia:
> Only loading a scene with specific objects (code?) causes it

Which objects would those be? Does it happen without the Sentry SDK? I wonder if it's something else we do in combination with that API.

pdyatlov:
> Which objects would those be? 
I'm sadly not sure. I tried to binary-search the culprit, but to no avail. It seems like there's a combination of objects / meshes / scripts

> Does it happen without the Sentry SDK? I wonder if it's something else we do in combination with that API.
[...]

Issue #2200, opened 2025-06-18
Improve SDK Initialization


bitsandfoxes:
## The Problem

Currently there are multiple ways to initialize the SDK, none of them ideal. There are multiple `Init` methods while the SDK is auto-initializing. Users have no straight forward way to control or delay initialization. 

## Context

There are multiple ways to initialize the SDK.

### `SentrySdk.Init()`

This is the entry point into the .NET SDK. It is discoverable and callable from anywhere inside the Unity project. The SDK calls this as part of `SentryUnity.Init`.

### `SentryUnity.Init()`

This is the entry point into the Unity SDK.
 https://github.com/getsentry/sentry-unity/blob/14882b7b53636e05ce995c8d7ba20c68d0a90f2e/src/Sentry.Unity/SentryUnity.cs#L31
The SDK is also using it as part of the auto-initialization through `SentryInitialization`

### `SentryInitialization.cs`
The SDK provides a `SentryInitialization.cs` as `.cs` script so it has access to precompiler directives and other, targeted platform specifics like IL2CPP methods
https://github.com/getsentry/sentry-unity/blob/14882b7b53636e05ce995c8d7ba20c68d0a90f2e/package-dev/Runtime/SentryInitialization.cs#L54-L56

Due to the `RuntimeInitializeOnLoadMethod` there's currently no way around the SDK attempting to auto-initialize. 
The flow is
1. The options get loaded from `/Resources`
2. The `ConfigureCallback` gets invoked, allowing user code to get executed
https://github.com/getsentry/sentry-unity/blob/14882b7b53636e05ce995c8d7ba20c68d0a90f2e/package-dev/Runtime/SentryInitialization.cs#L60
3. Native SDKs get setup/initialized (scope observer, background worker setup, e.t.c)
4. https://github.com/getsentry/sentry-unity/blob/14882b7b53636e05ce995c8d7ba20c68d0a90f2e/package-dev/Runtime/SentryInitialization.cs#L88-L96
5. The SDK calls `Init`
https://github.com/getsentry/sentry-unity/blob/14882b7b53636e05ce995c8d7ba20c68d0a90f2e/package-dev/Runtime/SentryInitialization.cs#L69

### Changes

1. Unify all `Init` into one class `SentryUnity`.
2. Hide .NET SDK init - we do have access to the source due to submodule. We can hide it, or have it partial?
[...]

PR #2206, opened 2025-06-20
fix: Make fetching scene name run on IL2CPP only

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/2181

### DIFF
--- a/src/Sentry.Unity/ScriptableSentryUnityOptions.cs
+++ b/src/Sentry.Unity/ScriptableSentryUnityOptions.cs
@@ -185,6 +185,8 @@ public class ScriptableSentryUnityOptions : ScriptableObject
             PerformanceAutoInstrumentationEnabled = AutoAwakeTraces,
         };
 
+        options.AddIntegration(new UnityScopeIntegration(application, unityInfo));
+
         if (!string.IsNullOrWhiteSpace(ReleaseOverride))
         {
             options.Release = ReleaseOverride;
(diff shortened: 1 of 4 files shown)


src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -135,7 +135,7 @@ internal SentryUnityOptions ToSentryUnityOptions(bool isBuilding, ISentryUnityIn
     {
         application ??= ApplicationAdapter.Instance;
 
-        var options = new SentryUnityOptions(isBuilding, application)
+        var options = new SentryUnityOptions(isBuilding, application, unityInfo)
```
bitsandfoxes, line 138: We need to pass the `UnityInfo` into the options if we want to have access to `IL2CPP` from within the integrations.
Alternatively, I'd need to add the integrations here but this would be a breaking change in behaviour for anyone that programmatically initializes the SDK.

I'll have to pick this back up when working on https://github.com/getsentry/sentry-unity/issues/2200

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - The SDK no longer attaches screenshots when capturing errors in the Unity Editor. ([#2163](https://github.com/getsentry/sentry-unity/pull/2163))
 
+### Fixes
+
+- When choosing the Mono scripting backend, and on certain NVIDIA platforms, the SDK would cause a crash with `EXCEPTION_ACCESS_VIOLATION_READ`. This is done by limiting adding the active scene name to the contexts to IL2CPP builds ([#2206](https://github.com/getsentry/sentry-unity/pull/2206))
```
bitsandfoxes:
```suggestion
- The SDK no longer causes crashes with `EXCEPTION_ACCESS_VIOLATION_READ` when using the Mono scripting backend. The SDK now adds the active scene name to the context in IL2CPP builds only ([#2206](https://github.com/getsentry/sentry-unity/pull/2206))
```

PR #2212, opened 2025-06-24
fix: Android `SIGABORT` due to detaching from JNI

bitsandfoxes:

We're making use of the fact that `AndroidJni.GetVersion()` returns `0` if the current thread is not attached.

### DIFF
--- a/src/Sentry.Unity.Android/SentryJava.cs
+++ b/src/Sentry.Unity.Android/SentryJava.cs
@@ -209,10 +209,8 @@ internal class SentryJava : ISentryJava
         }
         finally
         {
-             HandleJniThreadDetachment();
+            HandleJniThreadDetachment();
         }
-
-        _didSdkAttachToJni.Dispose();
     }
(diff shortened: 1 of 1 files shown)


src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -51,6 +52,8 @@ internal class SentryJava : ISentryJava
 {
     private readonly IAndroidJNI _androidJNI;
     private IDiagnosticLogger? _logger;
+    private ThreadLocal<bool> DidSdkAttachToJni = new();
```
Flash0ver: naming: fields should be lowercase and since it's `private ` prefixed with `_`
modifier: can be `readonly`

Flash0ver (approved):

bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-unity/pull/2215
See PR description for why this approach does not work.

Issue #928, opened 2022-08-02
User Feedback Prefab 


zoesyc:
Give users an easier way to provide user feedback using [prefabs](https://docs.unity3d.com/Manual/Prefabs.html)

bitsandfoxes:
## Implementation

Looking through the [JS documentation](https://docs.sentry.io/platforms/javascript/user-feedback/configuration/#bring-your-own-widget) for inspiration there are a two bigger points

### The Widget - A configurable prefab

A prefab with one `SentryUserFeedbackMonoBehaviour` sitting on the root. The MonoBehaviour reads the feedback widget control options during `Awake`. The options could be stored on the [`ScriptableSentryUnityOptions.asset`](https://github.com/getsentry/sentry-unity/blob/main/src/Sentry.Unity/ScriptableSentryUnityOptions.cs) that are saved to `/Resource`. They could also be their own `.asset`.

#### Controls what to show

- `Show Name`
  - `Is Name Required`
- `Show Email`
  - `Is Email Required`
- `Capture Screenshot`

#### Controls how to show it 

- `Form Title`
- `Submit Button Text`
- `Cancel Button Text`
- `Add Screenshot Button Text`
- `Message Input Field Title`
- `Message Input Field Placeholder`
- `Name Input Field Title`
- `Name Input Field Placeholder`
- `Email Input Field Title`
- `Email Input Field Placeholder`
- `Success Message Text`

### A way do provide a user-prefab

PR #2220, opened 2025-06-26
feat: User Feedback

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/928

Note: I've updated the prefab to include the logo on the top right. See the updated screenshot on Android.

## Goal

The idea is to provide users a simple to use prefab to add to their game to allow players to provide user feedback. This is a first iteration and the things we're looking to have added to the feedback is:
- Name (optional)
- Email (optional)
- A message - i.e. feedback (required)
- Screenshot (optional)

## Implementation Details

A prefab that holds a GameObject with a scaling canvas that can be dropped into any scene.
The prefab consists of two parts:
1. The button to pop up the form
2. The form to submit the feedback

Screenshot 2025-06-27 at 13 18 46

### One Prefab

Button and form are part of the same prefab. It make little sense to provide them separately to customize them. The button would need to be provided with the form prefab, making this a customization. So you'd always end up with changes to both anyway.
Attempting to pass this (or parts of it) into the editor configuration window would counter the Unity workflow of adding prefabs to scenes that are support to have the feedback.

### One `MonoBehaviour` that's shipped as `.cs`

Since we're providing a prefab to be dragged into scenes and we expect users to customize them, the accompanying `MonoBehaviour` should be as accessible as possible. We're providing it as a heavily commented "raw" `.cs` file that is getting compiled with the game. 
This also allows us **not** to take on `UnityEngine.UI` as a dependency for the Unity SDK.

#### Custom Inspector

We're providing the `SentryUserFeedbackMonoBehaviour ` it's own custom inspector to highlight broken or missing references

Screenshot 2025-06-27 at 13 19 07

- Missing optional fields are getting highlighted in yellow.
- Missing required fields are getting highlighted in red.

### How to allow for customization

The prefab can be dragged anywhere into the `/Assets` allowing for the creation of a [prefab variant](https://docs.unity3d.com/6000.1/Documentation/Manual/PrefabVariants.html). This variant "inherits" from the original and all values that have not been explicitly changed will get updated. 
The form is controlled by a `VertialLayoutGroup` which allows for optional inputs (i.e. name, email) to be dropped while retaining the layout.

### Adding to the scene

Adding the User Feedback to the scene can be done via drag&drop:

https://github.com/user-attachments/assets/ee43982f-8adb-472a-903b-2ce7ea903e05

## The Feature in Action

Clicking the megaphone button pops up the form to submit user feedback.

https://github.com/user-attachments/assets/8b114262-8a0c-483d-b21a-c885930e65b3

Since the canvas scales with the screen, this works on a wide range of displays and resolutions, i.e. mobile

Screenshot 2025-06-27 at 13 18 08

When submitting the report we're delaying to send it to the `EndOfFrame` so the screenshot does not contain the form.

Screenshot 2025-06-27 at 13 18 08

### DIFF
--- a/package-dev/Editor/SentryUserFeedbackEditor.cs.meta
+++ b/package-dev/Editor/SentryUserFeedbackEditor.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 788f34b6497dc4ab881dffd0e4932cd9


src/Sentry.Unity/SentryScreenshotUtility.cs
bitsandfoxes, line 1: I moved this out of the `ScreenshotEventProcessor` to be reused by `CaptureUserFeedback` as well.

samples/unity-of-bugs/Assets/Scenes/1_Bugfarm.unity
bitsandfoxes, line 1: Added the UserFeedback to the scene

package-dev/Editor/SentryUserFeedbackEditor.cs
bitsandfoxes, line 1: This is the custom inspector for the `SentryUserFeedback` MonoBehaviour that we're providing as `.cs`.

package-dev/Runtime/SentryUserFeedback.cs
```diff
@@ -0,0 +1,96 @@
+using System;
+using System.Collections;
+using UnityEngine;
+using UnityEngine.UI;
```
bitsandfoxes, line 3: We could have this MonoBehaviour as part of the SDK and ship it precompiled. But there are a couple for downsides

1. The SDK would need to take on an additional dependency on `UnityEngine.UI`. This is not part of the `ManagedPath/UnityEngine` but lives somewhere in `/ProjectPath/Library` instead. We'd need create some workaround like we do for the testrunner.
2. Users can't inspect the `.cs` in the Editor. 
3. The workflow of Embed&Change would not work that way

package-dev/Runtime/SentryUserFeedback.cs
```diff
@@ -0,0 +1,96 @@
...
+        [SerializeField] private Button _sendBugReportButton;
+        [SerializeField] private Button _cancelButton;
+
+        // Inputs
+        [SerializeField] private InputField _name;
+        [SerializeField] private InputField _email;
+        [SerializeField] private InputField _message;
+        [SerializeField] private Toggle _addScreenshot;
+
+        private void Awake()
```
bitsandfoxes, line 23: The checks here go to the console and not the SDK logger and all have `this` as context. This allows us to 
- Warn right in the Editor
- Clicking on the log highlights the GameObject in the Hierarchy
- The custom inspector highlights any missing references

bruno-garcia:
@sentry review

package-dev/Editor/SentryUserFeedbackEditor.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                EditorGUILayout.PropertyField(property, new GUIContent(displayName));
+            }
+            else
+            {
+                GUI.backgroundColor = Color.red;
+                EditorGUILayout.PropertyField(property, new GUIContent(displayName));
+                GUI.backgroundColor = Color.white;
+            }
+        }
```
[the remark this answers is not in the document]
bruno-garcia: xml docs on private functions? get out of here

package-dev/Runtime/SentryUserFeedback.cs
```diff
@@ -0,0 +1,96 @@
...
+            // Hide the feedback form
+            _feedbackForm.SetActive(false);
+
+            // We're waiting for the EndOfFrame so we can hide the feedback form before capturing the screenshot
+            yield return new WaitForEndOfFrame();
+
+            SentryUnity.CaptureFeedback(_message.text, _email.text, _name.text, _addScreenshot);
+
+            HideFeedbackForm();
+        }
```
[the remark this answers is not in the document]
bruno-garcia: Makes no sense 😢

src/Sentry.Unity/SentryUnity.cs
```diff
@@ -85,4 +85,10 @@ public static CrashedLastRun GetLastRunState()
 
         return UnitySdk.CrashedLastRun();
     }
+
+    /// <summary>
+    /// Captures a User Feedback
+    /// </summary>
+    public static void CaptureFeedback(string message, string? email, string? name, bool addScreenshot) =>
```
bruno-garcia, line 92: Hm this is odd that we add an API here because it creates further confusion between `SentrySdk` vs `SentryUnity`. Both will have `CaptureFeedback` from the `SentrySdk` and get the screenshot taken before calling it and doing `s => s.AddAttachment`?

bruno-garcia: also not totally against adding this here, just sayijng it's not ideal

package-dev/Editor/SentryUserFeedbackEditor.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            _openFeedbackButton = serializedObject.FindProperty("_openFeedbackButton");
+            _feedbackForm = serializedObject.FindProperty("_feedbackForm");
+            _sendBugReportButton = serializedObject.FindProperty("_sendBugReportButton");
+            _cancelButton = serializedObject.FindProperty("_cancelButton");
+            _name = serializedObject.FindProperty("_name");
+            _email = serializedObject.FindProperty("_email");
+            _message = serializedObject.FindProperty("_message");
+            _addScreenshot = serializedObject.FindProperty("_addScreenshot");
+        }
```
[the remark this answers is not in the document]
bitsandfoxes: The whole point of the custom inspector is to validate, handle and display properties. They are simply getting fetched in `OnEnable()`.

package-dev/Runtime/SentryUserFeedback.cs
```diff
@@ -0,0 +1,96 @@
...
+        public void HideFeedbackForm()
+        {
+            // Resetting the form
+            _name.text = string.Empty;
+            _email.text = string.Empty;
+            _message.text = string.Empty;
+            _addScreenshot.isOn = true;
+
+            _feedbackForm.SetActive(false);
+            _openFeedbackButton.SetActive(true);
```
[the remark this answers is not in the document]
bitsandfoxes: I've updated to validate required components in Awake and check for null during reset.

src/Sentry.Unity/SentryUnitySDK.cs
```diff
@@ -113,4 +113,24 @@ public SentryUnity.CrashedLastRun CrashedLastRun()
...
+            ? SentryHint.WithAttachments(
+                new SentryAttachment(
+                    AttachmentType.Default,
+                    new ByteAttachmentContent(SentryScreenshotUtility.Capture(_options)),
+                    "screenshot.jpg",
+                    "image/jpeg"))
+            : null;
+
+        SentrySdk.CaptureFeedback(message, email, name, hint: hint);
+    }
```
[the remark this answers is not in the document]
bitsandfoxes: This is getting handled by the underlying [.NET SDK](https://github.com/getsentry/sentry-dotnet/blob/e7240a46f9325df0b3942c5f1b04684bf7ec233f/src/Sentry/Internal/Hub.cs#L574-L588).

src/Sentry.Unity/SentryUnity.cs
bitsandfoxes: I think the .NET SDK one should be hidden along with its `Init`.
Add a screenshot to the feedback should not require to either
1. Implement their own screenshot capture to pass it in as attachment
2. The SDK to make the screenshot capture publicly accessible
3. Having to write code like this
```
SentryHint.WithAttachments(
    new SentryAttachment(
        AttachmentType.Default,
        new ByteAttachmentContent(SentryScreenshotUtility.Capture(_options)),
        "screenshot.jpg",
        "image/jpeg"))
```

package-dev/Runtime/SentryUserFeedback.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                Debug.LogError("Message input field is missing", this);
+                gameObject.SetActive(false);
+            }
+
+            if (!_addScreenshot)
+            {
+                Debug.LogError("Add Screenshot toggle is missing", this);
+                gameObject.SetActive(false);
+            }
```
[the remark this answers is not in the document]
bitsandfoxes: Add the curly brackets back in and it's looking just as repetitive. 
This is code we're providing to users as-is or as sample to be modified. These checks should be as simple to follow as possible.

src/Sentry.Unity/SentryScreenshotUtility.cs
```diff
@@ -0,0 +1,83 @@
...
+        RenderTexture.ReleaseTemporary(renderTextureResized);
+
+        var bytes = screenshot.EncodeToJPG(options.ScreenshotCompression);
+        Object.Destroy(screenshot);
+
+        options.DiagnosticLogger?.Log(SentryLevel.Debug,
+            "Screenshot captured at {0}x{1}: {2} bytes", null, width, height, bytes.Length);
+
+        return bytes;
+    }
```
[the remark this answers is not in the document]
bitsandfoxes: Good points in there.

package-dev/Runtime/SentryUserFeedback.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        // This coroutine allows us to hide the FeedbackForm before capturing so it's not visible on the screenshot
+        private IEnumerator HideFormAndCaptureFeedback()
+        {
+            // Hide the feedback form
+            _feedbackForm.SetActive(false);
+
+            // We're waiting for the EndOfFrame so the FeedbackForm gets updated before capturing the screenshot
+            yield return new WaitForEndOfFrame();
```
bruno-garcia, line 100: Nice!

src/Sentry.Unity/ScreenshotEventProcessor.cs
```diff
@@ -23,6 +23,14 @@ internal ScreenshotEventProcessor(SentryUnityOptions sentryOptions, IApplication
 
     public SentryEvent? Process(SentryEvent @event, SentryHint hint)
     {
+        // save event id
+        // wait for end of frame
+        // check if last id is event it
+        // send screenshot
+
+        // add workitem: screentshot for ID xxx
+        // sdk integration checking for work: if ID got sent, follow up with screenshot
```
bruno-garcia, line 32: needs to be cleaned up?

src/Sentry.Unity/SentryUnity.cs
bruno-garcia: > I think the .NET SDK one should be hidden along with its Init.


How do you capture events? It'll be on `SentrySdk`, right? some stuff on one, some on another

bruno-garcia (approved):

PR #2227, opened 2025-07-03
Rework Initialization

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/2200
Required https://github.com/getsentry/sentry-dotnet/pull/4322

# Goal

As stated in the issue: The SDK needs to initializeable in two ways: self, and manual. Both these paths need to provide the same functionality.

# Approach

1. Hide the .NET SDK static API by making it internal
2. Mirror the now hidden API into `Sentry.Unity.SentrySdk` - This way the Unity SDK can omit certain methods like the .NET SDKs `Init`
3. Hide the `SentryInitialization.cs` that is getting compiled with the game
4. Have all actual initialization - that includes platform specific configuration like scope sync - happen within the pre-compiled SDK. 
5. The code that gets compiled with the game is strictly for setting up and calling `Init`. This allows users to call the same method.

### Hide the .NET SDK

This is done by conditionally making the .NET SDKs `SentrySdk`s accessor `internal` https://github.com/getsentry/sentry-dotnet/pull/4322
The .NET SDK now considers Unity as a platform and our very own `.props` allows us to conditionally set a constant.

### Rename Rename `SentryUnity` to `SentrySdk`

Ideally, we'd only have one static API that allows for discoverability of what's available. This could either be `SentrySdk` or `SentryUnity`. I'm going with `SentrySdk` here as:
- it's a smoother migration to fix `usings`. This change will require users to change them from `using Sentry;` to `using Sentry.Unity`. Which is still easier than having to go through every instance of calling into the SDK and change these from `SentrySdk` to `SentryUnity`.
- it resides in the `Sentry.Unity` namespace and avoids conflicts. Attempting to access `Sentry.SentrySdk.Capture` will create an error `Cannot access internal class 'SentrySdk' here`.
    - We're helping out here by adding a `summary` to the internal `SentrySdk` guiding users towards the right namespace

### Make `Sentry.Unity.SentrySdk` a `partial` class

The Unity SDK version of `SentrySdk` is now split in two partial classes.

`SentrySdk.cs` is where the Unity SDK specific code resides and basically contains "overwritten" methods like `Init`, `GetLastRunState`.

`SentrySdk.Dotnet.cs` - **TODO**: This should be source generated and end up as `SentrySdk.Dotnet.g.cs`
The .NET SDK `SentrySdk` is now hidden and inaccessible. To give users access to the rest of the API we're putting all the rest of the static API in there hand have it forward the calls into the .NET SDK.

### Hide the `.cs` scripts `SentryInitialization` and `SentryIntegrations`

Our package ships with `io.sentry.asmdef` which creates our own assembly for us within the game's project. This allows us to make these classes `internal`, but keep their name. They are now undiscoverable for users - and should have been so in the first place.

### Changes to `SentryInitialization`

The `SentryInitialization` still receives the "first thing that happens when the game starts callback via the `RuntimeLoad` attribute. This allows us to set up the SDK all the way to be ready to be initialized. Instead of doing the actual initialization inside the script that gets compiled with the game, this now injects the platform specific configure calls into the SDK.

### Add a static `SentryPlatformServices` 

This is now the place that holds the injected platform specific configure callbacks and the `IUnityInfo` that we use to access version and platform specific code like IL2CPP runtime checks and backend calls for line number support.

We're setting these as the very first thing and can so guarantee that they are available, regardless whether the SDK auto-initializes or has the user call `Init` manually.

The `PlatformServices` stay on the options because
1. Ease of access
2. No global variable floating around somewhere that gets accessed from all kinds of places

Due to how things are getting initialized I had to add a `SentryEditorPlatformSetUp` that sets the `UnityInfo` by virtue of the `InitializeOnLoadMethod` attribute. This allows us to have access to this even during build time (no runtime initialization there) and we can keep requiring it on the `SentryUnityOptions`.

### Move all initialization code into the SDK

Having the platform specific configure callbacks available via the `SentryPlatformServices` class allows us to call conditionally compiled code from within the pre-compiled SDK. All initialization paths funnel back together in that one `Init` call.

# Migration Path

Unfortunately, this is a breaking change. As such, we need to provide a migration path. When updating the SDK users will face the following error on the console

Screenshot 2025-07-07 at 13 25 24

Since we can modify the now internal SDK's `<summary>` we can already put some help there

Screenshot 2025-07-07 at 13 27 41

IDEs like Rider will allow to automatically import the missing references
```
using Sentry;
using Sentry.Unity;
```

# Things to follow up on

## `using Sentry.Unity` vs. `using Sentry`
The entirety of the API is now discoverable within `SentrySdk.` but the namespace is split 
- `Sentry` - Contains the .NET SDK's types, i.e. `SentryId`
- `Sentry.Unity` - Contains the static API

So users end up in situations where they need to always bring `using Sentry.Unity` and sometimes additionally need `using Sentry`. That was an issue before already, it's just something to be aware of.

## Consider turning `IsEnabled` in the config window into `isSelfInitEnabled`

During build the SDK was opinionated regarding whether to bring a native SDK and tied setup and symbol upload to the `isEnabled` check. When we merged all option-configuration callbacks into one in https://github.com/getsentry/sentry-unity/pull/1888 the line already blurred.
And now, the SDK cannot reasonably be certain whether when it will be initialized at runtime. With this I think we have to consider
1. `IsEnabled` cannot be considered anymore
4. Is the symbol upload enabled -> upload symbols
5. Is the native support enabled -> add native SDK
6. On mobile: If `IsEnabled` and `InitializationType.BuildTime` -> setup native project

## Source Generator for `Sentry.Dotnet`

Right now the `Sentry.Dotnet.cs` is created by hand. It forwards all calls to the now internal static API while omitting `Init`. This allows us to keep the Unity specific methods in the `SentrySdk.cs` file but it also requires us to manually update if and when the .NET SDK's public API changes.

# Alternative (attempted) approaches

- I gave [`extension members`](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) a try by bumping to C#14. But this cannot be consumed by the user's code.
- Making only `SentrySdk.Init` methods `internal`. This somewhat works but gets us stuck in a place. We'd have the API split between `SentryUnity.Init` vs `SentrySdk.Capture`.
- Going with `SentryUnity` as name for the type for the static API. But - and that's an assumption on my part - that would break the actual calls from `SentrySdk.CaptureException()` `SentryUnity.CaptureException()` instead of having the IDE fix the `usings` and be done with it.
- Adding the Unity SDK init to the `Sentry` namespace. This did not work for conflicting reasons.


package-dev/Runtime/SentryInitialization.cs
```diff
@@ -53,58 +55,49 @@ public static class SentryInitialization
 #else
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
 #endif
-        public static void Init()
+        internal static void Init()
```
bruno-garcia: Does this make any difference to the final app?

I mean, since this is compiled with the game, doesn't it mean the game code can access it anyway?

If so, and since it gets called by the runtime anyway, could it be `private` instead?
And lastly, should we add some XML docs to the class itself saying this isnt' designed to be called by the user? (unless of course there are use cases where we do document users should be calling this programatically

samples/unity-of-bugs/Assets/Scripts/BugFarmButtons.cs
```diff
@@ -1,15 +1,17 @@
...
 using UnityEngine.Assertions;
 
+
 public class BugFarmButtons : MonoBehaviour
 {
     private void Awake()
     {
         Debug.Log("Sample 🐛");
+
+        SentrySdk.Init(options => {});
```
bruno-garcia: I realize it's a draft, I assume this testing things out?

package-dev/Runtime/SentryUserFeedback.cs
```diff
@@ -86,7 +86,7 @@ public void SendFeedback()
             else
             {
                 // Since there is no screenshot added we can capture the feedback right away
-                SentryUnity.CaptureFeedback(_description.text, _email.text, _name.text, addScreenshot: false);
+                SentrySdk.CaptureFeedback(_description.text, _email.text, _name.text, addScreenshot: false);
```
bitsandfoxes, line 89: This is the desired effect. All things reside within `SentrySdk`.

package-dev/Runtime/SentryInitialization.cs
```diff
@@ -39,7 +41,7 @@
 
 namespace Sentry.Unity
 {
-    public static class SentryInitialization
+    internal static class SentryInitialization
```
bitsandfoxes, line 44: We can safely make this `internal`. The SDK ships with it's own `.asmdef` that allows us to hide internal types from the user.

src/Sentry.Unity/NativeUtils/SentryPlatformServices.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace Sentry.Unity.NativeUtils;
```
bitsandfoxes, line 3: Putting the services into a very distinctive namespace to "discourage" usage.

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -120,12 +121,12 @@ public static string GetConfigPath(string? notDefaultConfigName = null)
     /// <remarks>
     /// Used for loading the SentryUnityOptions from the ScriptableSentryUnityOptions during runtime.
     /// </remarks>
-    public static SentryUnityOptions? LoadSentryUnityOptions(ISentryUnityInfo unityInfo)
+    public static SentryUnityOptions? LoadSentryUnityOptions()
```
bitsandfoxes, line 124: We no longer need to provide the `IUnityInfo` since these are getting set up during startup as part of the new platform services that are statically available.

src/Sentry.Unity/SentrySdk.Dotnet.cs
bitsandfoxes, line 1: This should get source generated in a follow up PR.

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -49,11 +49,11 @@ private SentryUnitySdk(SentryUnityOptions options)
             }
         }
 
-        unitySdk._dotnetSdk = SentrySdk.Init(options);
+        unitySdk._dotnetSdk = Sentry.SentrySdk.Init(options);
```
bitsandfoxes, line 52: A drawback is that we now need to fully qualify calls through the static API from within the Unity SDK. But our users won't have that problem.

test/Sentry.Unity.Tests/ContextWriterTests.cs
```diff
@@ -79,7 +79,7 @@ public void Arguments()
 
         // act
         MainThreadData.SentrySystemInfo = sysInfo;
-        SentryUnity.Init(options);
+        SentrySdk.Init(options);
```
bitsandfoxes, line 82: No more separation between `SentryUnity` and `SentrySdk`.

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@
 
 ### Breaking Changes
 
+- **Breaking Change**: The Unity SDK's static API has been simplified moved from `Sentry.Unity.SentryUnity` and `Sentry.SentrySdk`
+  to `Sentry.Unity.SentrySdk`. 
+  This change enables manual SDK initialization with full functionality, previously only available through auto-initialization.
+  The underlying .NET SDK's `SentrySdk` class is now internal, and several previously public classes like `SentryInitialization` 
+  and `SentryIntegrations` are now internal.
+  
+  **Migration**: Update your using statements from `using Sentry;` to `using Sentry.Unity;`. IDEs like Rider can automatically
```
bruno-garcia:
```suggestion
  **Migration**: Update your `using` directives from `using Sentry;` to `using Sentry.Unity;`. IDEs like Rider 
```

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@
...
+  to `Sentry.Unity.SentrySdk`. 
+  This change enables manual SDK initialization with full functionality, previously only available through auto-initialization.
+  The underlying .NET SDK's `SentrySdk` class is now internal, and several previously public classes like `SentryInitialization` 
+  and `SentryIntegrations` are now internal.
+  
+  **Migration**: Update your using statements from `using Sentry;` to `using Sentry.Unity;`. IDEs like Rider can automatically 
+  import the missing references. In some cases, you may need both `using Sentry.Unity;` (for the static API) and `using Sentry;` 
+  (for types like `SentryId`). No changes are required to your actual SDK method calls (e.g., `SentrySdk.CaptureException()` 
+  remains the same). ([#2227](https://github.com/getsentry/sentry-unity/pull/2227))
+- Updated Unity support by updating to Unity versions (2020+), removing Unity 2019 which reached End of Life in 2022 ([#2231](https://github.com/getsentry/sentry-unity/pull/2231))
```
bruno-garcia:
```suggestion
- Removed Unity 2019 support, which reached End of Life in 2022. Minimum supported version now is 2020 ([#2231](https://github.com/getsentry/sentry-unity/pull/2231))
```

package-dev/Runtime/SentryIntegrations.cs
```diff
@@ -50,6 +70,26 @@ public void Register(IHub hub, SentryOptions options)
...
+            Logger.LogInfo("Creating '{0}' transaction for runtime initialization.",
+                StartupTransactionOperation);
+
+            var runtimeStartTransaction =
+                SentrySdk.StartTransaction("runtime.initialization", StartupTransactionOperation);
+            SentrySdk.ConfigureScope(scope => scope.Transaction = runtimeStartTransaction);
+
+            Logger.LogDebug("Creating '{0}' span.", InitSpanOperation);
+            InitSpan = runtimeStartTransaction.StartChild(InitSpanOperation, "runtime initialization");
+            Logger.LogDebug("Creating '{0}' span.", SubSystemSpanOperation);
```
bruno-garcia: Could these two be combined in 1 log message? Debug log can be overwhelming at times

samples/unity-of-bugs/Assets/Scripts/BugFarmButtons.cs
```diff
@@ -1,10 +1,11 @@
 using System;
-using System.Globalization;
 using System.Runtime.CompilerServices;
 using Sentry;
+using Sentry.Unity;
 using UnityEngine;
 using UnityEngine.Assertions;
 
+
```
bruno-garcia:
```suggestion
```

src/Sentry.Unity/SentrySdk.Dotnet.cs
```diff
@@ -0,0 +1,546 @@
...
+
+    /// <summary>
+    /// Deliberately crashes an application, which is useful for testing and demonstration purposes.
+    /// </summary>
+    /// <remarks>
+    /// The method is marked obsolete only to discourage accidental misuse.
+    /// We do not intend to remove it.
+    /// </remarks>
+    [Obsolete("WARNING: This method deliberately causes a crash, and should not be used in a real application.")]
+    public static void CauseCrash(CrashType crashType) => Sentry.SentrySdk.CauseCrash(crashType);
```
bruno-garcia, line 542: Since this now will require manually keeping in sync. You could have a test that does reflection on `SentrySdk` and gets all public members, does the same on the .NET and Unity SDKs and compares the two to make sure there's no new APIs on the .NET side

src/Sentry.Unity/SentrySdk.cs
```diff
@@ -32,7 +33,29 @@ public static void Init(SentryUnityOptions options)
     {
         if (UnitySdk is not null)
         {
-            options.DiagnosticLogger?.LogWarning("The SDK has already been initialized.");
+            options.LogWarning("The SDK has already been initialized. Skipping initialization.");
+        }
+
+        if (SentryPlatformServices.UnityInfo is not null && SentryPlatformServices.PlatformConfiguration is not null)
```
bruno-garcia:
```suggestion
        if (SentryPlatformServices.UnityInfo is {} unityInfo && SentryPlatformServices.PlatformConfiguration is {} platConfig)
```

Use these ^ instead so you make sure concurrently these static values can't be set to `null` casing an error on the calls below

bruno-garcia (commented):
quick pass

package-dev/Runtime/SentryInitialization.cs
bitsandfoxes: Good point!

CHANGELOG.md
```diff
@@ -4,7 +4,18 @@
 
 ### Breaking Changes
 
-- Dropped support for Unity 2019. It reached End of Life in 2022 ([#2231](https://github.com/getsentry/sentry-unity/pull/2231))
+- Removed Unity 2019 support, which reached End of Life in 2022. Minimum supported version now is 2020 ([#2231](https://github.com/getsentry/sentry-unity/pull/2231))
+
+- **Breaking Change**: The Unity SDK's static API has been simplified moved from `Sentry.Unity.SentryUnity` and `Sentry.SentrySdk`
+  to `Sentry.Unity.SentrySdk`. 
+  This change enables manual SDK initialization with full functionality, previously only available through auto-initialization.
```
bruno-garcia:
```suggestion
  This change enables [manual/programatic SDK initialization](https://docs.sentry.io/platforms/unity/configuration/options/programmatic-configuration/) with full functionality, previously only available through auto-initialization.
```

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -22,13 +23,16 @@ public static class SentryNative
...
+        Configure(options, SentryPlatformServices.UnityInfo);
+
+    internal static void Configure(SentryUnityOptions options, ISentryUnityInfo? sentryUnityInfo)
     {
         Logger = options.DiagnosticLogger;
 
         Logger?.LogInfo("Attempting to configure native support via the Native SDK");
 
-        if (!sentryUnityInfo.IsNativeSupportEnabled(options, ApplicationAdapter.Instance.Platform))
+        if (!sentryUnityInfo?.IsNativeSupportEnabled(options, ApplicationAdapter.Instance.Platform) ?? false)
```
bruno-garcia: `sentryUnityInfo` being null is a bigger problem, no?

Shoujld we split this code so we can log specifically `sentryUnityInfo` being null as an error. And the native support being off can stay as debug, as-is?

src/Sentry.Unity.Native/SentryNativeBridge.cs
```diff
@@ -12,13 +12,11 @@ namespace Sentry.Unity.Native;
...
-public static class SentryNativeBridge
+internal static class SentryNativeBridge
 {
-    public static bool CrashedLastRun;
-
-    public static bool Init(SentryUnityOptions options, ISentryUnityInfo sentryUnityInfo)
+    public static bool Init(SentryUnityOptions options, ISentryUnityInfo? sentryUnityInfo)
     {
-        _isLinux = sentryUnityInfo.IsLinux();
+        _isLinux = sentryUnityInfo?.IsLinux() ?? false;
```
bruno-garcia: since this is all internal code, should we expect `null` `ISentryUnityInfo`? 
Seems to me the caller should either have the instance or avoid getting an insance of this altogether?

If we have that null here, we'd still go ahead initializing everything below. Is that creating a weird scenario where instead we should have avoided `Init` being successful?

src/Sentry.Unity.iOS/SentryNativeCocoa.cs
```diff
@@ -14,15 +15,14 @@ public static class SentryNativeCocoa
     /// Configures the native support.
     /// </summary>
     /// <param name="options">The Sentry Unity options to use.</param>
-    /// <param name="sentryUnityInfo">Infos about the current Unity environment</param>
-    public static void Configure(SentryUnityOptions options, ISentryUnityInfo sentryUnityInfo) =>
-        Configure(options, sentryUnityInfo, ApplicationAdapter.Instance.Platform);
+    public static void Configure(SentryUnityOptions options) =>
```
bruno-garcia, line 17: This is a public method, and we only need to pass `options`.
But if the static `SentryPlatformServices.UnityInfo` wasn't set properly before hand, we'll just pass that through below and it'll be hard to trace back why that was.

I think we could validate here (if we expect `SentryPlatformServices.UnityInfo` not to be null here) that it is indeed not null. And deal with it (log and continue, throw?)

src/Sentry.Unity.iOS/SentryNativeCocoa.cs
```diff
@@ -14,15 +15,14 @@ public static class SentryNativeCocoa
...
+        Configure(options, SentryPlatformServices.UnityInfo, ApplicationAdapter.Instance.Platform);
 
-    internal static void Configure(SentryUnityOptions options, ISentryUnityInfo sentryUnityInfo, RuntimePlatform platform)
+    internal static void Configure(SentryUnityOptions options, ISentryUnityInfo? sentryUnityInfo, RuntimePlatform platform)
     {
         options.DiagnosticLogger?.LogInfo("Attempting to configure native support via the Cocoa SDK");
 
-        if (!sentryUnityInfo.IsNativeSupportEnabled(options, platform))
+        if (!sentryUnityInfo?.IsNativeSupportEnabled(options, platform) ?? false)
```
bruno-garcia: same situation where it's not clear if it's a mix up with setting up this static thing before calling this, or `IsNativeSupportEnabled` returning `false`, and making things hard to debug

src/Sentry.Unity.iOS/SentryNativeCocoa.cs
```diff
@@ -67,7 +67,7 @@ internal static void Configure(SentryUnityOptions options, ISentryUnityInfo sent
         };
 
         options.NativeSupportCloseCallback += () => Close(options, sentryUnityInfo, platform);
-        if (sentryUnityInfo.IL2CPP)
+        if (sentryUnityInfo?.IL2CPP ?? false)
```
bruno-garcia: also, lets avoid this pattern throughout.

src/Sentry.Unity/SentrySdk.Dotnet.cs
bruno-garcia: That's a better approach than the test I recommended 👍

bruno-garcia (commented):
Sorry I didn't notice this before but while in general I think this is a great approach, we now expect the static fields to be set before our public methods are called but we don't verify anywhere that's the case and we have a lot of null propagation patterns which hinder the ability to debug what was going on. In particular when customers change the init code, possibly bypassing our auto init/static property set in order to do things "by hand".

src/Sentry.Unity/NativeUtils/SentryPlatformServices.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// The UnityInfo holds methods that rely on conditionally compilation, i.e. IL2CPP backend.
+    /// </summary>
+    public static ISentryUnityInfo UnityInfo
+    {
+        get => _unityInfo ?? throw new InvalidOperationException("UnityInfo is null.");
+        set
+        {
+            if (_unityInfo != null)
+            {
+                throw new InvalidOperationException("Should not set twice. lol.");
```
bruno-garcia: why `lol`? 

We don't want it to be set twice why?
If we want to let the user set but they have to disable auto init. We should say that on the exception message to help the user understand why something went wrong.

I understand it shouldn't happen in any "default" mode of the SDK, but given the many migrations paths, and other init ways we offered in the past, no way users can accidently try to set it twice?

```suggestion
                throw new InvalidOperationException("Should not set twice.");
```

src/Sentry.Unity/NativeUtils/SentryPlatformServices.cs
```diff
@@ -0,0 +1,36 @@
...
+public static class SentryPlatformServices
+{
+    private static ISentryUnityInfo? _unityInfo;
+
+    /// <summary>
+    /// The UnityInfo holds methods that rely on conditionally compilation, i.e. IL2CPP backend.
+    /// </summary>
+    public static ISentryUnityInfo UnityInfo
+    {
+        get => _unityInfo ?? throw new InvalidOperationException("UnityInfo is null.");
```
bruno-garcia:
```suggestion
        get => _unityInfo ?? throw new InvalidOperationException("UnityInfo was not set. By default it's set via RuntimeInitialization, did you disable that?");
```

src/Sentry.Unity/NativeUtils/SentryPlatformServices.cs
```diff
@@ -0,0 +1,36 @@
...
+/// </summary>
+/// <remarks>Consider this <c>internal</c>.</remarks>
+public static class SentryPlatformServices
+{
+    private static ISentryUnityInfo? _unityInfo;
+
+    /// <summary>
+    /// The UnityInfo holds methods that rely on conditionally compilation, i.e. IL2CPP backend.
+    /// </summary>
+    public static ISentryUnityInfo UnityInfo
```
bruno-garcia, line 22: Example, to try to help the user understand wha'ts going on:


```suggestion
	/// <exception cref="System.InvalidOperationException">
	/// Thrown when unity info was not set since it's expected to be set automatically during Unity initialization.
	/// A common case is when runtime init is disabled, and assignment wasn't done explicitly by the user
	/// </exception>
    public static ISentryUnityInfo UnityInfo
```

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -294,15 +295,24 @@ internal string? DefaultUserId
...
+
     public SentryUnityOptions() : this(false, ApplicationAdapter.Instance) { }
 
     internal SentryUnityOptions(bool isBuilding, IApplication application, ISentryUnityInfo? unityInfo = null) :
         this(SentryMonoBehaviour.Instance, application, isBuilding, unityInfo)
     { }
 
     // For testing
-    internal SentryUnityOptions(SentryMonoBehaviour behaviour, IApplication application, bool isBuilding, ISentryUnityInfo? unityInfo = null)
+    private SentryUnityOptions(SentryMonoBehaviour behaviour, IApplication application, bool isBuilding, ISentryUnityInfo? unityInfo)
```
bruno-garcia: private for testing? comment perhaps outdated?

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -294,15 +295,24 @@ internal string? DefaultUserId
...
     {
+        // This should never happen. The PlatformServices are set through the RuntimeLoad attribute in 'SentryInitialization.cs'
```
bruno-garcia: I debug a few "this should never happen" things in my life to be more humble LOL
how about:

```suggestion
        // This shouldn't happen. The PlatformServices are set through the RuntimeLoad attribute in 'SentryInitialization.cs'
```

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -49,15 +49,19 @@ private SentryUnitySdk(SentryUnityOptions options)
...
+        StartupTracingIntegration.StartTracing();
 
         // We can safely call this during initialization. If the SDK self-initialized we're right on time. If the SDK
         // was initialized manually, the RuntimeOnLoad attributes already triggered, making this call a no-op.
         StartupTracingIntegration.StartTracing();
 
         if (options.NativeContextWriter is { } contextWriter)
         {
-            SentrySdk.ConfigureScope((scope) =>
+            Sentry.SentrySdk.CurrentHub.ConfigureScope((scope) =>
```
bruno-garcia, line 64: You can add to the top (or via csproj iirc): `using Sntry = Sentry.SentrySdk` and use only `Sntry` throughout the file/project

src/Sentry.Unity/NativeUtils/SentryPlatformServices.cs
bitsandfoxes: You caught my wip push from before the weekend 😅.
I've updated the documentation and exception messages.

src/Sentry.Unity/SentryUnitySdk.cs
bitsandfoxes: This will lead to `ambiguous references` for the namespace `Sentry.Unity` - between `Sentry` and `SentrySdk`.

src/Sentry.Unity.iOS/SentryNativeCocoa.cs
bitsandfoxes: Nice catch. We're handling this now by having the UnityInfo living on the options and getting validated there.

src/Sentry.Unity.Native/SentryNativeBridge.cs
bitsandfoxes: Reworked the PlatformServices.UnityInfo handling by pulling this into the options.

src/Sentry.Unity.Native/SentryNative.cs
bitsandfoxes: Reworked the PlatformServices.UnityInfo handling by pulling this into the options.

package-dev/Runtime/SentryIntegrations.cs
bitsandfoxes: I'll follow up on this in a separate PR.

src/Sentry.Unity/SentryUnitySdk.cs
vaind: Could `Sntry`, without `e` (as Bruno suggested) work though?

PR #2230, opened 2025-07-05
chore(readme): side effect of pinning version

bruno-garcia:


README.md
```diff
@@ -23,7 +23,7 @@ Install the package via the Unity Package Manager using a Git URL to Sentry's SD
 https://github.com/getsentry/unity.git
 ```
 
-To use this specific version of the SDK, append `#3.2.3` to the URL.
+To use this specific version of the SDK, append `#3.2.3` to the URL, but note that this will stop Unity from suggesting version updates automatically.
```
bitsandfoxes, line 26:
```suggestion
To use this specific version of the SDK, append `#3.2.3` to the URL, but note that this will require you to manually update by providing a new GIT URL.
```

Unity does not suggest updates when installing packages via GIT URL. But when you pin the version the update button does nothing. Without the added tag the button click will fetch the newest available version.

bitsandfoxes (approved):

PR #2231, opened 2025-07-08
Drop support for Unity 2019

bitsandfoxes:
Version 2019 reached [EOL mid 2022](https://unity.com/releases/2019-lts).
Apart from that there are other considerations:
- Really helps out with https://github.com/getsentry/sentry-unity/pull/2227
    - No longer needing to special case a non-precompiled integration
- The usage is basically zero.
- We don't have IL2CPP support
- It's a pain to use locally as it does not seem to get along with current versions of macOS
- It's another version to support (but noone uses)
- It restricts the approach to some implementation. (I really wish I'd kept note here. Will do for 2020 which eoled 2023)


.github/workflows/build.yml
```diff
@@ -7,7 +7,7 @@ on:
         type: string
 
 env:
-  LOWEST_SUPPORTED_UNITY_VERSION: 2019
+  LOWEST_SUPPORTED_UNITY_VERSION: 2020
```
bitsandfoxes, line 10: This is the version used for packaging.

.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        unity-version: ["2019", "2022", "6000"]
+        unity-version: ["2020", "2022", "6000"]
```
bitsandfoxes, line 90: I have not found a way to get these out of the matrix into the environment.

.github/workflows/update-deps.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           ssh-key: ${{ secrets.CI_DEPLOY_KEY }}
 
       - name: Update sample ProjectVersion.txt
-        if: ${{ matrix.unity-prefix == '2019' }}
+        if: ${{ matrix.unity-prefix == '2020' }}
```
bitsandfoxes, line 64: This is actually irrelevant. Unity 2020 LTS reached its EOL in mid-2023. Specifically, the final patch version was 2020.3.48.

package-dev/Runtime/SentryIntegrations.cs
bitsandfoxes, line 1: The scene tracing can now go into the SDK itself.

samples/unity-of-bugs/Packages/manifest.json
bitsandfoxes, line 1: I've also update the packages. Previously, were using the sample project for smoke testing in CI. For that we've stripped all non-required packages to keep it as slim as possible. We're building a new smoke test project from scratch now for a while and I'd rather have the sample be the same that all users would have.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- Updated Unity support by updating to Unity versions (2020+), removing Unity 2019 which reached End of Life in 2022 ([#2231](https://github.com/getsentry/sentry-unity/pull/2231))
```
bruno-garcia: start with "Dropped support for Unity 2019" since this is the most importantly part of the note

bruno-garcia (approved):

PR #2234, opened 2025-07-09
chore: Move startup tracing into the SDK

bitsandfoxes:
This is in part to help with https://github.com/getsentry/sentry-unity/pull/2227

We can have startup tracing be its own integration within the SDK instead of shipping it as a `.cs` that compiles with the game.

Things to note: 
1. We can use the `RuntimeInitializeOnLoadMethod` attribute from within the SDK. The only requirement is that those are `static` methods and take no arguments
2. These methods get called whether the integration got registered or not, hence the `IsStartupTracingAllowed` check.
3. For each method we have to first confirm that it is "allowed" to create a span
4. We make sure this only ever gets executed once by setting the `StartupAlreadyCaptured` after scene loading.
5. Setting the `StartupAlreadyCaptured` also handles the case of manual initialization later than during startup since
    - Those methods always get called
    - We can make sure that late initialization does not (wrongly) start a new startup transaction

#skip-changelog


test/Sentry.Unity.Tests/StartupTracingIntegrationTests.cs
bitsandfoxes, line 1: We didn't have any tests on this prior. :(

package-dev/Runtime/SentryInitialization.cs
bitsandfoxes, line 1: Instead of having Startup Tracing be split between `SentryInitialization` and `SentryIntegration` we can delete the latter and move everything into `internal StartupTracing` inside the SDK.

test/Sentry.Unity.Tests/StartupTracingIntegrationTests.cs
bruno-garcia: How much of it was AI?

bruno-garcia (approved):

PR #2239, opened 2025-07-10
fix: Moved late option mutation into `Init`

bitsandfoxes:
This is a follow-up on https://github.com/getsentry/sentry-unity/pull/2227 as this was getting already kinda big.

## Goal

This PR moves adding integrations that depend on certain flags into the `Init` call to be added "late" to keep the SDKs behavious consistent between self and manual initialization.

## Context

There are two ways `SentryUnityOptions` gets created:
- Self-Initializing: The SDK reads the `.asset` and creates new options from the ScriptableObject
- Manual-Initialization: User code initializes the SDK

In both cases, changes to the options need to be made at the same time. This leaves either the constructor or right before initialization.

## Note

One could point out the inconsistency how integrations get controlled and added, i.e. `AnrIntegration`, `ScreenshotProcessor` and `ViewHierarchyProcessor`. We could make changes to this but we'd require users to update from
```
options.AttachScreenshot = true;
```
to
```
options.AddEventProcessor(new ViewHierarchyEventProcessor(options));
```
which is not all that much better.

### DIFF
--- a/src/Sentry.Unity/Il2CppEventProcessor.cs
+++ b/src/Sentry.Unity/Il2CppEventProcessor.cs
@@ -17,12 +17,15 @@ internal class UnityIl2CppEventExceptionProcessor : ISentryEventExceptionProcess
     private static ISentryUnityInfo UnityInfo = null!; // private static will be initialized in the constructor
     private readonly Il2CppMethods _il2CppMethods;
 
-    public UnityIl2CppEventExceptionProcessor(SentryUnityOptions options, ISentryUnityInfo unityInfo)
+    public UnityIl2CppEventExceptionProcessor(SentryUnityOptions options)
     {
         Options = options;
-        UnityInfo = unityInfo;
-        _il2CppMethods = unityInfo.Il2CppMethods ?? throw new ArgumentNullException(nameof(unityInfo.Il2CppMethods),
-            "Unity IL2CPP methods are not available.");
+
+        // We're throwing here but this should never happen. We're validating UnityInfo before adding the processor.
+        UnityInfo = SentryPlatformServices.UnityInfo
+                    ?? throw new ArgumentNullException(nameof(SentryPlatformServices.UnityInfo), "UnityInfo is null");
+        _il2CppMethods = UnityInfo.Il2CppMethods
+                    ?? throw new ArgumentNullException(nameof(UnityInfo.Il2CppMethods), "Unity IL2CPP methods are not available.");
 
         Options.SdkIntegrationNames.Add("IL2CPPLineNumbers");
     }
--- a/src/Sentry.Unity/ScriptableSentryUnityOptions.cs
+++ b/src/Sentry.Unity/ScriptableSentryUnityOptions.cs
@@ -136,7 +136,7 @@ public class ScriptableSentryUnityOptions : ScriptableObject
     {
         application ??= ApplicationAdapter.Instance;
 
-        var options = new SentryUnityOptions(isBuilding, application, unityInfo)
+        var options = new SentryUnityOptions(isBuilding, application, unityInfo, SentryMonoBehaviour.Instance)
         {
             Enabled = Enabled,
             Dsn = Dsn,
@@ -186,6 +186,12 @@ public class ScriptableSentryUnityOptions : ScriptableObject
             PerformanceAutoInstrumentationEnabled = AutoAwakeTraces,
         };
 
+        // By default, the cacheDirectoryPath gets set on known platforms. The option is enabled by default
+        if (!EnableOfflineCaching)
+        {
+            options.CacheDirectoryPath = null;
+        }
+
         if (!string.IsNullOrWhiteSpace(ReleaseOverride))
         {
             options.Release = ReleaseOverride;
@@ -227,64 +233,17 @@ public class ScriptableSentryUnityOptions : ScriptableObject
         // Without setting up here we might miss out on logs between option-loading (now) and Init - i.e. native configuration
         options.SetupUnityLogging();
 
-        if (options.AttachViewHierarchy)
-        {
-            options.AddEventProcessor(new ViewHierarchyEventProcessor(options));
-        }
-        if (options.AttachScreenshot)
-        {
-            options.AddEventProcessor(new ScreenshotEventProcessor(options));
-        }
-
-        if (!application.IsEditor && options.Il2CppLineNumberSupportEnabled && unityInfo is not null)
-        {
-            options.AddIl2CppExceptionProcessor(unityInfo);
-        }
-
-        HandlePlatformRestrictedOptions(options, unityInfo, application);
+        // ExceptionFilters are added by default to the options.
         HandleExceptionFilter(options);
 
+        // The AnrDetectionIntegration is added by default. Since it is a ScriptableUnityOptions-only property we have to
+        // remove the integration when creating the options through here
         if (!AnrDetectionEnabled)
         {
             options.DisableAnrIntegration();
         }
 
         return options;
-    }
-
-    internal void HandlePlatformRestrictedOptions(SentryUnityOptions options, ISentryUnityInfo? unityInfo, IApplication application)
-    {
-        if (unityInfo?.IsKnownPlatform() == false)
-        {
-            options.DisableFileWrite = true;
-
-            // This is only provided on a best-effort basis for other than the explicitly supported platforms.
-            if (options.BackgroundWorker is null)
-            {
-                options.DiagnosticLogger?.LogDebug("Platform support for background thread execution is unknown: using WebBackgroundWorker.");
-                options.BackgroundWorker = new WebBackgroundWorker(options, SentryMonoBehaviour.Instance);
-            }
-
-            // Disable offline caching regardless whether it was enabled or not.
-            options.CacheDirectoryPath = null;
-            if (EnableOfflineCaching)
-            {
-                options.DiagnosticLogger?.LogDebug("Platform support for offline caching is unknown: disabling.");
-            }
-
-            // Requires file access, see https://github.com/getsentry/sentry-unity/issues/290#issuecomment-1163608988
-            if (options.AutoSessionTracking)
-            {
-                options.DiagnosticLogger?.LogDebug("Platform support for automatic session tracking is unknown: disabling.");
-                options.AutoSessionTracking = false;
-            }
-
-            return;
-        }
-
-        // Only assign the cache directory path if we're on a "known" platform. Accessing `Application.persistentDataPath`
-        // implicitly creates a directory and leads to crashes i.e. on the Switch.
-        options.CacheDirectoryPath = EnableOfflineCaching ? application.PersistentDataPath : null;
     }
 
     private void HandleExceptionFilter(SentryUnityOptions options)
--- a/src/Sentry.Unity/SentryUnityOptions.cs
+++ b/src/Sentry.Unity/SentryUnityOptions.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using Sentry.Unity.Integrations;
 using Sentry.Extensibility;
+using Sentry.Unity.NativeUtils;
 using UnityEngine;
 using CompressionLevel = System.IO.Compression.CompressionLevel;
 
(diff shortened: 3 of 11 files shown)


src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -227,23 +233,11 @@ internal SentryUnityOptions ToSentryUnityOptions(bool isBuilding, ISentryUnityIn
...
-        {
-            options.AddEventProcessor(new ScreenshotEventProcessor(options));
-        }
-
-        if (!application.IsEditor && options.Il2CppLineNumberSupportEnabled && unityInfo is not null)
-        {
-            options.AddIl2CppExceptionProcessor(unityInfo);
-        }
-
-        HandlePlatformRestrictedOptions(options, unityInfo, application);
```
bitsandfoxes: This got moved into `SentryUnitySdk.Init` to keep the SDKs behaviour consistent.

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -252,41 +246,6 @@ internal SentryUnityOptions ToSentryUnityOptions(bool isBuilding, ISentryUnityIn
         return options;
     }
 
-    internal void HandlePlatformRestrictedOptions(SentryUnityOptions options, ISentryUnityInfo? unityInfo, IApplication application)
```
bitsandfoxes: Same as above. Even if users call `Init` the SDK should make an effort not to crash the game.

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -294,39 +295,38 @@ internal string? DefaultUserId
 
     internal List<string> SdkIntegrationNames { get; set; } = new();
 
-    public SentryUnityOptions() : this(false, ApplicationAdapter.Instance) { }
-
-    internal SentryUnityOptions(bool isBuilding, IApplication application, ISentryUnityInfo? unityInfo = null) :
-        this(SentryMonoBehaviour.Instance, application, isBuilding, unityInfo)
```
bitsandfoxes, line 304: Got rid of a redundant `internal` constructor.

src/Sentry.Unity/SentryUnityOptionsExtensions.cs
```diff
@@ -63,18 +63,6 @@ internal static void SetupUnityLogging(this SentryUnityOptions options)
...
-    {
-        if (unityInfo.Il2CppMethods is not null)
-        {
-            options.AddExceptionProcessor(new UnityIl2CppEventExceptionProcessor(options, unityInfo));
-        }
-        else
-        {
-            options.DiagnosticLogger?.LogWarning("Failed to find required IL2CPP methods - Skipping line number support");
-        }
-    }
```
bitsandfoxes: Moved to `Init`.

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -137,4 +125,77 @@ public void CaptureFeedback(string message, string? email, string? name, bool ad
...
+                FileAccess.ReadWrite, FileShare.None);
+        }
+        catch (Exception ex)
+        {
+            options.DiagnosticLogger?.LogWarning("An exception was thrown while trying to " +
+                                                 "acquire a lockfile on the config directory: .NET event cache will be disabled.", ex);
+            options.CacheDirectoryPath = null;
+            options.AutoSessionTracking = false;
+        }
+    }
```
bitsandfoxes: To keep the `Init` more readable.

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -137,4 +125,77 @@ public void CaptureFeedback(string message, string? email, string? name, bool ad
...
+            options.Il2CppLineNumberSupportEnabled)
+        {
+            if (SentryPlatformServices.UnityInfo.Il2CppMethods is not null)
+            {
+                options.AddExceptionProcessor(new UnityIl2CppEventExceptionProcessor(options));
+            }
+            else
+            {
+                options.DiagnosticLogger?.LogWarning("Failed to find required IL2CPP methods - Skipping line number support");
+            }
```
bitsandfoxes: The `IL2CPP Exception Processor` is critical and "dangerous" as it calls into the IL2CPP backend and requires certain functions to be there. These functions change depending of Unity versions and are provided via  `IUnityInfo`. As such, this does not get added by default but late during Init with extensive checks.

bruno-garcia:
> This is a follow-up on https://github.com/getsentry/sentry-unity/pull/2227 as this was getting already kinda big.

"kinda" 😅

src/Sentry.Unity/Il2CppEventProcessor.cs
```diff
@@ -17,12 +17,15 @@ internal class UnityIl2CppEventExceptionProcessor : ISentryEventExceptionProcess
...
-    public UnityIl2CppEventExceptionProcessor(SentryUnityOptions options, ISentryUnityInfo unityInfo)
+    public UnityIl2CppEventExceptionProcessor(SentryUnityOptions options)
     {
         Options = options;
-        UnityInfo = unityInfo;
-        _il2CppMethods = unityInfo.Il2CppMethods ?? throw new ArgumentNullException(nameof(unityInfo.Il2CppMethods),
-            "Unity IL2CPP methods are not available.");
+
+        // We're throwing here but this should never happen. We're validating UnityInfo before adding the processor.
+        UnityInfo = SentryPlatformServices.UnityInfo
```
bruno-garcia: great! this was my only feedback in the other PR, validation that we're at the right state before going ahead

Also note that you copy a reference to use the same/consistent reference. This also means there's no path for a user to change these values after-the-fact.

But I understand that's the goal, and we then should follow this pattern when it makes sense. Instead of accessing the static mutable property in many places try to keep it to a minimum.

src/Sentry.Unity/SentryUnityOptions.cs
```diff
@@ -294,39 +295,38 @@ internal string? DefaultUserId
...
+    public SentryUnityOptions() :
+        this(false, ApplicationAdapter.Instance, SentryPlatformServices.UnityInfo, SentryMonoBehaviour.Instance)
```
bruno-garcia: we access `SentryPlatformServices.UnityInfo` here but also in other places where they take SentryOptions but also `ISentryUnityInfo`. This makes it a lot more confusing.

Can we then only get it from here?

Once you do `new Options` we read `SentryPlatformServices.UnityInfo` and if it's null, we throw. Or go into the mode where this doesn't exist if that's a thing. But from that point any access to UnityInfo is through options. If you find references of the getter for `SentryPlatformServices.UnityInfo` you'd only see tests and this contructor.

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -30,24 +31,11 @@ private SentryUnitySdk(SentryUnityOptions options)
...
-                options.DiagnosticLogger?.LogWarning("An exception was thrown while trying to " +
-                                                     "acquire a lockfile on the config directory: .NET event cache will be disabled.", ex);
-                options.CacheDirectoryPath = null;
-                options.AutoSessionTracking = false;
-            }
-        }
+        // Some integrations are controlled through a flag and opt-in. Adding these integrations late so we have the
+        // same behaviour whether we're self or manually initializing
+        HandleLateIntegrations(options);
+        HandlePlatformRestrictedOptions(options, SentryPlatformServices.UnityInfo);
```
bruno-garcia: example where we take options and also access the static mutable public property

bruno-garcia: btw, I'm saying "static public mutable property" a lot because it's a thing you should be aware, I didn't read it btw just got the first link: 

https://www.reddit.com/r/learnprogramming/comments/7nzlec/why_are_mutable_static_variables_bad/

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -137,4 +125,77 @@ public void CaptureFeedback(string message, string? email, string? name, bool ad
 
         Sentry.SentrySdk.CurrentHub.CaptureFeedback(message, email, name, hint: hint);
     }
+
+    internal static void HandleWindowsPlayer(SentryUnitySdk unitySdk, SentryUnityOptions options)
```
bruno-garcia: Not sure this method is called. "Handle"? It seems it's "Setting Up"? or Initializing?

It's being called by `Init` so perhaps `InitWindowsPlayer`?

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -137,4 +125,77 @@ public void CaptureFeedback(string message, string? email, string? name, bool ad
...
+                options.AddExceptionProcessor(new UnityIl2CppEventExceptionProcessor(options));
+            }
+            else
+            {
+                options.DiagnosticLogger?.LogWarning("Failed to find required IL2CPP methods - Skipping line number support");
+            }
+        }
+    }
+
+    internal static void HandlePlatformRestrictedOptions(SentryUnityOptions options, ISentryUnityInfo? unityInfo)
```
bruno-garcia: Again method name is off, seems the guard is: `IsKnownPlatform` so how about:

`InitUnknownPlatform`

bruno-garcia: It's a bit odd to see on the callsite:

```
InitWindows
InitOtherPlatform
```

Since you'd expect the guards to be outside the functions as the function names already 'declare' what they are supposed to be for.

But this is all internal code so not worth splitting hairs 🙈 And I don't have a great suggestion on how to improve this other than maybe trying yet other method names "TryInit" or whatever

test/Sentry.Unity.Tests/ScriptableSentryUnityOptionsTests.cs
```diff
@@ -178,42 +166,6 @@ public void ToSentryUnityOptions_UnityBadGatewayExceptionFilterAdded()
...
-        scriptableOptions.HandlePlatformRestrictedOptions(options, _fixture.UnityInfo, _fixture.Application);
-
-        Assert.IsTrue(options.DisableFileWrite);
-        Assert.IsNull(options.CacheDirectoryPath);
-        Assert.IsFalse(options.AutoSessionTracking);
-        Assert.IsTrue(options.BackgroundWorker is WebBackgroundWorker);
-    }
-
-    [Test]
-    public void HandlePlatformRestrictedOptions_KnownPlatform_SetsRestrictedOptions()
```
bruno-garcia, line 205: this was lost on purpose?

test/Sentry.Unity.Tests/ScriptableSentryUnityOptionsTests.cs
```diff
@@ -127,18 +127,6 @@ public void ToSentryUnityOptions_HasOptionsConfiguration_GetsCalled(bool isBuild
         Assert.AreEqual(optionsConfiguration.GotCalled, !isBuilding);
     }
 
-    [Test]
-    public void ToSentryUnityOptions_UnknownPlatforms_DoesNotAccessDisk()
```
bruno-garcia, line 131: This was also deleted on purpsoe?

src/Sentry.Unity/SentryUnityOptions.cs
bitsandfoxes: This was addressed and fixed in https://github.com/getsentry/sentry-unity/pull/2227

src/Sentry.Unity/SentryUnitySdk.cs
bitsandfoxes: Thanks. Addressed it.

test/Sentry.Unity.Tests/ScriptableSentryUnityOptionsTests.cs
bitsandfoxes: Since it's no longer done in the `ScriptableOptions` but part of the the `Init` I moved the tests there too.

test/Sentry.Unity.Tests/ScriptableSentryUnityOptionsTests.cs
bitsandfoxes: Since it's no longer done in the `ScriptableOptions` but part of the the `Init` I moved the tests there too.

src/Sentry.Unity/SentryUnitySdk.cs
bitsandfoxes: I gave the bunch some more descriptive names.

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -141,4 +126,77 @@ public void CaptureFeedback(string message, string? email, string? name, bool ad
...
+        // https://forum.unity.com/threads/unsupported-internal-call-for-il2cpp-mutex-createmutex_internal-named-mutexes-are-not-supported.387334/
+        if (ApplicationAdapter.Instance.Platform is not RuntimePlatform.WindowsPlayer ||
+            options.CacheDirectoryPath is null)
+        {
+            return;
+        }
+
+        try
+        {
+            unitySdk._lockFile = new FileStream(Path.Combine(options.CacheDirectoryPath, "sentry-unity.lock"), FileMode.OpenOrCreate,
```
bruno-garcia: oh this looks odd. We're accessing a field from that instance in here? If that's `internal` name it `LockFIle` or something (Pascal Case)

bruno-garcia (approved):

Issue #2224, opened 2025-06-30
Attached screenshots blank


bitsandfoxes:
## Problem

Events captured by the SDK some times contain blank or black screenshots. This obviously makes them less than useful.

## Context

Based on the [Unity documentation](https://docs.unity3d.com/6000.1/Documentation/ScriptReference/ScreenCapture.CaptureScreenshotAsTexture.html), capturing screenshots should be done at the end of rendering - typically `EndOfFrame`. Capturing screenshots at any time before that leads to unpredictable and undefined results. Our [`ScreenshotEventProcessor`](https://github.com/getsentry/sentry-unity/blob/f770e6cd8c977ddded2d6ad8640f57151e5b710e/src/Sentry.Unity/ScreenshotEventProcessor.cs) does exactly that. It captures the screen at the exact time of the error.

## Proposal

The fix for this is less than straight forward. We cannot halt the SDK's execution until the end-of-frame. And we cannot use Unity's `yield new WaitForEndOfFrame()` from within the SDK.

### Attachments can be sent separately

The idea now is to separate screenshot-capture from event-processing. A possible approach would be to have the `ScreenshotEventProcessor` note the event-ID and pass this on to our `SentryMonoBehaviour` where we do have access to `yield new WaitForEndOfFrame()` and send the attachment for event-ID the processor held.

This comes with the issue that the eventprocessor has no way of knowing whether the event got dropped or not and not having access to the hub to fetch the `LastEventId`. But `SdkIntegrations` do have access to the hub.

A possible solution could be

1. The `ScreenshotSdkIntegration` hooks into the update loop for `EndOfFrame` via adding itself to `PlayerLoop.GetCurrentPlayerLoop()`
2. The `ScreenshotProcessor` stores the event ID
3. In the next `EndOfFrame` update the `ScrenshotSdkIntegration` checks the stored ID vs the `LastId` and captures a screenshot.

While at it, we can take a look at [reading pixels asynchronously](https://docs.unity3d.com/6000.1/Documentation/ScriptReference/ScreenCapture.CaptureScreenshotIntoRenderTexture.html) when capturing a screenshot.

Note, that the `PlayerLoop.GetCurrentPlayerLoop()` is only accessible starting from version `2021 and newer`.

Issue #1827, opened 2024-10-07
Don't attach screenshots on events coming from the same frame


bruno-garcia:
When Attach Screenshot is enabled, Sentry will attach a screenshot to all events. At firs this is the wanted behavior. 
The problem comes from when an error happens on every frame. Resulting in lots of events with screenshots of basically the same game view.

Some ideas on how to improve this:
We should consider not attaching a screenshot if the UI didn't change from the previous error (e.g: same frame). Or perhaps if the timeframe between errors is to short.

Note:
* [Event deboucing is disabled by default](https://docs.sentry.io/platforms/unity/configuration/event-debouncing/)

bitsandfoxes:
Depends on https://github.com/getsentry/sentry-unity/issues/745

I imagine we could have the screenshots be added by some sort of processor that keeps track of when a screenshot was taken last and skip if it's too early.

bruno-garcia:
Here's an example when that happens: https://sentry.sentry.io/issues/5987813063/attachments/?attachmentFilter=screenshot&node=error-cfa4cebfdec847f4910ba5dde06201fd&project=5905698&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=1h&stream_index=0

Couple hundred events, all with screenshots, a lot quite similar/same:
![Image](https://github.com/user-attachments/assets/c80f0d90-3f20-464c-8325-e324c31328a4)

![Image](https://github.com/user-attachments/assets/749cc650-ce8f-4029-9260-4b7163451877)

bitsandfoxes:
Side note: If we slow it down enough we might have something like a session replay! Not really.

bitsandfoxes:
Resolved with #2240

PR #2240, opened 2025-07-10
fix: Screenshot Capture

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/2224, https://github.com/getsentry/sentry-unity/issues/1827 🎉 
Relies on the .NET SDK https://github.com/getsentry/sentry-dotnet/pull/4357

### Context

The SDK has to wait for `EndOfFrame` before capturing a screenshot as the behaviour is unreliable otherwise. See [Unity docs](https://docs.unity3d.com/6000.1/Documentation/ScriptReference/ScreenCapture.CaptureScreenshotAsTexture.html).

### Implementation

The `ScreenshotProcessor` uses the `SentryMonoBehaviour` to start a coroutine that waits for `End of Frame` to capture the screenshot. The screenshot gets sent as an envelop with a single attachment as envelope item.


src/Sentry.Unity/ScreenshotEventProcessor.cs
```diff
@@ -4,61 +4,22 @@
...
-        _application = application ?? ApplicationAdapter.Instance;
-    }
+    public ScreenshotEventProcessor(SentryUnityOptions sentryOptions) : this(sentryOptions, SentryMonoBehaviour.Instance) { }
 
-    public SentryEvent? Process(SentryEvent @event)
+    internal ScreenshotEventProcessor(SentryUnityOptions sentryOptions, SentryMonoBehaviour? sentryMonoBehaviour)
     {
-        return @event;
+        _options = sentryOptions;
+        _sentryMonoBehaviour = sentryMonoBehaviour ?? SentryMonoBehaviour.Instance;
```
bruno-garcia: We have a public ctor that passes `SentryMonoBehaviour.Instance` already, having also `??` seems redundant.

Since this is an internal ctor, how about:
```suggestion
        _sentryMonoBehaviour = sentryMonoBehaviour ?? throw new ArgumentNullException("sentryMonoBehaviour");
```

src/Sentry.Unity/SentryMonoBehaviour.cs
```diff
@@ -112,3 +113,26 @@ private void Awake()
...
+/// <summary>
+/// A MonoBehaviour that captures screenshots
+/// </summary>
+public partial class SentryMonoBehaviour
+{
+    // Todo: keep track of event ID to not capture double/more than once per frame
+
+    public void CaptureScreenshotForEvent(SentryUnityOptions options, SentryId eventId)
+    {
+        StartCoroutine(CaptureScreenshot(options, eventId));
```
bruno-garcia: This method returns before the coroutine starts/ends, right? how do we know the screenshot will be taken before the event is created?

src/Sentry.Unity/ScreenshotEventProcessor.cs
bitsandfoxes: Fixed the constructors.

src/Sentry.Unity/SentryMonoBehaviour.cs
bitsandfoxes: The `ScreenshotProcessor` forwards the call to the `SentryMonoBehaviour` that starts the coroutine for the current EndOfFrame.

bruno-garcia:
CI is unhappy

src/Sentry.Unity/ScreenshotEventProcessor.cs
```diff
@@ -1,64 +1,69 @@
...
-    private readonly IApplication _application;
-    public ScreenshotEventProcessor(SentryUnityOptions sentryOptions) : this(sentryOptions, null) { }
+    private readonly ISentryMonoBehaviour _sentryMonoBehaviour;
+    private bool _isCapturingScreenshot;
 
-    internal ScreenshotEventProcessor(SentryUnityOptions sentryOptions, IApplication? application)
+    internal Func<SentryUnityOptions, byte[]> ScreenshotCaptureFunction = SentryScreenshot.Capture;
+    internal Action<SentryId, SentryAttachment> AttachmentCaptureFunction = (eventId, attachment) =>
+        ((Hub)Sentry.SentrySdk.CurrentHub).CaptureAttachment(eventId, attachment);
+    internal Func<YieldInstruction> WaitForEndOfFrameFunction = () => new WaitForEndOfFrame();
```
bitsandfoxes: `WaitForEndOfFrame()` does not work in the Editor and cannot be properly tested otherwise.

src/Sentry.Unity/ScreenshotEventProcessor.cs
```diff
@@ -1,64 +1,69 @@
...
     private readonly SentryUnityOptions _options;
-    private readonly IApplication _application;
-    public ScreenshotEventProcessor(SentryUnityOptions sentryOptions) : this(sentryOptions, null) { }
+    private readonly ISentryMonoBehaviour _sentryMonoBehaviour;
+    private bool _isCapturingScreenshot;
 
-    internal ScreenshotEventProcessor(SentryUnityOptions sentryOptions, IApplication? application)
+    internal Func<SentryUnityOptions, byte[]> ScreenshotCaptureFunction = SentryScreenshot.Capture;
+    internal Action<SentryId, SentryAttachment> AttachmentCaptureFunction = (eventId, attachment) =>
+        ((Hub)Sentry.SentrySdk.CurrentHub).CaptureAttachment(eventId, attachment);
```
bitsandfoxes: It's not pretty but this way I can replace these functions during tests.
It's especially annoying since I can't use the `HubAdapter` here because `CaptureAttachment` is `internal` on the actual hub.

test/Scripts.Integration.Test/Scripts/SmokeTester.cs
```diff
@@ -92,92 +94,108 @@ private static string GetTestArg()
 
     internal static Func<int> CrashedLastRun = () => -1;
 
-    public static void SmokeTest()
+    private IEnumerator SmokeTestCoroutine()
     {
         t.Start("SMOKE");
-        try
```
bitsandfoxes, line 98: Can't have `yield` inside a `try`. Turns out the `catch` is not even necessary since we're printing the expected fail-message in `Exit`.

test/Scripts.Integration.Test/Scripts/SmokeTester.cs
```diff
@@ -259,10 +277,7 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
         private void Receive(HttpRequestMessage message)
         {
             var msgText = message.Content.ReadAsStringAsync().Result;
-            // Setting "Sentry" as tag to prevent the UnityLogHandlerIntegration from capturing this message and
-            // adding it as a breadcrumb, which in turn multiplies it on following (intercepted) HTTP requests...
-            // Note: remove the prefix once setting breadcrumb log level is possible - https://github.com/getsentry/sentry-unity/issues/60
```
bitsandfoxes, line 264: Yay - cleanup time! The messages are now filtered in the OptionsConfiguration based on `LogType`.

src/Sentry.Unity/ScreenshotEventProcessor.cs
bruno-garcia: This means if Sentry is disabled and this is called, it'll throw and exception. You'd want instead:

```suggestion
    internal Action<SentryId, SentryAttachment> AttachmentCaptureFunction = (eventId, attachment) =>
        (Sentry.SentrySdk.CurrentHub as Hub)?.CaptureAttachment(eventId, attachment) ?? SentryId.Empty;
```

src/Sentry.Unity/ScreenshotEventProcessor.cs
```diff
@@ -1,64 +1,69 @@
...
+        _sentryMonoBehaviour = sentryMonoBehaviour;
     }
 
-    public SentryEvent? Process(SentryEvent @event)
+    public SentryEvent Process(SentryEvent @event)
     {
+        // Only ever capture one screenshot per frame
+        if (!_isCapturingScreenshot)
+        {
+            _isCapturingScreenshot = true;
```
bruno-garcia: this code is not atomic, and event processor is called by `CaptureXyz` public API which can be called in any thread.
You probably want a `Interlocked.Exchange` with a `int` (or bool I think in newer .NETs which you can't access from Unity yet).

src/Sentry.Unity/SentryUnitySdk.cs
```diff
@@ -126,6 +127,26 @@ public void CaptureFeedback(string message, string? email, string? name, bool ad
...
+        try
+        {
+            if (Sentry.SentrySdk.CurrentHub is Hub hub)
+            {
+                hub.CaptureAttachment(eventId, attachment);
+                _options.LogDebug("Attachment captured for event {0}", eventId);
+            }
+            else
+            {
+                _options.LogError("Capturing the attachment failed due to the current hub.");
```
bruno-garcia:
```suggestion
                _options.LogError("Capturing attachment was not possible because the current hub is of type {0}.", Sentry.SentrySdk.CurrentHub?.GetType());
```

bruno-garcia:
@sentry review

bitsandfoxes:
@sentry review

test/Scripts.Integration.Test/Scripts/SmokeTester.cs
```diff
@@ -92,92 +94,103 @@ private static string GetTestArg()
...
+
+        // Wait for screenshot capture to complete
+        yield return null;
+
+        currentMessage++; // The error event
+
+        t.ExpectMessage(currentMessage, "'type':'event'");
+        t.ExpectMessage(currentMessage, $"LogError(GUID)={guid}");
+        // Contexts
+        t.ExpectMessage(currentMessage, "'type':'app',");
```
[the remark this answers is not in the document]
bitsandfoxes: Yeah, while I agree, this one is a bit of a bigger undertaking and explodes the scope of this PR.

test/Scripts.Integration.Test/Scripts/SmokeTester.cs
```diff
@@ -92,92 +94,103 @@ private static string GetTestArg()
...
+            if (t.CheckMessage(currentMessage, "'type':'transaction'"))
             {
-                if (t.CheckMessage(currentMessage, "'type':'transaction'"))
-                {
-                    break;
-                }
+                break;
             }
-            Debug.Log($"SMOKE TEST: Done skipping session requests. Last one was: #{currentMessage}");
+        }
```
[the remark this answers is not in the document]
bitsandfoxes: Not to worry. Screenshot capture blocks until the end of execution and `yield return null` makes us go back to the next frame.

scripts/smoke-test-android.ps1
```diff
@@ -24,7 +24,7 @@ Write-Host "#####################################################"
 
 $BuildDir = $(GetNewProjectBuildPath)
 $ApkFileName = "test.apk"
-$ProcessName = "com.DefaultCompany.$(GetNewProjectName)"
+$ProcessName = "io.sentry.unity.integrationtest"
```
bitsandfoxes, line 27: With Unity 6 integration tests seem to fail to have their proper identifier set, causing the build to fail. We're setting this now explicitly in the `Builder.cs` and making use of the hardcoded name here.

src/Sentry.Unity/ScreenshotEventProcessor.cs
```diff
@@ -1,64 +1,85 @@
...
     }
+
+    internal virtual byte[] CaptureScreenshot(SentryUnityOptions options)
+        => SentryScreenshot.Capture(options);
+
+    internal virtual void CaptureAttachment(SentryId eventId, SentryAttachment attachment)
+        => (Sentry.SentrySdk.CurrentHub as Hub)?.CaptureAttachment(eventId, attachment);
+
+    internal virtual YieldInstruction WaitForEndOfFrame()
+        => new WaitForEndOfFrame();
```
bitsandfoxes, line 84: I'm making these virtual so I can create a `TestScreenshotEventProcessor` that overwrites these for testing purposes.

test/Scripts.Integration.Test/Editor/Builder.cs
```diff
@@ -41,14 +43,6 @@ public static void BuildIl2CPPPlayer(BuildTarget target, BuildTargetGroup group,
...
 #endif
 
-
-        // This is a workaround for build issues with Unity 2022.3. and newer.
-        // https://discussions.unity.com/t/gradle-build-issues-for-android-api-sdk-35-in-unity-2022-3lts/1502187/10
-#if UNITY_2022_3_OR_NEWER
-        Debug.Log("Builder: Setting Android target API level to 33");
-        PlayerSettings.Android.targetSdkVersion = AndroidSdkVersions.AndroidApiLevel33;
-#endif
-
```
bitsandfoxes, line 51: Workaround that is no longer needed.

PR #2255, opened 2025-07-28
chore: Update to Unity `6000.0.54f1`

bitsandfoxes:
Bumped Unity 6 and fixed `update-deps.yml` in the process.

#skip-changelog


.github/workflows/update-deps.yml
```diff
@@ -69,7 +69,7 @@ jobs:
       - name: Update ci-env.ps1
         run: |
           $file = "scripts/ci-env.ps1"
-          $regexVersion = '${{ matrix.unity-prefix }}' + "[0-9.a-z]+"
+          $regexVersion = '${{ matrix.unity-prefix }}' + "\.[0-9]+\.[0-9]+[a-z][0-9]+"
```
bruno-garcia: Please assign all GH variables to an env var first, before using it in the `run` command.
like env:
 UNITY_PREFIX: ${{ matrix.unity-prefix }}

Then use it within double quotes `"$UNITY_PREFIX"`

bitsandfoxes: Doing this in a separate PR: https://github.com/getsentry/sentry-unity/pull/2257

scripts/ci-env.ps1
```diff
@@ -17,10 +17,10 @@ switch ($name) {
         return "2023.2.20f1"
     }
     "unity6000" {
-        return "6000.0.49f1"
+        return "6000.0.54f1"
     }
     "unity6100" {
-        return "6000.1.12f1"
+        return "6000.1.13f1"
```
bitsandfoxes:
```suggestion
        return "6000.1.12f1"
```

.github/workflows/update-deps.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        unity-prefix: ['2020', '2021', '2022', '6000']
+        unity-prefix: ['2020', '2021', '2022', '6000', '6100']
```
vaind: looks like this doesn't play well with the versions in ci-env.ps1 and the current solution is a bit hackish.

maybe change the declaration to:
```suggestion
        unity-prefix: ['2020', '2021', '2022', '6000.0', '6000.1']
```

vaind: I didn't mean do it for all the versions, but yeah, works 😁

bitsandfoxes: I know. But it ended up looking a different kind of weird. And I ended up having to special case the regex because regexing is hard. 😅

PR #2264, opened 2025-08-04
chore: Create `iOS-only` xcframework in CI

bitsandfoxes:


scripts/setup-cocoa-sdk.ps1
```diff
@@ -0,0 +1,143 @@
...
+    
+    [Parameter(Mandatory=$true)]
+    [string]$iOSDestination,
+    
+    [Parameter(Mandatory=$true)]
+    [string]$macOSDestination,
+    
+    [switch]$iOSOnly
+)
+
```
vaind, line 21:
```suggestion

Set-StrictMode -Version latest
$ErrorActionPreference = 'Stop'
$PSNativeCommandUseErrorActionPreference = $true

```

vaind (approved):

PR #2266, opened 2025-08-04
chore: Clean `package-release` before repacking

bitsandfoxes:
I'm running repack to manually update the `.snapshot` and it keeps getting stuck if the `package-release` stuff already exists. Well, not any longer it doesn't!

#skip-changelog

### DIFF
--- a/scripts/repack.ps1
+++ b/scripts/repack.ps1
@@ -1,12 +1,10 @@
 # Clean up previous release artifacts
 if (Test-Path "package-release") {
     Remove-Item -Path "package-release" -Recurse -Force
-    Write-Host "Removed package-release directory"
 }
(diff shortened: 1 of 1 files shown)


scripts/repack.ps1
```diff
@@ -1,3 +1,12 @@
+# Clean up previous release artifacts
+if (Test-Path "package-release") {
+    Remove-Item -Path "package-release" -Recurse -Force
```
mujacica, line 3: Potential permissions issue could happen? Do we need to handle that somehow?

bitsandfoxes: This is a helper script for local development only. I'm not worried about it.

Issue #2247, opened 2025-07-21
`MainThreadData.CollectData` is getting called twice


bitsandfoxes:
### Description

`MainTheadData.CollectData` is required to be for the SDK to be able to access certain functionality, i.e. scope-sync on Android, and some device data during event processing i.e. 
https://github.com/getsentry/sentry-unity/blob/fcd5e999a16bc71b214702ed04aa96d11f5381e9/src/Sentry.Unity/UnityEventProcessor.cs#L52-L57

To guarantee that the scope sync on Android is up and running safely we moved the collection into the platform specific `Configure` call 
https://github.com/getsentry/sentry-unity/blob/fcd5e999a16bc71b214702ed04aa96d11f5381e9/src/Sentry.Unity.Android/SentryNativeAndroid.cs#L25

With the rework to the Initialization Flow in https://github.com/getsentry/sentry-unity/pull/2227 we pulled invoking `Configure` into `Init`. So we can drop the collection duplication.

PR #2281, opened 2025-08-12
fix: Collect `MainThreadData` first thing in `Init`

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/2247

Since the native configuration now gets called from within the SDK it does not need to be pulled out into `Android.Configure` and can universally be the first thing the SDK does.

This works because Init is getting automatically called in `SentryInitialization.cs` https://github.com/getsentry/sentry-unity/blob/c46352d4a286e22e6e1fef6d84a2a69fc77e2cd7/package-dev/Runtime/SentryInitialization.cs#L58-L71

#skip-changelog

### DIFF
--- a/package-dev/Runtime/SentryInitialization.cs
+++ b/package-dev/Runtime/SentryInitialization.cs
@@ -57,8 +57,8 @@ namespace Sentry.Unity
 #endif
         private static void Init()
         {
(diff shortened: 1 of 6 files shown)


limbonaut (approved):

limbonaut:
LGTM

Issue #2195, opened 2025-06-15
Sentry 'Upload Symbols Task' is added on each build producing duplicate task build error


Awais6:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)
sentry.io

Which version of the SDK?

How did you install the package? (Git-URL, Assetstore)

Which version of Unity?

Is this happening in Unity (editor) or on a player like Android, iOS, Windows?

### Steps to Reproduce

1) Install Sentry and Configure it in fresh Unity Project.
2) Build project for Android (Should be a successful build)
3) Again Build Project for Android

### Expected Result

What you thought would happen.

### Actual Result

What actually happened. Maybe a screenshot/recording? Maybe some logs?

### Any logs or screenshots

below task is added multiple times after first successful build

task sentryUploadSymbols {
*
*
*
}
// Autogenerated Sentry symbol upload task [end]

bruno-garcia:
Thanks for the details. This is odd since we check before adding the target. 

do you have a custom gradle file?

Issue #2170, opened 2025-05-22
multiple sentryUploadSymbols added to build.gradle file


dcohenb:
I have a recurring issue where the `sentryUploadSymbols` is being added to the `build.gradle` file more then once resulting in failed builds.
it seems there is a missing check in the `DebugSymbolUpload.cs` to see if the task has already been added.
unfortunately I cannot patch it as it's part of the dll

origin: https://github.com/getsentry/sentry-unity/blob/cc4a46795ad34621bbf870ef3d722cebe62c4dcd/src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs#L165

### Environment

Sentry SaaS with Unity SDK

`Which version of the SDK?`
v3.2.2

`How did you install the package? (Git-URL, Assetstore)`
Installed via package manager with Git-URL

`Which version of Unity?`
6000.0.41f

`Is this happening in Unity (editor) or on a player like Android, iOS, Windows?`
Editor - Causing builds to fail

### Steps to Reproduce

1. Run a build more then once

### Expected Result

successful build

### Actual Result

Build failing due to duplicate `sentryUploadSymbols` task in the `build.gradle` file

### Any logs or screenshots

```
stderr[

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/...ProjectPath.../Library/Bee/Android/Prj/IL2CPP/Gradle/launcher/build.gradle' line: 137

* What went wrong:
A problem occurred configuring project ':launcher'.
> Cannot add task 'sentryUploadSymbols' as a task with that name already exists.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 5s
]
stdout[

> Configure project :unityLibrary
Variant 'debug', will keep symbols in binaries for:
  'libunity.so'
  'libil2cpp.so'
  'libmain.so'
Variant 'release', symbols will be stripped from binaries.

> Configure project :launcher
Variant 'debug', will keep symbols in binaries for:
  'libunity.so'
  'libil2cpp.so'
  'libmain.so'
Variant 'release', symbols will be stripped from binaries.
]
exit code: 1
```

bruno-garcia:
thanks for raising this. Would be up to making a PR? If not, that's OK we'll get to it eventually but might take a few days before we get to it

carlossalasamper:
[...]

PR #2300, opened 2025-08-27
fix: Prevent appending `uploadTask` more than once

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/2195, https://github.com/getsentry/sentry-unity/issues/2170

### Problem

I still fail to reproduce this on both macOS and Windows. But given that the issue is appending the upload task multiple times there are only two places where this could fail:

1. The check whether the upload task is there is insufficient https://github.com/getsentry/sentry-unity/blob/dbbb63c2086bc21055c79825c28888cbeeb37f6e/src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs#L185
2. Removing the task is failing https://github.com/getsentry/sentry-unity/blob/dbbb63c2086bc21055c79825c28888cbeeb37f6e/src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs#L191-L192

### Proposal

The SDK is needs to be more defensive about adding the upload task to the Gradle file and prevent duplicated appending.

1. Check for the actual marker instead of the `.properties`
2. Update the regex to handle line endings between the markers (this is the only thing I can think of: The SDK fails to properly remove the entire task due to `regex.Replace` not doing what we expect it to do)
3. Add a sanity check before adding the task with proper logging how to unblock yourself


src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
```diff
@@ -304,7 +304,7 @@ internal void SetupSymbolsUpload(string unityProjectPath, string gradleProjectPa
         {
             symbolsUpload.RemoveUploadFromGradleFile();
 
-            if (_options is { Il2CppLineNumberSupportEnabled: true })
+            if (_options is { Enabled: true, Il2CppLineNumberSupportEnabled: true })
```
bitsandfoxes, line 307: Unrelated but: To reduce logging noise when the SDK is disabled.

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
```diff
@@ -102,8 +103,6 @@ public DebugSymbolUpload(IDiagnosticLogger logger,
 
     public void AppendUploadToGradleFile(string sentryCliPath)
     {
-        RemoveUploadFromGradleFile();
```
bitsandfoxes, line 105: This now happens outside and before calling this method. Added a sanity check before actually writing to disk.

bitsandfoxes:
@sentry review

bitsandfoxes:
bugbot review

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
```diff
@@ -182,15 +190,23 @@ public void RemoveUploadFromGradleFile()
...
-        if (!gradleBuildFile.Contains("sentry.properties"))
+        if (!gradleBuildFile.Contains(SymbolUploadTaskStartComment))
         {
             _logger.LogDebug("No previous upload task found.");
             return;
         }
 
-        var regex = new Regex(Regex.Escape(SymbolUploadTaskStartComment) + ".*" + Regex.Escape(SymbolUploadTaskEndComment), RegexOptions.Singleline);
+        var pattern = Regex.Escape(SymbolUploadTaskStartComment) + @"\r?\n.*?\r?\n" + Regex.Escape(SymbolUploadTaskEndComment);
+        var regex = new Regex(pattern, RegexOptions.Singleline);
```
vaind: This does not look right, why is the change necessary? `RegexOptions.Singleline` already makes regex match `\n` by `.`.

vaind (changes requested):

src/Sentry.Unity.Editor/Android/DebugSymbolUpload.cs
bitsandfoxes: Ah, good catch. I've reduced it to a non-greedy `.*?` to remove any amount of duplicate entries.

Issue #2235, opened 2025-07-09
Trace ID generation duplication


bitsandfoxes:
### Description

There are now multiple integrations creating and generating a new trace.
1. [Trace Generation Integration](https://github.com/getsentry/sentry-unity/blob/main/src/Sentry.Unity/Integrations/TraceGenerationIntegration.cs)
2. [Scene Manager Tracing](https://github.com/getsentry/sentry-unity/blob/main/src/Sentry.Unity/Integrations/SceneManagerTracingIntegration.cs)
3. [Startup Tracing Integration](https://github.com/getsentry/sentry-unity/blob/606140df87e2fce8753ce9c18cc22e4e1c5f59c3/package-dev/Runtime/SentryIntegrations.cs#L32)

To remove resolve duplicate, redundant, and erroneous creation the first two could no-op during startup. See https://github.com/getsentry/sentry-unity/pull/2234

PR #2301, opened 2025-08-28
fix: Deduplicate trace generation during startup and scene loading

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/2235

The flow differs from Editor vs actual build since not all RuntimeInitialization attributes get triggered.

1. Init gets triggered by `RuntimeInitialization SubsystemRegistration` https://github.com/getsentry/sentry-unity/blob/dbbb63c2086bc21055c79825c28888cbeeb37f6e/package-dev/Runtime/SentryInitialization.cs#L68
2. The `SentryUnitySdk` initializes the .NET SDK https://github.com/getsentry/sentry-unity/blob/dbbb63c2086bc21055c79825c28888cbeeb37f6e/src/Sentry.Unity/SentryUnitySdk.cs#L38
3. All integrations get registered
    - **NEW** The `TraceGenerationIntegration` refrains from regenerating any traces that would be covered by other integrations
4. The `StartupTracingIntegration` creates a transaction and it's child spans https://github.com/getsentry/sentry-unity/blob/dbbb63c2086bc21055c79825c28888cbeeb37f6e/src/Sentry.Unity/SentryUnitySdk.cs#L42
5. The first scene loads. 
    - Note: For the first scene, the [SceneManagerTracingIntegration](https://github.com/getsentry/sentry-unity/blob/dbbb63c2086bc21055c79825c28888cbeeb37f6e/src/Sentry.Unity/Integrations/SceneManagerTracingIntegration.cs) is not getting invoked. This happens outside of the `SceneManagerAPI`.
    - This means there is no need to deduplicate between `StartupTracing` and `SceneManagerTracing`
6. For all subsequent scene loading, the `SceneManagerTracingIntegration` is solely responsible for generating a transaction. Finishing the transaction generates a new trace implicitely.

### DIFF
--- a/src/Sentry.Unity/Integrations/SceneManagerTracingIntegration.cs
+++ b/src/Sentry.Unity/Integrations/SceneManagerTracingIntegration.cs
@@ -35,24 +35,12 @@ public class SceneManagerTracingAPI : SceneManagerAPI
     public SceneManagerTracingAPI(IDiagnosticLogger? logger) =>
         _logger = logger;
 
-    protected override UnityEngine.AsyncOperation LoadSceneAsyncByNameOrIndex(string sceneName, int sceneBuildIndex, LoadSceneParameters parameters, bool mustCompleteNextFrame)
+    protected override AsyncOperation LoadSceneAsyncByNameOrIndex(string sceneName, int sceneBuildIndex, LoadSceneParameters parameters, bool mustCompleteNextFrame)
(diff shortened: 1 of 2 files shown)


bitsandfoxes:
@sentry review

bitsandfoxes:
@sentry review

PR #2313, opened 2025-09-11
feat: Drop Unity 2020

bitsandfoxes:
Unity 2020 reached EOL 2023 with no xLTS being available.

This has the following benefits:
1. All those `#if UNITY_2021` checks can go away
2. We no longer need to pollute `UnityInfo` with `IsKnownPlatform` and `IsNativeSupportEnabled` since those enums now baseline exist
3. We can make that functionality `internal`
4. Cleaner all around, making it easier to add additional support i.e. Xbox

I also recreated the `unity-of-bugs` sample from a new `2021` project so we have the proper default setup and settings that come OOTB instead of having updated the project twice (from 2019 to 2021).


mujacica:
CI Failure here expected?

bitsandfoxes:
> CI Failure here expected?

I would not use the word "expected". This is the smoke tests running on the Android emulator being flaky. :(
I'll have to address this outside of this PR

PR #2317, opened 2025-09-15
chore: Update Unity versions in CI 

bitsandfoxes:
### Problem

The current CI workflow tests all Unity versions (2021.3, 2022.3, 6000.0, 6000.1) across all platforms for every pull request. With only so many build server licenses, if there are multiple PRs they start blocking each other.

Minor annoyance that I ran into again when dropping support for Unity 2020: We've got the versions to test in a whole lot of different place. Ideally, this would be unified somewhere.

## Proposal

Split versions to be tested in 3 groups:
- `main` branch: Test everything: `2021`, `2022`, `6.0`, `6.1`. We'll need to look into `6.2` in the near future.
- `PR` branch: Test against the highest impacted versions: `2022`, `6.0`. We don't have any `#ifdef` as part of the initialization anymore, making this much saver.
- `Release` branches: Keep as is, a sanity check

Come up with a unified mechanism that creates the Unity-version-platform-matrix.

## Implementation

For this we have to:

1. Create matrix creation workflow `.github/workflows/create-unity-matrix.yml`:
    - Holds and defines which Unity versions to use based on where the action was triggered from
2. Updated CI workflow `.github/workflows/ci.yml`:
    - Replaced hardcoded Unity version arrays with dynamic matrices
    - All smoke test jobs now depend on the matrix creation job
3. Updated Build Workflow (.github/workflows/build.yml):
    - Removed Unity version restrictions on packaging steps. We're only building with one now.

## Impact

Pull Requests (50% reduction):
- Before: 4 Unity versions × multiple platforms = ~16-24 test combinations **per** platform
- After: 2 Unity versions × multiple platforms = ~8-12 test combinations **per** platform

Main Branch (no change):
- Continues testing all supported Unity versions for full compatibility validation

## Notes

- Reduced License Pressure: 50% fewer Unity license checkouts during PR validation
- Faster PR Feedback: Reduced CI runtime for pull request validation
- Less chance for flake
- Maintained Quality: Full version coverage still runs on main branch

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,6 @@ on:
         type: string
 
 env:
-  LOWEST_SUPPORTED_UNITY_VERSION: 2021.3
   DOTNET_CLI_TELEMETRY_OPTOUT: 1
   DOTNET_NOLOGO: 1
   GITHUB_ACTOR: ${{ github.actor }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -7,7 +7,6 @@ on:
         type: string
 
 env:
-  LOWEST_SUPPORTED_UNITY_VERSION: 2021.3
```
bitsandfoxes, line 10: Since we dropped `2019` and `2020`, we don't have any special casing for which version to package anymore and this defaults to `2021`.

.github/workflows/ci.yml
```diff
@@ -28,6 +27,11 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
+  create-unity-matrix:
+    uses: ./.github/workflows/create-unity-matrix.yml
+    with:
+      event-name: ${{ github.event_name }}
+
```
bitsandfoxes, line 34: This now needs to run before any of our smoke testing.

PR #2329, opened 2025-09-23
fix: Xbox debug symbol upload

bitsandfoxes:
Followup:
- Add Xbox platform to automated symbol upload
- Fix `Sentry.Unity.Native` to target Xbox Series platforms
- Fix `executable path` when targeting Xbox

### DIFF
--- a/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
+++ b/src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
@@ -43,8 +43,16 @@ public static class BuildPostProcess
         var isMono = PlayerSettings.GetScriptingBackend(targetGroup) == ScriptingImplementation.Mono2x;
 #pragma warning restore CS0618
 
-        var executableName = Path.GetFileName(executablePath);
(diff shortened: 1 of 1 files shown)


modules/sentry-java
vaind, line 1: Unrelated?

bitsandfoxes: Yes.

src/Sentry.Unity.Editor/Native/BuildPostProcess.cs
```diff
@@ -97,7 +105,7 @@ private static void AddCrashHandler(IDiagnosticLogger logger, BuildTarget target
                 return;
             case BuildTarget.GameCoreXboxSeries:
             case BuildTarget.GameCoreXboxOne:
-                // TODO: Figure out if we need to ship with a crash handler
+                // No standalone crash handler for Xbox - comes with Breakpad
```
bruno-garcia, line 108: Note that we plan to move to crashpad. 
I'll add a link to this comment to that ticket

PR #2337, opened 2025-09-26
chore: Remove deprecated runtime/build time configuration

bitsandfoxes:
### Description

The changes in https://github.com/getsentry/sentry-unity/pull/1888 replaced the runtime/build time configuration setup into one `Option Configuration` setup. The same callback would get invoked wether it is on runtime or at build time.

We kept the old setup around as to not break people when upgrading but it's been 2 mayor releases now.

A migration guide is already in place: https://docs.sentry.io/platforms/unity/migration/#changes-to-the-programmatic-configuration.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,7 +63,7 @@ jobs:
         if: runner.os != 'Windows'
         uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4
         with:
-          global-json-file: src/sentry-dotnet/global.json
+          global-json-file: global.json
         
       - name: Install Android dotnet workflow
         run: dotnet workload install android --temp-dir "$RUNNER_TEMP"
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -21,6 +21,14 @@ jobs:
           -o -name "*.m"
           | xargs clang-format -i -style=file
 
+      - name: Install .NET SDK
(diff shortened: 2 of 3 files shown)


samples/unity-of-bugs/Assets/Resources/Sentry/SentryOptions.asset
```diff
@@ -66,9 +66,8 @@ MonoBehaviour:
   <WindowsNativeSupportEnabled>k__BackingField: 1
   <MacosNativeSupportEnabled>k__BackingField: 1
   <LinuxNativeSupportEnabled>k__BackingField: 1
+  <XboxNativeSupportEnabled>k__BackingField: 1
```
bitsandfoxes, line 69: Missed checking this in to the samples when adding Xbox support.

src/Sentry.Unity/ScriptableSentryUnityOptions.cs
```diff
@@ -182,6 +178,7 @@ internal SentryUnityOptions ToSentryUnityOptions(
             AndroidNativeSupportEnabled = AndroidNativeSupportEnabled,
             NdkIntegrationEnabled = NdkIntegrationEnabled,
             NdkScopeSyncEnabled = NdkScopeSyncEnabled,
+            PostGenerateGradleProjectCallbackOrder = PostGenerateGradleProjectCallbackOrder,
```
bitsandfoxes, line 181: This was previously wrongly only on the scriptable object and not on the actual options, making them inaccessible programmatically.

mujacica (approved):

PR #2327, opened 2025-09-22
fix: Breadcrumb issues in Apple native bridge

bitsandfoxes:
Superseds https://github.com/getsentry/sentry-unity/pull/2321

### The Problem

We've received different crash events originating from the SDK when running on Apple platforms.

#### Date Formatting

The first crash we have is 

```
EXC_BAD_ACCESS (SIGSEGV)
  Exception Type: EXC_BAD_ACCESS (SIGSEGV)
  Exception Codes: KERN_INVALID_ADDRESS at 0x0000000000000000

  Application Specific Information:
  Crashed in icu::Calendar::clear()

  Thread 0 Crashed:
  0   libicucore.A.dylib                  0x1981de8c4 icu::Calendar::clear() + 164
  1   CoreFoundation                      0x18e59f1dc CFDateFormatterGetAbsoluteTimeFromString + 632
  2   CoreFoundation                      0x18e59e030 CFDateFormatterCreateDateFromString + 724
  3   Foundation                          0x18d1642a8 getObjectValue + 244
  4   Foundation                          0x18d165438 -[NSDateFormatter getObjectValue:forString:errorDescription:] + 192
  5   Foundation                          0x18d163ff0 -[NSDateFormatter dateFromString:] + 52
  6   Sentry.framework                    0x10cbb42b0 __SentryNativeBridgeAddBreadcrumb_block_invoke + 116
  7   Sentry.framework                    0x100969784 -[SentryHub configureScope:] + 52
  8   Sentry.framework                    0x100995224 +[SentrySDK configureScope:] + 56
  9   Sentry.framework                    0x10cbb41c8 SentryNativeBridgeAddBreadcrumb + 652
  10  Sentry.framework                    0x113bd2448 SentryCocoaBridgeProxy_AddBreadcrumb + 3366
  11  Sentry.framework                    0x113bbe034 NativeScopeObserver_AddBreadcrumbImpl + 1522
  12  Sentry.framework                    0x113af037c ScopeObserver_AddBreadcrumb + 828
  13  Sentry.framework                    0x113aefaac HubExtensions_AddBreadcrumb + 556
  14  Sentry.framework                    0x113bcfc9c UnityApplicationLoggingIntegration_OnLogMessageReceived + 29343
```

This hints at invalid date formatting with `NSDateFormatter.dateFromString` attempting to parse the timestamp using an invalid memory pointer, causing `icu::Calendar::clear()` to dereference `NULL (x0: 0x0)`.

#### Breadcrumb Serialization Issue

[...]

PR #2338, opened 2025-09-26
fix: Added timestamp validation to iOS bridge

bitsandfoxes:
Follow up on https://github.com/getsentry/sentry-unity/pull/2327 based on the following feedback

> In SentryNativeBridgeAddBreadcrumb, the dateFromString: result is directly assigned to breadcrumb.timestamp. If date parsing fails, this can unexpectedly set the timestamp to nil, creating an inconsistency with macOS behavior where the parsed date is validated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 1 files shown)


philipphofmann (approved):
Thanks.

PR #2342, opened 2025-09-29
chore: Pin SDK version

bitsandfoxes:
Pinning the the SDK version so `dotnet` doesn't just take whatever is there.

#skip-changelog

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -21,6 +21,15 @@ jobs:
           -o -name "*.m"
           | xargs clang-format -i -style=file
 
+      - name: Install .NET SDK
+        if: runner.os != 'Windows'
+        uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4
+        with:
+          global-json-file: global.json
+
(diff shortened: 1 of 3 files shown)


global.json
```diff
@@ -1,5 +1,8 @@
 {
   "sdk": {
+    "version": "9.0.304",
```
Flash0ver, line 3: note: matches `getsentry/sentry-dotnet` ✅
https://github.com/getsentry/sentry-dotnet/blob/44b0c8288722d7e03f506ee3cf862c0e10954cd5/global.json

.github/workflows/format-code.yml
```diff
@@ -21,6 +21,15 @@ jobs:
           -o -name "*.m"
           | xargs clang-format -i -style=file
 
+      - name: Install .NET SDK
+        if: runner.os != 'Windows'
```
Flash0ver: suggestion: remove condition; "Format Code" runs only on `ubuntu-latest`

```suggestion
```

Or do we want to intentionally keep this step in sync with the step in `build.yml`, to be an exact duplicate?

Flash0ver (approved):
One little (non-blocking) nitpick.

PR #2368, opened 2025-10-22
feat: Structured Logging

bitsandfoxes:
This PR adds support for `Structured Logging`. 

## Implementation Context

The tricky part is the separation between `UnityLogHandlerIntegration` and `UnityApplicationLoggingIntegration`.

### `UnityLogHandlerIntegration`

- Receives the actual excpetion object we need for line number processing
- Receives all pre-formatted log messages <- we use this to capture structured logs

### `UnityApplicationLoggingIntegration`

- Receives materialized log messages allowing us to capture breadcrumbs without additional allocation
- Receives stringified Unity stack traces we use to attach to events captured via `Debug.LogError`
- Allows us to capture exceptions on `WebGL`

## Result

Screenshot 2025-10-22 at 16 49 31

Screenshot 2025-10-22 at 16 51 56

### DIFF
--- a/test/Sentry.Unity.Tests/UnityLogHandlerIntegrationTests.cs
+++ b/test/Sentry.Unity.Tests/UnityLogHandlerIntegrationTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using NUnit.Framework;
+using Sentry.Internal;
 using Sentry.Protocol;
 using Sentry.Unity.Integrations;
 using Sentry.Unity.Tests.SharedClasses;
@@ -23,7 +24,7 @@ public sealed class UnityLogHandlerIntegrationTests
             var application = new TestApplication();
             var integration = StructuredLogger != null
                 ? new UnityLogHandlerIntegration(application, () => StructuredLogger)
-                : new UnityLogHandlerIntegration(application);
+                : new UnityLogHandlerIntegration(application, () => DisabledSentryStructuredLogger.Instance);
             integration.Register(Hub, SentryOptions);
             return integration;
         }
@@ -45,7 +46,7 @@ public sealed class UnityLogHandlerIntegrationTests
     public void CaptureException_ExceptionCapturedAndMechanismSet()
     {
         var sut = _fixture.GetSut();
-        var message = NUnit.Framework.TestContext.CurrentContext.Test.Name;
+        var message = "test message" + Guid.NewGuid();
 
         sut.ProcessException(new Exception(message), null);
 


src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
bitsandfoxes, line 1: It looks like a lot but I've split up the implementation of everything inside of `OnLogMessageReceived` into:
- `IsGettingDebounced`
- `ProcessException`       <- responsible for capturing exceptions via logs on WebGL
- `ProcessError`               <- responsible for creating events out of Debug.LogError when configured
- `ProcessBreadcrumbs` <- responsible for creating breadcrumbs out of logs

src/Sentry.Unity/Integrations/UnityErrorLogException.cs
```diff
@@ -24,6 +24,7 @@ internal class UnityErrorLogException : Exception
         private readonly IDiagnosticLogger? _logger;
 
         public UnityErrorLogException(string logString, string logStackTrace, SentryOptions? options)
+            : base(logString)
```
bitsandfoxes, line 27: This is fixing an actual bug where the .NET SDK would automatically add a breadcrumb for exceptions but the actual logmessage would be missing, causing a generic "Exception of type X was captured" breadcrumb.

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -49,72 +49,100 @@ internal void OnLogMessageReceived(string message, string stacktrace, LogType lo
...
+        }
+        else
         {
-            _hub.AddBreadcrumb(message: message, category: "unity.logger", level: ToBreadcrumbLevel(logType));
+            _hub?.CaptureMessage(message, level: SentryLevel.Error);
         }
     }
 
-    private void OnQuitting()
+    private void ProcessBreadcrumbs(string message, LogType logType)
```
bitsandfoxes, line 123: This is the actual change in this file: Capture breadcrumbs only when not capturing structured logs or when explicitly opted in.

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -97,10 +155,10 @@ private void OnQuitting()
         // 'OnQuitting' is invoked even when an uncaught exception happens in the ART. To make sure the .NET
         // SDK checks with the native layer on restart if the previous run crashed (through the CrashedLastRun callback)
         // we'll just pause sessions on shutdown. On restart they can be closed with the right timestamp and as 'exited'.
-        if (_sentryOptions?.AutoSessionTracking is true)
+        if (_options.AutoSessionTracking)
         {
             _hub?.PauseSession();
         }
-        _hub?.FlushAsync(_sentryOptions?.ShutdownTimeout ?? TimeSpan.FromSeconds(1)).GetAwaiter().GetResult();
+        _hub?.FlushAsync(_options.ShutdownTimeout).GetAwaiter().GetResult();
```
bitsandfoxes, line 162: There is a separate PR that cleans up the unreleated SessionTracking in the LogHandlerIntegration.

src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
```diff
@@ -49,72 +49,100 @@ internal void OnLogMessageReceived(string message, string stacktrace, LogType lo
...
+        {
+            // Capturing of breadcrumbs for exceptions happens inside the .NET SDK
+            return;
+        }
+
+        // Breadcrumb collection on top of structure log capture must be opted in
+        if (_options.Experimental is { EnableLogs: true, AttachBreadcrumbsToEvents: false })
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
bitsandfoxes: No, this is correct.

src/Sentry.Unity/Integrations/UnityLogHandlerIntegration.cs
```diff
@@ -75,19 +80,72 @@ internal void CaptureException(Exception exception, UnityEngine.Object? context)
...
+        if (_hub?.IsEnabled is not true || !_options.Experimental.EnableLogs)
+        {
+            return;
+        }
+
+        // We're not capturing the SDK's own logs.
+        if (args.Length > 1 && Equals(args[0], UnityLogger.LogTag))
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
bitsandfoxes: This is not correct. Usage of the `UnityLogger` guarantees that there will always be at least two arguments, the tag and the actual message. There are no single-argument logs coming from within the SDK.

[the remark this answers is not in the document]
bitsandfoxes: If it is not a string and it fails then it still does what it is intended to do, filter out SDK logs.

Issue #533, opened 2022-01-28
Crash Free Sessions can be misleading without native platform support


bruno-garcia:
We've documented here: https://docs.sentry.io/platforms/unity/configuration/releases/

> The Crash Free Session Rate requires native support. On platforms that don't yet have native support, the SDK reports sessions that did not end gracefully as Abnormal, which leads to 100% crash free sessions being recorded. If you're building on a platform that doesn't have native support yet, you should monitor your app's Abnormal rate.

This could potentially be improved in the product to surface a distinction when "crash" isn't possible. Either by focusing on 'abnormal' or another strategy. Additionally we could document or automate a way to have a distinction in the "release name" for builds that support "crash free rate" vs those that only report healthy/abnormal.

Related to:
* https://github.com/getsentry/sentry-dart/issues/542
* https://github.com/getsentry/sentry-javascript/issues/4466

bitsandfoxes:
Requires documentation which explicit platform we support for the Crash Free Sessions feature.

bitsandfoxes:
1. The SDK allows for `SessionEndStatus.Unhandled`, see https://github.com/getsentry/sentry-unity/pull/2376. This allows us to be more granular with the information provided
2. Any additional features provided should go through the improved support outlined in https://github.com/getsentry/sentry-unity/issues/668
3. Added this to the known limitations on the [docs](https://github.com/getsentry/sentry-docs/pull/15485)

Issue #668, opened 2022-03-30
WebGL JavaScript & native error handling


vaind:
After the dotnet WebGL support is merged in #657, we can add support for javascript errors. There are two main approaches, the selection of the best one is part of this issue:
1. either include `sentry-javascript` and let it do the reporting
2. or implement the error handlers same as sentry-javascript does but forward errors to dotnet via wasm calls (if that's even possible - there's a list of all native functions available in development build sources)

In both cases, we would need to add an editor script that updates the generated JS sources (that supports both development & release builds, the latter including compression). 

There's already a button to trigger a JS error - look for ThrowJavaScript(), enable it in NativeSupportScene.cs - but there would likely need to be more, e.g. to also include it in the SmokeTest.

As a related topic, we should check whether the selected approach also covers errors in the native code that are unhandled - you can reproduce those within the "Native Samples" scene in the sample repo.

vaind:
Additional off-topic TODO:  warn/error out during build when `"com.unity.modules.unitywebrequest": "1.0.0"` package is disabled - otherwise Unity just throws a vague error: `Cannot create web request without initializing the system` that I couldn't get any useful resolution for by googling 

It requires the build script to be present so sounds like a good time to do it in a single PR.

bruno-garcia:
> either include sentry-javascript and let it do the reporting

I think this is the preferred solution as it's our largest SDK and has a lot of development going on there.

vaind:
I've tried to integrate the browser `bundle.js` and it kinda works, as in the event is reported to sentry.io, however, I'm not sure how to get any stack trace info, because the scripts are loaded as blobs, which are, in case of a release build, also compressed...

Maybe we'd need to create a custom integration that does the stack parsing.

[unity-web.zip](https://github.com/getsentry/sentry-unity/files/8489074/unity-web.zip)

vaind:
OK, so the stack traces for the embedded code coming from Unity `.jslib` are pretty limited but I don't think there's much we can do, they're just not there on the error reported by the browser. 

The integration could still be useful at least to get the error type/text + stack traces if they're coming from standard javascript files.

What we would need to do is:
1. Add sentry-js bundle to the sentry-unity package (either in repo or in CI) - I guess we're fine with including only the minified version, regardless of what kind of app user builds: https://browser.sentry-cdn.com/6.19.6/bundle.min.js
[...]

PR #2404, opened 2025-11-10
chore: Disable auto session tracking on WebGL

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/533

The SDK has no mechanism to determine whether the last run was crashed or not, painting a very incomplete picture regarding the crash free rate.

This should be picked up when tackling https://github.com/getsentry/sentry-unity/issues/668


bitsandfoxes:
Thinking about it again, the SDK now supports `SessionEndStatus.Unhandled`, see https://github.com/getsentry/sentry-unity/pull/2376. This allows us to be more granular with the information provided.

PR #2410, opened 2025-11-11
chore: Reverted sample back to `2021.3.45f2`

bitsandfoxes:
I bumped the sample version in https://github.com/getsentry/sentry-unity/pull/2392 without realizing, that this is for enterprise only.

#skip-changelog


bitsandfoxes:
Skipping CI. Affects `main` only.

mujacica (approved):

PR #2422, opened 2025-11-14
chore: CI Improvements

bitsandfoxes:
These improvements include but are not limited to
- Fixing license parallelization
- Fixing step naming
- Added appropriate caching all over the place to speed up Unity builds
- Added missing skip checks in places where caching works its magic

#skip-changelog

### DIFF
--- a/test/Scripts.Integration.Test/Editor/Builder.cs
+++ b/test/Scripts.Integration.Test/Editor/Builder.cs
@@ -154,6 +154,18 @@ public class Builder
     public static void BuildAndroidIl2CPPPlayer()
     {
         Debug.Log("Builder: Building Android IL2CPP Player");
+
+#if UNITY_2021_2_OR_NEWER && !UNITY_6000_0_OR_NEWER
+        // Clean Android gradle cache to force regeneration of gradle files
+        // This prevents Unity from reusing gradle files that may contain Sentry symbol upload tasks from previous builds
+        var androidGradlePath = Path.Combine(Directory.GetCurrentDirectory(), "Library/Bee/Android");
+        if (Directory.Exists(androidGradlePath))
+        {
+            Debug.Log($"Builder: Cleaning Android gradle cache at '{androidGradlePath}'");
+            Directory.Delete(androidGradlePath, true);
+        }
+#endif
+
         BuildIl2CPPPlayer(BuildTarget.Android, BuildTargetGroup.Android, BuildOptions.StrictMode);
     }
     public static void BuildAndroidIl2CPPProject()


scripts/ci-docker.sh
```diff
@@ -50,9 +50,7 @@ echo $licenseConfig | $suexec -i $container sh -c "cat > /usr/share/unity3d/conf
 $suexec $container chown -R $uid /usr/share/unity3d/config/
 
 # Unity 2021+ tries to write to this directory during asset import...
-if [[ $unityPrefix -ge 2021 ]]; then
-    $suexec $container chmod -R 755 /opt/unity/Editor/Data/UnityReferenceAssemblies/
-fi
+$suexec $container chmod -R 755 /opt/unity/Editor/Data/UnityReferenceAssemblies/
```
bitsandfoxes, line 63: This was failing with
```
./scripts/ci-docker.sh: line 53: [[: 6000.0: syntax error: invalid arithmetic operator (error token is ".0")
```
and we're running on `2021+` only anyway.

.github/workflows/build.yml
```diff
@@ -134,9 +139,7 @@ jobs:
             package-release.zip
 
       - name: Run Unity tests (playmode)
-        run: |
-          docker exec unity dotnet msbuild /t:UnityConfigureSentryOptions /p:TestDsn= /p:Configuration=Release /p:OutDir=other src/Sentry.Unity
-          docker exec unity dotnet msbuild /t:UnityPlayModeTest /p:Configuration=Release /p:OutDir=other test/Sentry.Unity.Tests
+        run: docker exec unity dotnet msbuild /t:UnityPlayModeTest /p:Configuration=Release /p:OutDir=other test/Sentry.Unity.Tests
```
bitsandfoxes, line 138: Running it twice was a leftover from a time where we had to manually set/replace the `TestDsn`. That's no longer needed.

.github/workflows/sdk.yml
```diff
@@ -81,11 +81,13 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
 
       - name: Install .NET SDK
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
```
bitsandfoxes, line 84: Added missing skips if the cache was hit.

.github/workflows/build.yml
```diff
@@ -38,10 +48,6 @@ jobs:
         env:
           UNITY_SCRIPT_ARG: unity${{ env.UNITY_VERSION }}
 
-      - run: echo "::add-mask::$LICENSE_SERVER_URL"
```
bitsandfoxes, line 41: The masking does nothing
```
Can't add secret mask for empty string in ##[add-mask] command.
```

.github/workflows/build.yml
```diff
@@ -72,11 +78,6 @@ jobs:
       - name: Restore .NET Workload
         run: dotnet workload restore
 
-      - name: Install Android dotnet workflow
```
bitsandfoxes, line 75: `dotnet workload restore` already does that for us

.github/workflows/build.yml
```diff
@@ -38,17 +48,13 @@ jobs:
         env:
           UNITY_SCRIPT_ARG: unity${{ env.UNITY_VERSION }}
 
-      - run: echo "::add-mask::$LICENSE_SERVER_URL"
-        env:
-          LICENSE_SERVER_URL: ${{ secrets.LICENSE_SERVER_URL }}
-
-      - name: Restore Unity Packages
```
bitsandfoxes, line 45: I don't even remember why we were trying to cache the packages only. But it doesn't work in any case:
```
Cache not found for input keys
```
We're just caching the full `Library` like in the rest of CI instead. This lets us skip the full on creation of the assetdatabase with every single run.

PR #2479, opened 2026-01-15
feat: Add `IThrottler` to allow message & stack trace based event throttling

### LINKED ISSUE #2004: Improve event debouncing and throtteling
## Goal 
The goal is to provide more granular control over event debouncing. Particularly, for high-frequency events like those from `Update()` functions. This allows users to preserve quota while keeping error reporting intact.

## Current Problem
1. Events from `Update()` functions that run every frame can quickly consume error reporting quotas
2. Current debouncing is not flexible enough for different types of events. 
i.e. There's not debouncing for `exceptions` 
3. Existing workarounds in `BeforeSend` (stack trace checking + random sampling) are hacky and might miss important errors

### Current Implementation

We're using the following debouncer https://github.com/getsentry/sentry-unity/blob/main/src/Sentry.Unity/TimeDebounceBase.cs

in the `Application Logging Integration`
https://github.com/getsentry/sentry-unity/blob/596ad1ddcd260f44fff90c82cd737e2b32bdabc0/src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs#L57-L71 

## Proposal

Replace the current solution with a content-based deduplication system using an interface pattern similar to callbacks in the SDK (like `BeforeSend`, `BeforeBreadcrumb`, `TracesSampler` etc.).

The idea is to add an `IDebounce` to the `SentryUnityOptions` with a default implementation that can be overwritten by the user.
The default Implementation would take the log message and a couple of stack frames, create a hash and hold a fixed size buffer of those for a set amount of time. The individual variables would not be configurable, users can implement their own solution.
The added benefit is that we would run this before IL2CPP stack trace processing, further minimizing the overhead.

### Additional Consideration

The .NET SDK has its own [debouncer](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Debouncer.cs) now (for memory dumps). Can this be reused?

The Godot SDK has event throtteling implemented as part of its [logging intgration](https://github.com/getsentry/sentry-godot/pull/28).

bitsandfoxes:
Closes #2004

## Problem Statement

The existing time-based log debouncing system throttles events purely based on time intervals per log type. This approach has limitations:
- Identical errors occurring in rapid succession consume quota unnecessarily
- Different errors of the same log type get incorrectly throttled together
- No distinction between repeated errors and unique errors happening close together

### Deprecate Debouncing
Replace (deprecate) purely log type & time-based debouncing with content-based event throttling that deduplicates based on the actual error content (message + stacktrace fingerprint). 

### Add `IThrottler`

Adding a property to the options for users to provide their own implementation. Since the purpose of the throttler is to provide a way to minimize noise and data duplication, the interface allows to check for events, exceptions, breadcrumbs and structured logs.
The default implementation provided by the SDK only implements throttling for events and exceptions and does not interfere with breadcrumbs and logs.

```
public interface IThrottler
{
    bool ShouldCaptureEvent(string message, string stackTrace, LogType logType);
    bool ShouldCaptureBreadcrumb(string message, LogType logType);
    bool ShouldCaptureStructuredLog(string message, LogType logType);
    bool ShouldCaptureException(Exception exception);
}
```

This ensures:
- Repeated identical errors are deduplicated within a time window
- Unique errors are always captured, regardless of timing
- Quota is preserved without losing distinct error information

The default implementation using `message + stacktrace` fingerprinting with a configurable dedupe window (default: 1000ms) configurable through the config window.

### Config Window

For simplicity, the option is opt-in through the config window and will create a new instance of the `ErrorEventThrottler` if enabled. Custom implementations need to be provided via the programmatic options.

Screenshot 2026-01-29 at 15 30 00

### DIFF
--- a/src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
+++ b/src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs
@@ -88,7 +88,7 @@ internal class UnityApplicationLoggingIntegration : ISdkIntegration
 
     private void ProcessError(string message, string stacktrace, LogType logType)
     {
-        if (logType is not LogType.Error || !_options.CaptureLogErrorEvents)
+        if (logType is not (LogType.Error or LogType.Assert) || !_options.CaptureLogErrorEvents)
         {
             return;
         }


TheCrossBoy:
been following along on this for awhile and am very excited to see it so close!

minor thing: does `ILogThrottler` throttle _log messages_ in addition to events / errors / messages? current name suggests it might but i believe it's just for the latter, which may be confusing

bitsandfoxes:
> been following along on this for awhile and am very excited to see it so close!
> 
> minor thing: does `ILogThrottler` throttle _log messages_ in addition to events / errors / messages? current name suggests it might but i believe it's just for the latter, which may be confusing

Excellent point. I renamed it to `IErrorEventThrottler`.

test/Sentry.Unity.Tests/ContentBasedThrottlerTests.cs
```diff
@@ -0,0 +1,247 @@
...
+        Assert.IsTrue(result2);
+    }
+
+    [Test]
+    public void ShouldCapture_LogTypeLog_AlwaysReturnsTrue()
+    {
+        var throttler = new ContentBasedThrottler(TimeSpan.FromSeconds(10));
+        var message = "test message";
+
+        var result1 = throttler.ShouldCapture(message, "stacktrace", LogType.Log);
```
mujacica: Is there a need to also handle logs? Or make it configurable, for cases where a repeated log in game thread could be spamming a lot?

bitsandfoxes: I updated the interface to allow implementing throttling for breadcrumbs and logs, but the default implementation provided by the SDK only affects errors & exceptions.

PR #2502, opened 2026-01-23
fix: Active scene change no longer refreshes trace ID

bitsandfoxes:
Followup on https://github.com/getsentry/sentry-unity/pull/2496 and https://github.com/getsentry/sentry-unity/pull/2374

The SDK refrains from creating a new propagation context due to lifecycle events, see https://github.com/getsentry/sentry-unity/pull/2374

But it still does so on active scene change, leading to fragmentation of logs over the lifespan of the game.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Behavioral Changes
+
+- The SDK no longer refreshes the trace ID when changing active scene. This follows the changes to not do so based on lifecycle events, meaning that the trace ID persists from game start to game end, as intended in `4.0.0`. ([#2502](https://github.com/getsentry/sentry-unity/pull/2502))
+
 ### Fixes
 
 - The SDK now specifies the files and directories targeted for debug symbol upload instead of pointing sentry-cli at the build output directory ([#2485](https://github.com/getsentry/sentry-unity/pull/2485))


src/Sentry.Unity/Integrations/LifeCycleIntegration.cs
```diff
@@ -30,6 +30,7 @@ public void Register(IHub hub, SentryOptions sentryOptions)
 
         if (!_options.AutoSessionTracking)
         {
+            sentryOptions.LogDebug("AutoSessionTracking is disabled. Skipping {0}.", nameof(LifeCycleIntegration));
```
bitsandfoxes, line 33: Missing log when `Register` actually skips registering.

src/Sentry.Unity/Integrations/SceneManagerTracingIntegration.cs
bitsandfoxes, line 1: Just using the same pattern as everywhere else.

src/Sentry.Unity/Integrations/StartupTracingIntegration.cs
bitsandfoxes, line 1: Missing log when Register actually skips registering. Following the pattern everywhere else in the SDK.

src/Sentry.Unity/Integrations/TraceGenerationIntegration.cs
```diff
@@ -20,30 +21,18 @@ internal TraceGenerationIntegration(ISentryMonoBehaviour sentryMonoBehaviour, IS
...
-
-        // Set up scene change handling if tracing is disabled or auto scene load traces are disabled
-        if (!isTracingEnabled || !unityOptions.AutoSceneLoadTraces)
-        {
-            _sceneManager.ActiveSceneChanged += (_, _) =>
-            {
-                options.DiagnosticLogger?.LogDebug("Active Scene changed. Creating new Trace.");
-                hub.ConfigureScope(scope => scope.SetPropagationContext(new SentryPropagationContext()));
-            };
-        }
```
bitsandfoxes, line 47: This is the actually relevant change.

Issue #2052, opened 2025-03-04
Support for Nintendo Switch


bruno-garcia:
Sentry has support for Nintendo Switch through an official integration with Nintendo where they send the native dumps directly to Sentry.

Today, those crashes are agnostic of game engine. So if you're using the Sentry SDK for Unity, you tags (set via C#) will be in crashes on Mobile, Web and Desktop. But not on those Switch crashes.

In addition to that, Release Health (crash free rates) are not properly accounted for Switch since the Sentry SDK for Unity is not aware if the game crashes natively or not.

This ticket tracks demand for adding native support for Switch.

Relates to:
* https://github.com/getsentry/sentry-unity/issues/2051
* https://github.com/getsentry/sentry-unity/issues/2050

bruno-garcia:
Internal ticket with NDA protected info (sorry): https://github.com/getsentry/team-gdx/issues/20

PR #2503, opened 2026-01-26
feat: Nintendo Switch native support

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/2052
Depends on https://github.com/getsentry/sentry-switch/pull/79

### Summary

Adds native scope sync support for Nintendo Switch. While crash reporting is already happening via the underlying sentry-switch native SDK, this PR enables the Unity SDK to synchronize scope data (tags, breadcrumbs, user info, context) to the native layer, ensuring native crashes have the same rich context as managed events.

#### New Assembly Output
Similar to Sentry.Unity.Native.PlayStation, this introduces a separate Sentry.Unity.Native.Switch.dll assembly that compiles against the SENTRY_NATIVE_SWITCH define. This allows platform-specific code paths while sharing the common native bridge infrastructure.

#### User-Provided Native Library Model
Unlike other platforms where we ship prebuilt binaries, Switch requires users to compile and provide their own sentry-switch static library due to platform restrictions. The SDK expects:
- Assets/Plugins/Sentry/Switch/libsentry.a
- Assets/Plugins/Sentry/Switch/libzstd.a

#### Build Pre-Process for Native Plugin Management
A new ´NativePluginBuildPreProcess´ detects whether users have provided the native libraries:
- All files present: Disables no-op stubs, links real library
- No files present: Enables stubs (managed SDK works, native features are no-op)
- Partial files: Warns about misconfiguration

#### Platform-Specific Adaptations
- File Access: Switch has limited file write access - DisableFileWrite is enabled, and session tracking is disabled
- Cache Directory: Uses mounted temporary storage instead of Application.persistentDataPath (which would crash)
- Background Worker: Uses WebBackgroundWorker for HTTP transport
- IL2CPP: Line number support disabled (not available on Switch)
- Default User ID: Retrieved from Switch system APIs

#### Debug Symbol Upload
`BuildPostProcess´ now supports Switch, uploading IL2CPP symbols, Burst debug info, and user-provided Sentry plugin symbols via sentry-cli.

### DIFF
--- a/test/Scripts.Tests/package-release.zip.snapshot
+++ b/test/Scripts.Tests/package-release.zip.snapshot
@@ -268,6 +268,10 @@ Runtime/Sentry.Unity.Native.PlayStation.dll
 Runtime/Sentry.Unity.Native.PlayStation.dll.meta
 Runtime/Sentry.Unity.Native.PlayStation.pdb
 Runtime/Sentry.Unity.Native.PlayStation.pdb.meta
+Runtime/Sentry.Unity.Native.Switch.dll
+Runtime/Sentry.Unity.Native.Switch.dll.meta
+Runtime/Sentry.Unity.Native.Switch.pdb
+Runtime/Sentry.Unity.Native.Switch.pdb.meta
 Runtime/Sentry.Unity.pdb
 Runtime/Sentry.Unity.pdb.meta
 Runtime/Sentry.xml


mujacica (approved):
Beautiful! Do we need switch release as well before we release this?

bitsandfoxes:
> Beautiful! Do we need switch release as well before we release this?

Yes. And I'm adding integration tests to the switch repo, see https://github.com/getsentry/sentry-switch/pull/92

PR #2516, opened 2026-02-03
fix: Gracefully handle missing native SDK on consoles

bitsandfoxes:
Follow-up on https://github.com/getsentry/sentry-unity/pull/2503

If there is no console specific lib found, the SDK needs to gracefully degrade as to not break builds.
This is achieved via a unified `sentry_native_stubs.c` that is getting managed at compile time via the `NativePluginBuildPreProcess.cs`. This should (tm) work even when installing the package via the package manager due to `isOverridable: 1` in the stubs' .meta file.


bitsandfoxes:
Well that's a bunch of nonsense.

PR #2547, opened 2026-02-13
chore: Use `Sentry.Analyzer`

bitsandfoxes:
We have it. We added it. We're just not using it.

#skip-changelog

### DIFF
--- a/.github/workflows/create-unity-matrix.yml
+++ b/.github/workflows/create-unity-matrix.yml
@@ -27,12 +27,16 @@ jobs:
       - name: Set Unity version matrices based on event type
         id: set-matrix
         shell: bash
+        env:
+          EVENT_NAME: ${{ inputs.event-name }}
+          PR_UNITY_VERSIONS_VALUE: ${{ env.PR_UNITY_VERSIONS }}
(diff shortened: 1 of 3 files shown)


Directory.Build.props
```diff
@@ -40,6 +40,7 @@
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
     <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="3.3.0" PrivateAssets="All" />
     <PackageReference Include="Nullable" Version="1.3.0" PrivateAssets="All" />
+    <ProjectReference Include="$(RepoRoot)src/sentry-dotnet/src/Sentry.Analyzers/Sentry.Analyzers.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" PrivateAssets="all" />
```
Flash0ver, line 46: note: I believe it's common to group `PackageReference` and `ProjectReference` in separate `ItemGroup`s ... but this is nitpicking of the highest order

Flash0ver (approved):

PR #2548, opened 2026-02-13
chore: Utilize app-runner for Android

bitsandfoxes:
## Goal

On Android, this moves the the `Smoke Tests` over to `Integration Tests`, utilizing `app-runner` to have actual end to end coverage.

## Implementation

We only want to move the tests for Android specifically. To make use of the already existing infrastructure I introduced a `TestLauncher` that depending on the test case provided instantiates the correct `MonoBehaviour`. This has the side effect that the `IntegrationTester` also already compiles for all platforms, giving a positive signal, even tho it is not being used right now.

#skip-changelog

### DIFF
--- a/test/Scripts.Integration.Test/Editor/Builder.cs
+++ b/test/Scripts.Integration.Test/Editor/Builder.cs
@@ -37,7 +37,7 @@ public class Builder
 #endif
 
         Debug.Log("Builder: Configuring code stripping level");
-#if UNITY_6000_0_OR_NEWER
+#if UNITY_2022_1_OR_NEWER
         PlayerSettings.SetManagedStrippingLevel(NamedBuildTarget.FromBuildTargetGroup(group), ManagedStrippingLevel.High);
 #else
         PlayerSettings.SetManagedStrippingLevel(NamedBuildTarget.FromBuildTargetGroup(group), ManagedStrippingLevel.Low);


test/Scripts.Integration.Test/Scripts/TestLauncher.cs
```diff
@@ -0,0 +1,80 @@
...
+#else
+        var args = Environment.GetCommandLineArgs();
+        if (args.Length > 2 && args[1] == "--test")
+        {
+            arg = args[2];
+        }
+#endif
+        return arg;
+    }
+#endif
```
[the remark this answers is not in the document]
bitsandfoxes: True. But we want to straight up delete the SmokeTester after we're done migrating. So this is fine.

.github/workflows/ci.yml
```diff
@@ -270,7 +271,8 @@ jobs:
       fail-fast: false
       matrix:
         unity-version: ${{ fromJSON(needs.create-unity-matrix.outputs.unity-matrix).unity-version }}
-        api-level: [30, 31, 34]
+        # api-level: [30, 31, 34]
+        api-level: [30]
```
bitsandfoxes, line 275: I'll check and update the API levels in a followup.

src/sentry-dotnet
bitsandfoxes, line 1: We need a tiny bump on the .NET SDK here that hardcodes `isAOT = false` to fix some false positives on older Unity versions.

test/Scripts.Integration.Test/Editor/Builder.cs
```diff
@@ -29,16 +29,15 @@ public static void BuildIl2CPPPlayer(BuildTarget target, BuildTargetGroup group,
...
 
-        // This should make IL2CCPP builds faster, see https://forum.unity.com/threads/il2cpp-build-time-improvements-seeking-feedback.1064135/
-        Debug.Log("Builder: Setting IL2CPP generation to OptimizeSize");
+        Debug.Log("Builder: Setting IL2CPP generation to OptimizeSpeed");
 #if UNITY_2022_1_OR_NEWER
-        PlayerSettings.SetIl2CppCodeGeneration(NamedBuildTarget.FromBuildTargetGroup(group), UnityEditor.Build.Il2CppCodeGeneration.OptimizeSize);
+        PlayerSettings.SetIl2CppCodeGeneration(NamedBuildTarget.FromBuildTargetGroup(group), UnityEditor.Build.Il2CppCodeGeneration.OptimizeSpeed);
 #elif UNITY_2021_2_OR_NEWER
-        EditorUserBuildSettings.il2CppCodeGeneration = UnityEditor.Build.Il2CppCodeGeneration.OptimizeSize;
+        EditorUserBuildSettings.il2CppCodeGeneration = UnityEditor.Build.Il2CppCodeGeneration.OptimizeSpeed;
```
bitsandfoxes, line 36: A more likely setting for actual release builds.

test/Scripts.Integration.Test/Editor/Builder.cs
```diff
@@ -29,16 +29,15 @@ public static void BuildIl2CPPPlayer(BuildTarget target, BuildTargetGroup group,
...
 #endif
 
         Debug.Log("Builder: Configuring code stripping level");
-#if UNITY_2022_1_OR_NEWER
+#if UNITY_6000_0_OR_NEWER
```
bitsandfoxes:
```suggestion
#if UNITY_2022_1_OR_NEWER
```

test/Scripts.Integration.Test/Scripts/SmokeTester.cs
```diff
@@ -126,9 +126,6 @@ private IEnumerator SmokeTestCoroutine()
 
         t.ExpectMessage(currentMessage, "'type':'transaction");
         t.ExpectMessage(currentMessage, "'op':'app.start'"); // startup transaction
-#if !UNITY_EDITOR
-        t.ExpectMessage(currentMessage, "'op':'awake','description':'Main Camera.SmokeTester'"); // auto instrumentation
-#endif
```
bitsandfoxes, line 131: The current change from having a `TestLauncher` instead of having the MonoBehaviour already available  during scene load makes this unfeasible. We can add this back in after the migration.

src/Sentry.Unity.Editor/Android/AndroidManifestConfiguration.cs
bitsandfoxes, line 1: We missed setting the `Dist` until now.

src/Sentry.Unity.Editor/ConfigurationWindow/SentryEditorWindowInstrumentation.cs
```diff
@@ -50,6 +50,12 @@ private static void ConfigureOptions(Dictionary<string, string> args, [CallerMem
             OptionsConfigurationItem.SetScript(value);
         }
 
+        if (args.TryGetValue("dsn", out value))
+        {
+            Debug.LogFormat("{0}: Setting DSN", functionName);
+            options.Dsn = value;
+        }
+
```
bitsandfoxes, line 58: We're overwriting the DSN now from CI to protect the project and for later test validation via Pester.

test/IntegrationTest/CommonTestCases.ps1
bitsandfoxes, line 1: Extracted since these apply to all test cases, capture-event, capture-crash e.t.c.

test/Scripts.Integration.Test/Scripts/IntegrationTester.cs
```diff
@@ -0,0 +1,148 @@
...
+        // Just wait a bit for the queued crash report to be sent, then quit.
+        StartCoroutine(WaitAndQuit());
+    }
+
+    private IEnumerator WaitAndQuit()
+    {
+        // Wait for the crash report to be sent
+        yield return new WaitForSeconds(10);
+
+        SentrySdk.FlushAsync(TimeSpan.FromSeconds(5)).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
bitsandfoxes: It's in an integration test environment. It's fine.

test/Scripts.Integration.Test/Scripts/TestLauncher.cs
bitsandfoxes, line 1: This is a temporary solution so we can use the old `SmokeTester` and the new `IntegrationTester` in parallel.

PR #2572, opened 2026-03-04
chore: Utilize app-runner for WebGL

bitsandfoxes:
#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -310,7 +310,6 @@ jobs:
     secrets: inherit
     strategy:
       fail-fast: false
-      max-parallel: 2
       matrix:
         unity-version: ${{ fromJSON(needs.create-unity-matrix.outputs.unity-matrix).unity-version }}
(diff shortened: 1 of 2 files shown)


test/IntegrationTest/Integration.Tests.WebGL.ps1
```diff
@@ -0,0 +1,181 @@
...
+
+# Import app-runner modules
+. $PSScriptRoot/../../modules/app-runner/import-modules.ps1
+
+# Import shared test cases and utility functions
+. $PSScriptRoot/CommonTestCases.ps1
+
+BeforeAll {
+    # Run integration test action via WebGL (HTTP server + headless Chrome)
+    function Invoke-TestAction {
```
tustanivsky, line 21: I wonder if this logic could be moved to a separate WebGL device provider in the `app-runner` module. This would allow launching the web app with a single `Invoke-DeviceApp` call and could also be useful for the Godot SDK which currently lacks web integration tests.

PR #2595, opened 2026-03-19
chore: inline unity version matrix

bitsandfoxes:
Put the unity versions directly in `ci.yml` instead of some other workflow. Inline it. Also makes for a cleaner summary with the dependency lines.

#skip-changelog


bitsandfoxes:
Let's leave it as is. The current approach seems to be the most lightweight.

PR #2642, opened 2026-04-13
fix: drop screenshots for events filtered by `before_send`

### LINKED ISSUE #2641: Unity SDK sends screenshot when the event is dropped via BeforeSend
### SDK Version

4.1.0+111a23511e96e576f1ffc7583b6cabd3b46f7b35

### Unity Version

NA

### Platform

- [ ] Editor
- [ ] Mobile (Android, iOS)
- [ ] Desktop (Windows, macOS, Linux)
- [ ] WebGL
- [ ] Console

### Sentry Environment

Sentry SaaS (sentry.io)

### Other Error Monitoring Solution

No

### Steps to Reproduce

1. Enable screenshots in the SDK: https://docs.sentry.io/platforms/unity/enriching-events/screenshots/
2. Filter the error event in BeforeSend: https://docs.sentry.io/platforms/unity/configuration/filtering/#using-before-send

### Expected vs Actual Result

Expected result: as screenshots are only sent as an attachment to errors, if the error is dropped by the SDK for any reason, the screenshot should also be dropped. 

Actual result: the SDK sends the screenshot, which is accepted by the server but not visible and consumes available attachment quota for the organisation. 

Additionally, the server can handle this in the server side: if the event for an attachment is not stored, the attachment is discarded.

JoshuaMoelans:
Fixes #2641 

Since screenshots await `EndOfFrame()` , we start a coroutine that ends up sending an envelope with just the screenshot item. However, in the meantime, the related event might have been discarded by the `before_send` filter. In this case, we don't want to send unnecessary data over to Relay.

As a temporary workaround, setting `SetBeforeCaptureScreenshot` to the same filter used in `SetBeforeSend` prevents these screenshots from being sent.

## Proposed Fix

- Add internal `bool WasCaptured` to `SentryEvent` ([sentry-dotnet/SentryEvent.cs:249](https://github.com/getsentry/sentry-dotnet/blob/6.3.1/src/Sentry/SentryEvent.cs#L13)) — a flag that tracks whether an event actually made it through the entire send pipeline. We gate it behind a `#if SENTRY_UNITY` to avoid unnecessary fields for all of the other .NET SDKs. Defaults to `false`, which is the safe default: if the flag is never set, the screenshot is skipped rather than sent as an orphan.

- Set it to `true` on the success path in `DoSendEvent` ([sentry-dotnet/SentryClient.cs:416](https://github.com/getsentry/sentry-dotnet/blob/6.3.1/src/Sentry/SentryClient.cs#L414)) — placed immediately after `CaptureEnvelope` returns true, which is the only point where we know the event was actually sent. The flag is set on the original @event parameter (not processedEvent) because event processors and `before_send` can swap the event object, but the screenshot coroutine's closure holds a reference to the original. All five drop paths (exception filters, event processors, before_send, sampling, transport failure) return early and never reach this line, leaving the flag false.

- Check the flag in the screenshot coroutine before capturing ([ScreenshotEventProcessor.cs:48](https://github.com/getsentry/sentry-unity/blob/4.2.0/src/Sentry.Unity/ScreenshotEventProcessor.cs#L45)) — after `WaitForEndOfFrame` resumes the coroutine (at which point `DoSendEvent` has already completed synchronously), we read `WasCaptured`. If false, we yield break and skip the screenshot entirely. This check runs before any screenshot capture or CaptureAttachment call, preventing orphaned attachment envelopes from ever being created.

### Future Ideas
- Would it make sense to have our own `ISentryClient` in Unity which overwrites some methods (either deriving or implement custom one that wraps)
  - If user binds their own client, wrapping makes more sense than doing inheritance
  - The wrapper could then "attach" additional data to `SentryEvent`s via a Unity-SDK-owned [ConditionalWeakTable<,>](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2), that the Unity-ScreenshotEventProcessor reads, to have a fully Unity-owned implementation without requiring changes+release on `sentry-dotnet`

---

Related PRs:
- https://github.com/getsentry/sentry-unity/pull/2240
- https://github.com/getsentry/sentry-dotnet/pull/4357

### DIFF
--- a/test/Sentry.Unity.Tests/ScreenshotEventProcessorTests.cs
+++ b/test/Sentry.Unity.Tests/ScreenshotEventProcessorTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.IO;
+using System.Text.RegularExpressions;
 using NUnit.Framework;
 using Sentry.Unity.Tests.Stubs;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


JoshuaMoelans:
awaiting release of sentry-dotnet with the first part of the fix from https://github.com/getsentry/sentry-dotnet/pull/5162/#issue-4315842194

bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-unity/pull/2661

PR #2664, opened 2026-05-06
chore: retrieve events by ID vs tag

bitsandfoxes:
The flake I keep seeing seems to originate from CI failing to retrieve events based on a tag within the timeout. So instead, we try to retrieve it via ID instead and see if that does anything. 

#skip-changelog


bitsandfoxes:
no worksi

PR #2665, opened 2026-05-06
chore: bump actions in ci

bitsandfoxes:
This addresses the 79 warnings that 
```
Node.js 20 actions are deprecated.
```
which will hit us in June. 

#skip-changelog


.github/actions/wait-for-artifact/action.yml
bitsandfoxes, line 1: We previously relied on this fork https://github.com/vaind/download-artifact (that is in dire need of getting synced), or, like here, we wrap `action/download-artifact` in 10 lines of bash and call it a day.

JoshuaMoelans (approved):

Issue #1961, opened 2025-01-11
ANR Detection Improvements


Angelodaniel:
As of today Unity has their own watchdog thread that is checking if ANR's are happening. 
We've disabled leaning on the Android SDK for the ANR detection but the Android SDK has progressed over time to have more types of ANR detection with their own pros/cons.

So there is another ticket that will be responsible for tagging the Mechanism to a Sentry event which is already great. 
But we also need to have more precise ANR detection on this, hence the ticket. 

@bruno-garcia, @romtsn & @bitsandfoxes have been in conversation about this.

bruno-garcia:
> But we also need to have more precise ANR detection on this, hence the ticket.

What would that be?

romtsn:
nope we do not capture AppExits as events still

Angelodaniel:
Any luck? 🥁 @romtsn

romtsn:
@Angelodaniel this requires a bigger project, but I don't see it being a high priority item at the moment, so no update here. Maybe next quarter we'll be able to tackle it (or the one after).

bitsandfoxes:
Leaving the Google docs for [Debug ANRs](https://developer.android.com/games/engines/unity/unity-anrs) here for future reference.
Linking to the team issue tracking ANR across gaming SDKs: https://github.com/getsentry/team-gdx/issues/113

bitsandfoxes:
In the meantime the Unity SDK could provide options to switch between C# and "native" ANR reporting. See https://github.com/getsentry/sentry-unity/issues/2446

imkoi:
Hello, previously my issue was closed as duplicate, so I will write here once again.

Recently we face problems with ANRs increase rate on android platform.

During the investigation we enabled ANR detect mechanism in sentry unity SDK and saw that we don't have any stack trace with information regarding ANR.

As we saw in documentation this behavior is expected since we get the stack trace from background thread. Instead, we would request to do mechanism that will work from native part of the sentry and make stack trace dump from all threads including the game thread (Unity Main Thread).

As result we would see ANRs on android same way as we see iOS App Hang events - with attached stack traces from main game thread

@Angelodaniel

mujacica:
Hey @imkoi! Thanks for the detailed explanation. Let us take a look internally and we will get back to you!

cobyeastwood183:
[...]

PR #2671, opened 2026-05-08
feat: added `AndroidNativeAnrEnabled` and related options

bitsandfoxes:
## Problem

Unity games can get stuck in multiple places and as of right now the SDK surfaces only the C# watchdog integration to capture ANRs. This comes with several limitations. `sentry-java` has its own ANR integrations that the SDK currently  hardcoded opts-out.

## Context

Unity games have their own main thread. The `sentry-java` SDK has a v1 and a v2 ANR integration that monitors the Android UI thread.

There are these native options:
- `enableAnr`
- `enableScopePersistence`
- `isReportHistoricalAnrs`
- `isAttachAnrThreadDump`

Since the work next to each other we keep the umbrella flags `AnrDetectionEnabled` and `AndroidNativeAnrEnabled` separated.

The way this works is on API >= 30 `sentry-java` will use `ApplicationExitInfo` to read OS-reported ANRs to generate events out of those. Since this happens on the next launch, `enableScopePersistence` allows the SDK to store the scope to disk to then enrich the event with the scope at the time of the ANR. 

### Proposal

We can surface these options and pass them to `sentry-java`. These are especially valuable if users have their own launcher, or plugins running outside of the Unity game loop.
`sentry-java` takes care of picking the supported ANR integration. On API < 30 this is "just" a watchdog, similar to what we have for C#.

### Follow Up

There are a couple of follow-ups:
1. The `sentry-java` manifest needs to accept the historic anr capture flag, see https://github.com/getsentry/sentry-java/pull/5387
2. Improvements to our C# anr capture mechanism see https://github.com/getsentry/sentry-unity/issues/1961
3. Add ANR capture to e2e tests
4. Add native ANR trigger to samples

### DIFF
--- a/src/Sentry.Unity.Android/SentryJava.cs
+++ b/src/Sentry.Unity.Android/SentryJava.cs
@@ -139,6 +139,7 @@ internal class SentryJava : ISentryJava
                 androidOptions.Call("setEnableNdk", options.NdkIntegrationEnabled);
                 androidOptions.Call("setEnableScopeSync", options.NdkScopeSyncEnabled);
                 androidOptions.Call("setNativeSdkName", "sentry.native.android.unity");
+                androidOptions.Call("setAnrEnabled", options.AndroidAnrV2Enabled);
 
                 // Options that are not to be set by the user
                 // We're disabling some integrations as to not duplicate event or because the SDK relies on the .NET SDK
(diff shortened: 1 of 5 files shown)


src/Sentry.Unity.Android/SentryJava.cs
```diff
@@ -140,6 +140,11 @@ public void Init(SentryUnityOptions options)
                 androidOptions.Call("setEnableScopeSync", options.NdkScopeSyncEnabled);
                 androidOptions.Call("setNativeSdkName", "sentry.native.android.unity");
 
+                androidOptions.Call("setAnrEnabled", options.AndroidNativeAnrEnabled);
+                androidOptions.Call("setEnableScopePersistence", options.AndroidNativeAnrEnabled);
```
bitsandfoxes, line 144: The `﻿setEnableScopePersistence` only affects the ANR v2 and the tombstone integration. And the Unity SDK only surfaces the ANR with this PR. So I'm opting to keep this option merged into the `AnrNativeAnrEnabled`.

JoshuaMoelans (approved):

PR #2673, opened 2026-05-08
chore: update `setup-unity` and run macOS e2e tests

bitsandfoxes:
Follow up of https://github.com/getsentry/sentry-unity/pull/2665

We are using an action to install the Editor for Window and macOS. With the release of `arm` this already became brittle. The original action now has been archived in January. Since our fork is brittle and hard to maintain we're bumping to a stipped down version. 

This is now trying a feature branch where we iterated on how `setup-unity` works.


scripts/unity-versions.json
bitsandfoxes, line 1: using this to hold the unity version data. we provide the changeset now from here instead, so we don't need to rely on some brittle magic to figure out downloading those versions in the `setup-unity` action.

.github/workflows/test-build-windows.yml
```diff
@@ -24,12 +24,16 @@ jobs:
...
+          $u = (Get-Content scripts/unity-versions.json -Raw | ConvertFrom-Json).'${{ env.UNITY_VERSION }}'
+          "unityVersion=$($u.version)"     >> $env:GITHUB_OUTPUT
+          "unityChangeset=$($u.changeset)" >> $env:GITHUB_OUTPUT
 
       - name: Setup Unity
-        uses: getsentry/setup-unity@3bdc8c022b6d30ecf2d21d12a564bfa55a54fa2e
+        uses: getsentry/setup-unity@1bbd385ae3b0505069aacc56d47f0a16366f10b1
         with:
           unity-version: ${{ steps.env.outputs.unityVersion }}
+          unity-version-changeset: ${{ steps.env.outputs.unityChangeset }}
```
bitsandfoxes, line 36: providing it instead of trying some magic to find it.

.github/workflows/test-build-macos.yml
bitsandfoxes, line 1: macos e2e tests are back!

mujacica (approved):

PR #2680, opened 2026-05-13
chore: restructure dotnet build targets for local development

bitsandfoxes:
QOL feature for local development, so we (I) don't have to clean and build with every change.
Sample usage:
- `dotnet build` - builds the entire SDK stack including native sdks
- `dotnet msbuild /t:BuildAndroidSDK src/Sentry.Unity` cleans the output artifacts, and rebuilds `sentry-native` and `sentry-java`

#skip-changelog


build/local-dev.targets
bitsandfoxes, line 1: These are the targets used to build players for tests, or utility to help with setup.

build/native-sdks.targets
bitsandfoxes, line 1: These targets are intended to allow for iterative building, to overwrite the output inside of `package-dev/Plugins/` on each call. No more manual deletion before building.
The `Ensure` targets are hooked up to the regular build to allow calling `dotnet build` from the root and the repo automatically builds itself and its native layer.

scripts/build-native-ndk-local.ps1
bitsandfoxes, line 1: This is a helper to build `sentry-native` for `sentry-java` locally. These will get picked up from `sentry-java` via `mavenLocal`.

mujacica (approved):

PR #2692, opened 2026-05-21
feat: native crash backend as option for `Windows`

bitsandfoxes:
_wip_


JoshuaMoelans (approved):
lgtm! 

one small q for the native backend in general; we have [these three options](https://github.com/getsentry/sentry-native/blob/master/include/sentry.h#L1866-L1923) (`sentry_options_set_minidump_mode` , `sentry_options_set_crash_reporting_mode ` and `sentry_options_set_crash_upload_mode`) for the native backend, would it make sense to expose these in the Unity SDK too?

bitsandfoxes:
> one small q for the native backend in general; we have [these three options](https://github.com/getsentry/sentry-native/blob/master/include/sentry.h#L1866-L1923) (sentry_options_set_minidump_mode , sentry_options_set_crash_reporting_mode and sentry_options_set_crash_upload_mode) for the native backend, would it make sense to expose these in the Unity SDK too?

Fair point. I'll follow up on this.

PR #2709, opened 2026-06-09
feat: app hang capture on desktop

bitsandfoxes:
With https://github.com/getsentry/sentry-native/pull/1806 `sentry-native` gained the ability to capture app hangs in-proc for macOS, Linux, and Windows. 
The options controlling this feature got added as `experimental` and default to `false` right now.
Added the app hang capture to our e2e testing suite.


test/Scripts.Integration.Test/Scripts/IntegrationOptionsConfiguration.cs
```diff
@@ -41,6 +41,12 @@ public override void Configure(SentryUnityOptions options)
         // Disable ANR to avoid test interference
         options.DisableAnrIntegration();
 
+        // App-Hang detection. EnableAppHangTracking drives iOS (sentry-cocoa);
+        // EnableNativeAppHangTracking drives the desktop sentry-native path. Both share AppHangTimeout.
+        options.EnableAppHangTracking = true;
```
bitsandfoxes:
```suggestion
        options.Experimental.EnableNativeAppHangTracking = true;
```

CHANGELOG.md
```diff
@@ -26,6 +32,10 @@
   - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#351)
   - [diff](https://github.com/getsentry/sentry-cli/compare/3.5.0...3.5.1)
 
+## Unreleased
+
+- Extended `EnableAppHangTracking` (default `true`) and `AppHangTimeout` (default `5s`) to enable app hang detection via `sentry-native` on macOS, Windows, and Linux. For macOS this requires the backend to be switched to `native` in the Advanced -> Experimental settings. `sentry-native` monitors the main thread and produces a stack trace for the hang event. When enabled, the Unity SDK's C# watchdog is skipped to avoid duplicate reports. ([#2679](https://github.com/getsentry/sentry-unity/pull/2679))
+
```
JoshuaMoelans: This section is inside the 4.5.0 block (and points to the https://github.com/getsentry/sentry-unity/pull/2679 PR, not this one). Not sure if the wording of this paragraph is the preferred one over the other section in the top `## Unreleased`


```suggestion
```

src/Sentry.Unity.Native/SentryNative.cs
```diff
@@ -87,6 +87,22 @@ internal static void Configure(SentryUnityOptions options, RuntimePlatform platf
         }
         options.CrashedLastRun = () => crashedLastRun;
 
+        if (options.Experimental.EnableNativeAppHangTracking)
+        {
+            Logger?.LogDebug("Starting the app-hang heartbeat coroutine.");
+            SentryMonoBehaviour.Instance.StartAppHangHeartbeat(SentryNativeBridge.AppHangHeartbeat);
```
JoshuaMoelans, line 100: My clanker might be worrying a bit too much, but do we support/document re-init of the SDK anywhere? if so, the starting of the AppHangHeartbeat would lead to two coroutines staying alive (since sdk close doesn't close the coroutine, only if the actual app closes in full).

> Nit: StartAppHangHeartbeat returns a Coroutine handle but it's discarded — on Close+Re-init the old coroutine keeps running alongside a new one. Same pattern exists for the Quitting handler at line 59. Not a blocker since re-init isn't a supported flow, but storing and stopping the handle would be a cheap guard.

So we could keep the handle around & on sentry_close also stop the coroutine?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Added the experimental `options.Experimental.EnableNativeAppHangTracking` (default `false`) to enable app hang detection via `sentry-native` on macOS, Windows, and Linux. On macOS, this requires the crash reporting backend to be switched to `Native` in the Advanced -> Experimental settings. `sentry-native` monitors the main thread and produces an event including a stack trace for the hang, reusing the top-level `AppHangTimeout` (default `5s`). When effective, the Unity SDK's C# watchdog is skipped to avoid duplicate reports. iOS app hang detection remains controlled by the top-level `EnableAppHangTracking`. ([#2709](https://github.com/getsentry/sentry-unity/pull/2709))
```
JoshuaMoelans: (nit) slightly confusing wording on 'crash reporting backend'
```suggestion
- Added the experimental `options.Experimental.EnableNativeAppHangTracking` (default `false`) to enable app hang detection via `sentry-native` on macOS, Windows, and Linux. On macOS, this requires the macOS backend to be switched to `Native` instead of `Cocoa` in the Advanced -> Experimental settings. `sentry-native` monitors the main thread and produces an event including a stack trace for the hang, reusing the top-level `AppHangTimeout` (default `5s`). When effective, the Unity SDK's C# watchdog is skipped to avoid duplicate reports. iOS app hang detection remains controlled by the top-level `EnableAppHangTracking`. ([#2709](https://github.com/getsentry/sentry-unity/pull/2709))
```

JoshuaMoelans (changes requested):
A few comments, but looks good overall! Only softly blocking on the stubs & fixing the changelog \:)

---

In https://github.com/getsentry/sentry-unity/blob/feat/native-app-hang/src/Sentry.Unity.iOS/SentryNativeCocoa.cs#L45-L51 we disable the `AnrIntegration` if `EnableAppHangTracking` is true; should we do the same for macOS (it is currently gated behind `platform == RuntimePlatform.IPhonePlayer`

---

Do we need stubs for the new `sentry_options_set_enable_app_hang_tracking`, `sentry_options_set_app_hang_timeout`, and `sentry_app_hang_heartbeat` in https://github.com/getsentry/sentry-unity/blob/feat/native-app-hang/package-dev/Plugins/Switch/sentry_native_stubs.c ?

src/Sentry.Unity.Native/SentryNative.cs
bitsandfoxes: Fair enough. And easy enough. Done.

bitsandfoxes:
> In https://github.com/getsentry/sentry-unity/blob/feat/native-app-hang/src/Sentry.Unity.iOS/SentryNativeCocoa.cs#L45-L51 we disable the AnrIntegration if EnableAppHangTracking is true; should we do the same for macOS (it is currently gated behind platform == RuntimePlatform.IPhonePlayer

When targeting macOS and using Cocoa, then there is no ANR support other than the C# `AnrIntegration`. But something is brewing over at https://github.com/getsentry/sentry-cocoa/issues/8118 to improve on that. With the `sentry-native` on macOS, that code in `SentryNativeCocoa` never gets executed.

bitsandfoxes:
> Do we need stubs for the new sentry_options_set_enable_app_hang_tracking, sentry_options_set_app_hang_timeout, and sentry_app_hang_heartbeat in https://github.com/getsentry/sentry-unity/blob/feat/native-app-hang/package-dev/Plugins/Switch/sentry_native_stubs.c ?

Great catch! Totally missed that. Thanks!

JoshuaMoelans (approved):
LGTM, thanks for addressing all the comments 🫶

PR #2715, opened 2026-06-15
chore: Added docs to `CaptureFeedback`

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/2626

Because it is not obvious that `CaptureFeedbackResult` does not actually inform whether the feedback made it to Sentry.

#skip-changelog

### DIFF
--- a/src/Sentry.Unity/SentrySdk.Dotnet.cs
+++ b/src/Sentry.Unity/SentrySdk.Dotnet.cs
@@ -313,6 +313,10 @@ public static partial class SentrySdk
     /// <summary>
     /// Captures feedback from the user.
     /// </summary>
+    /// <param name="feedback">The feedback to send to Sentry.</param>
+    /// <param name="configureScope">The callback to configure the scope.</param>
+    /// <param name="hint">An optional hint providing high-level context for the source of the event.</param>
+    /// <returns>The Id of the captured feedback. Returns <see cref="SentryId.Empty"/> when capture failed.</returns>
     [DebuggerStepThrough]
     public static SentryId CaptureFeedback(SentryFeedback feedback, Action<Scope> configureScope, SentryHint? hint = null)
         => Sentry.SentrySdk.CaptureFeedback(feedback, configureScope, hint);
@@ -328,6 +332,7 @@ public static partial class SentrySdk
     /// </param>
     /// <param name="configureScope">The callback to configure the scope.</param>
     /// <param name="hint">An optional hint providing high-level context for the source of the event.</param>
+    /// <returns>The Id of the captured feedback. Returns <see cref="SentryId.Empty"/> when capture failed.</returns>
     /// <remarks>
     /// Capturing is asynchronous and non-blocking. The envelope is handed to a background worker. Non-success 
     /// results inform only on whether the capture succeeded or failed on the client, i.e. empty message, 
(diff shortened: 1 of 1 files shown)


src/Sentry.Unity/SentrySdk.Dotnet.cs
```diff
@@ -320,6 +320,19 @@ public static SentryId CaptureFeedback(SentryFeedback feedback, Action<Scope> co
...
+    /// The outcome of capturing the feedback. <see cref="CaptureFeedbackResult.Success"/> means the
+    /// feedback was accepted and handed to the transport. It passed validation and event processors and
+    /// was queued for delivery. It does <b>not</b> confirm that the feedback was transmitted to Sentry.
+    /// </param>
+    /// <param name="configureScope">The callback to configure the scope.</param>
+    /// <param name="hint">An optional hint providing high-level context for the source of the event.</param>
+    /// <remarks>
+    /// Capturing is asynchronous and non-blocking. The envelope is handed to a background worker. Non-success 
+    /// results inform only on whether the capture succeeded or failed on the client, i.e. empty message, 
+    /// disabled SDK, dropped by an event processor.
```
Flash0ver, line 339: question: consistent wording

Here, we mention that the feedback is "handed to the transport".
In the `<remarks>` below, we say that "The envelope is handed to a background worker".

Is this intended?
Or should we chose either transport or worker to then use in both places?

Or should we actually mention both, first the background worker and then the transport ... but that could be a case of "too much (implementation) details" (at this level).

src/Sentry.Unity/SentrySdk.Dotnet.cs
```diff
@@ -335,6 +348,19 @@ public static SentryId CaptureFeedback(SentryFeedback feedback, Scope? scope = n
     /// <summary>
     /// Captures feedback from the user.
     /// </summary>
+    /// <param name="feedback">The feedback to send to Sentry.</param>
+    /// <param name="result">
```
Flash0ver, line 361: question: consistent overloads

Should the overload, that does not have a `out CaptureFeedbackResult` parameter also get similar XML doc comments for consistency?

```CSharp
    /// <summary>
    /// Captures feedback from the user.
    /// </summary>
    [DebuggerStepThrough]
    public static SentryId CaptureFeedback(SentryFeedback feedback, Action<Scope> configureScope, SentryHint? hint = null)
        => Sentry.SentrySdk.CaptureFeedback(feedback, configureScope, hint);
```

src/Sentry.Unity/SentrySdk.Dotnet.cs
```diff
@@ -335,6 +348,19 @@ public static SentryId CaptureFeedback(SentryFeedback feedback, Scope? scope = n
...
+    /// The outcome of capturing the feedback. <see cref="CaptureFeedbackResult.Success"/> means the
+    /// feedback was accepted and handed to the transport. It passed validation and event processors and
+    /// was queued for delivery. It does <b>not</b> confirm that the feedback was transmitted to Sentry.
+    /// </param>
+    /// <param name="scope">The scope to apply to the feedback.</param>
+    /// <param name="hint">An optional hint providing high-level context for the source of the event.</param>
+    /// <remarks>
```
Flash0ver, line 369: question: completeness

Should we also add a `<returns>` tag, that describes that `SentryId.Empty` is returned on non-success?

src/Sentry.Unity/SentrySdk.Dotnet.cs
bitsandfoxes: I don't understand. One thing is explaining why `Success` might not be what users expect. The other is a remark providing  additional information as to why capture might not behave as expected.

Flash0ver: oh ... I was referring to the text, where we mention the feedback being (not) successfully handed off to the `"transport"` in one spot, and the `"background worker"` in another ... I was wondering if we should use either `"transport"` or `"background worker"` in both locations ... but both are actually correct ... so there isn't really an "issue" with it

Flash0ver (approved):

PR #2721, opened 2026-06-17
feat: Support Unity `6.5`

bitsandfoxes:
#skip-changelog


test/Scripts.Integration.Test/DependencyConflict/package/Runtime/DependencyConflict.dll.meta
```diff
@@ -0,0 +1,70 @@
...
+  isOverridable: 1
+  isExplicitlyReferenced: 0
+  validateReferences: 1
+  platformData:
+    Android:
+      enabled: 1
+      settings:
+        AndroidLibraryDependee: UnityLibrary
+        AndroidSharedLibraryType: Executable
+        CPU: ARMv7
```
pdyatlov: Just wanted to note that ARMv7 here tells Unity to treat the managed DLL as a native ARMv7 library, which can fail on an arm64 emulator (API 30), which is probably causing the Android tests to fail

.github/workflows/create-unity-matrix.yml
```diff
@@ -14,9 +14,10 @@ on:
 
 env:
   # Unity versions used in PRs
-  PR_UNITY_VERSIONS: '["2022.3", "6000.0", "6000.3"]'
+  # TEMPORARY: reduced to 6000.5 only to probe Unity 6.5 CI on this branch
+  PR_UNITY_VERSIONS: '["6000.5"]'
   # Unity versions used on main branch
-  FULL_UNITY_VERSIONS: '["2021.3", "2022.3", "6000.0", "6000.3"]'
+  FULL_UNITY_VERSIONS: '["6000.5"]'
```
pdyatlov, line 20: Just a quick comment, should this change be merged into main?

test/Scripts.Integration.Test/DependencyConflict/package/Runtime/DependencyConflict.dll.meta
bitsandfoxes: Ah, great catch! Thanks for pointing that out!

.github/workflows/create-unity-matrix.yml
bitsandfoxes: No, definitely not. This is just while it's in a draft state to cut down on CI time.

bitsandfoxes:
superseded by https://github.com/getsentry/sentry-unity/pull/2727 and https://github.com/getsentry/sentry-unity/pull/2726

Issue #2717, opened 2026-06-16
IL2CPP Failed on 6000.5.xx build


zhiyan114:
### SDK Version

4.4.0

### Unity Version

6000.5.0f1

### Platform

- [x] Editor
- [ ] Mobile (Android, iOS)
- [x] Desktop (Windows, macOS, Linux)
- [x] WebGL
- [ ] Console

### Sentry Environment

Sentry SaaS (sentry.io)

### Other Error Monitoring Solution

No

### Steps to Reproduce

1. Setup a unity project or upgrade existing version
2. Install/Upgrade latest unity sentry package
3. Build project as IL2CPP (tested with Win IL2CPP and WebGL)

### Expected vs Actual Result

Expected:

Compilation works as unity 6.4.xxx builds

------------------------------------------------
Actual:

Compilation failed, related to Sentry.System.Memory.dll

Sample outout:
```
[...]

PR #2726, opened 2026-06-23
fix: dependency resolution via `assembly alias` for .NET Standard 2.1

### LINKED ISSUE #1777: Assembly Alias causes  `System.Text.Json` exception
When building the SDK with TFM `netstandard2.1` (i.e. using Unity 2022 or newer) and then running the aliasing tool causes 
```
TypeLoadException: VTable setup of type System.Text.Json.Utf8JsonWriter failed
```
when the SDK attempts to serialize an envelope.

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/1777
Relates to the Unity 6.5 update and https://github.com/getsentry/sentry-unity/issues/2717
Builds on top of https://github.com/getsentry/dotnet-assembly-alias/pull/24, hence the update to `0.5.0`.

Unity itself already resolves any of the `BCL`. So we exclude them from being aliased (renamed). The renaming forces Unity to keep them, causing duplication of i.e. `<Span>`, resulting in build failures such as

```
System.InvalidOperationException: Sequence contains no matching element
```

To make sure these changes to our dependency resolution do not break with other packages I added a `DependencyConflictPackage` that brings similar dependencies that are unaliased.


scripts/repack.ps1
```diff
@@ -7,8 +7,7 @@ if (Test-Path "package-release.zip") {
     Remove-Item -Path "package-release.zip" -Force
 }
 
-assemblyalias --target-directory "package-dev/Runtime" --internalize --prefix "Sentry." --assemblies-to-alias "Microsoft*;System*"
-assemblyalias --target-directory "package-dev/Editor" --internalize --prefix "Sentry." --assemblies-to-alias "Microsoft*;Mono.Cecil*"
+& "$PSScriptRoot/alias-assemblies.ps1"
```
JoshuaMoelans, line 12: q: should we check the return (ExitCode) value here before continuing?

test/IntegrationTest/CommonTestCases.ps1
```diff
@@ -72,10 +86,6 @@ $CommonTestCases = @(
             $SentryEvent.breadcrumbs.values | Should -Not -BeNullOrEmpty
             $SentryEvent.breadcrumbs.values | Where-Object { $_.message -eq "Integration test started" } | Should -Not -BeNullOrEmpty
             $SentryEvent.breadcrumbs.values | Where-Object { $_.message -eq "Context configuration finished" } | Should -Not -BeNullOrEmpty
-
-            $dataCrumb = $SentryEvent.breadcrumbs.values | Where-Object { $_.message -eq "Context configuration finished" }
-            $dataCrumb | Should -Not -BeNullOrEmpty
-            $dataCrumb.data.integration_test_key | Should -Be "integration_test_value"
```
JoshuaMoelans, line 78: q: not quite sure why this was removed (here and in `IntegrationTester.cs`)

JoshuaMoelans (approved):
lgtm, just two q's remaining

test/IntegrationTest/CommonTestCases.ps1
bitsandfoxes: Great catch! I messed up somewhere when merging/moving things around. Thanks!

PR #2753, opened 2026-07-06
fix: gracefully degrade with no native SDK on `Switch`

bitsandfoxes:
Without this the SDK silently swallows the error, calling into the stub with not even logs telling what's going on.


JoshuaMoelans (approved):
🚀 🚢

Issue #2751, opened 2026-07-03
[Regression] CS0012: ReadOnlySpan (System.Memory) missing in Editor causes compilation failure for SentrySdk.Metrics in 4.6.0


gal20040:
### SDK Version

4.6.0

### Unity Version

6000.3.15f1

### Platform

- [x] Editor
- [ ] Mobile (Android, iOS)
- [ ] Desktop (Windows, macOS, Linux)
- [ ] WebGL
- [ ] Console

### Sentry Environment

Sentry SaaS (sentry.io)

### Other Error Monitoring Solution

Yes

### Description
After upgrading Sentry Unity SDK `4.5.0` -> `4.6.0`, compiling any script that interacts with `SentrySdk.Metrics` (e.g., `EmitCounter`, `EmitGauge`) throws a **CS0012** compilation error in the Unity Editor: `error CS0012: The type 'ReadOnlySpan<>' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Memory, Version=4.0.1.2, Culture=neutral, PublicKeyToken=null'.`

This seems to be a regression caused by the recent assembly aliasing removal (PR #2726 / Issue #2717). While fixing IL2CPP issues, `System.Memory.dll` was left disabled for the `Editor` platform in its `.meta` file. Because `SentrySdk.Metrics` methods contain overloads/signatures that require `ReadOnlySpan<>`, the C# compiler fails to resolve the overload resolution candidates during Editor compilation, explicitly demanding `System.Memory, Version=4.0.1.2`.

### Steps to Reproduce
1. Install Sentry Unity SDK `4.6.0` in a Unity project, `Packages\manifest.json`:
   ```json
   {
     "dependencies": {
       ...
       "io.sentry.unity": "https://github.com/getsentry/unity.git#4.6.0",
       ...
     }
   }
   ```
3. Create a custom `.asmdef` that references `io.sentry.unity.runtime`.
4. Create a C# script inside this assembly and call a metric method:
   ```csharp
   SentrySdk.Metrics.EmitCounter("test_counter", 1.0, new[] { new KeyValuePair<string, object>("key", "value") });
   ```
4. Switch back to Unity to trigger compilation.

### Environment

- Sentry Unity SDK Version: 4.6.0
- API Compatibility Level: .NET Standard 2.1
- Platform: Unity Editor (Windows)
- Unity: 6000.3.15f1 (c1aa84e375f6)

### Expected Behavior

[...]

PR #2754, opened 2026-07-06
fix: resolved `ReadOnlySpan` missing

bitsandfoxes:
Addresses https://github.com/getsentry/sentry-unity/issues/2751

We excluded those libraries from being aliased in https://github.com/getsentry/sentry-unity/pull/2726.
This had the side effect of also setting the `PublicKeyToken` to `null`. When users would call [`EmitCounter<T>(string name, T value, ReadOnlySpan<KeyValuePair<string, object>> attributes).`](https://github.com/getsentry/sentry-dotnet/blob/04a98549b92420bc08f3505311ba3d0b6060262c/src/Sentry/SentryMetricEmitter.Counter.cs#L53) this would cause 
```
error CS0012: The type 'ReadOnlySpan<>' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Memory, Version=4.0.1.2, Culture=neutral, PublicKeyToken=null'.
```
Since we're already not aliasing those assemblies and we let Unity handle the resolution, we can simply rely on what Unity already comes with.


JoshuaMoelans (approved):
🚢

PR #2764, opened 2026-07-16
fix: sync environment to native SDKs

bitsandfoxes:
This builds and relies on https://github.com/getsentry/sentry-dotnet/pull/5365
And secondarily on https://github.com/getsentry/sentry-java/issues/5769

The idea is to be able to `scope.SetEnvironment("whatever_it_is)` at runtime and have that be synched across all layers.


JoshuaMoelans (approved):
LGTM 🚀

PR #2793, opened 2026-07-31
fix: cleanup stale native sdk during build

bitsandfoxes:
Fixes an issue when switching native backend from cocoa to native or the other way around.


bitsandfoxes:
got gaslit..

PR #2798, opened 2026-08-04
ci: Replace secrets: inherit with explicit secret maps

oioki:
## Summary
- 15 locations used `secrets: inherit`, exposing every repository secret to called reusable workflows that only need a small named subset.
- 14 locations in `ci.yml` call internal reusable workflows (`build.yml`, `test-create.yml`, `test-build-*.yml`, `test-run-*.yml`) that read `UNITY_LICENSE_SERVER_CONFIG`, `SENTRY_AUTH_TOKEN`, and/or `SENTRY_TEST_DSN` but never declared these under `workflow_call.secrets`. Added explicit secret schemas to each called workflow and updated `ci.yml` to pass only the secrets each workflow actually needs, by name.
- `changelog-preview.yml` (the 15th location) is disabled on GitHub (`disabled_manually`) and unused — removed the workflow entirely instead of fixing its secret handling.

## Test plan
- [x] Validated all touched workflow YAML parses correctly
- [x] Cross-referenced every `secrets.*` reference in each called workflow against the new explicit `secrets:` maps in `ci.yml` to ensure nothing was dropped
- [x] Confirmed `changelog-preview.yml` workflow state is `disabled_manually` via GitHub API before deleting
- [x] Confirm CI runs (build, test-create, android/ios/webgl/linux/windows/macos build+run jobs) still succeed with the explicit secret passing

Fixes VULN-2345

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bitsandfoxes:
Thanks!

bitsandfoxes (approved):

PR #2809, opened 2026-08-11
chore: local test running improvements

bitsandfoxes:
#skip-changelog

### DIFF
--- a/scripts/unity-test-pipeline.ps1
+++ b/scripts/unity-test-pipeline.ps1
@@ -73,7 +73,7 @@ function Get-TestSummary([string] $requestedMode, [object] $result) {
             }
         })
 
-    $success = $summary.Total -gt 0 -and $summary.Failed -eq 0 -and $summary.Inconclusive -eq 0
+    $success = $summary.Total -gt 0 -and $summary.Failed -eq 0
     [pscustomobject]@{
         Mode         = $result.Mode
         Duration     = $result.Duration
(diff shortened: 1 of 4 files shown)


scripts/unity-test-pipeline.ps1
```diff
@@ -109,7 +109,7 @@ function Invoke-UnityPipelineTests(
         }
 
         if ($testPlatform -eq "playmode") {
-            $started = Invoke-UnityPipelineCommand -ProjectPath $projectPath -Command "run_tests" -CommandArguments ($commandArguments + @("--async_tests", "true"))
+            $started = Invoke-UnityPipelineCommand -ProjectPath $projectPath -Command "run_tests" -CommandArguments ($commandArguments + @("--async_tests", "true")) -TimeoutSeconds ($timeoutSeconds + 30)
```
bitsandfoxes, line 112: Timeout was missing.

src/Sentry.Unity/Integrations/AnrIntegration.cs
```diff
@@ -195,8 +195,11 @@ internal AnrWatchDogSingleThreaded(IDiagnosticLogger? logger, SentryMonoBehaviou
             logger?.LogDebug("Stopping ANR detection coroutine.");
             _watch.Stop();
 
-            MonoBehaviour.StopCoroutine(_updateUiStatusCoroutine);
-            _updateUiStatusCoroutine = null;
+            if (_updateUiStatusCoroutine is not null)
```
bitsandfoxes, line 198: Not a real but but following the established pattern from below. If triggering the putting-on-background twice this would throw. It's a non-existing issue on actual builds in an already deprecated feature.

tustanivsky (approved):

PR #2817, opened 2026-08-20
fix: unity 2022 android line numbers

bitsandfoxes:
`lld` puts notes with different alignments in separate `PT_NOTE` segments. NDK r23 (Unity 2022.3) ships an x86_64 `crtbegin_so.o` whose `.note.android.ident` has `sh_addralign = 2`, while `.note.gnu.build-id = 4` — so they split, and the build-id ends up in the segment IL2CPP never looks at.

Instead, if the IL2CPP processor fails to get the `ImageUuid` it falls back to finding the image by name provided by sentry-native.

Successful run via Unity 2022: https://github.com/getsentry/sentry-unity/actions/runs/32381917935

### DIFF
--- a/test/IntegrationTest/Integration.Tests.ps1
+++ b/test/IntegrationTest/Integration.Tests.ps1
@@ -132,6 +132,91 @@ BeforeAll {
         return $runResult
     }
 
+    # Read a property that may be missing from the fetched event. StrictMode turns a missing property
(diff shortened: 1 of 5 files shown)


test/IntegrationTest/Integration.Tests.ps1
bitsandfoxes, line 1: qol for future debugging

JoshuaMoelans (approved):

Issue #2818, opened 2026-08-24
crash report don't work on windows due to dll name conflict


lgarczyn:
### SDK Version

4.5

### Unity Version

2022.3.62f2

### Platform

- [ ] Editor
- [ ] Mobile (Android, iOS)
- [x] Desktop (Windows, macOS, Linux)
- [ ] WebGL
- [ ] Console

### Sentry Environment

Sentry SaaS (sentry.io)

### Other Error Monitoring Solution

No

### Steps to Reproduce

Native crash capture has never worked for us in a Windows build. Users report them by hand.

A Mono build ships two files named sentry.dll. The managed Sentry SDK lands in
Managed/, the real native library sits at the build root. Mono probes the calling
assembly's own folder first, and Sentry.Unity.Native lives in Managed/, so
DllImport("sentry") binds to the managed assembly and then dies looking for a C
function that isn't there.

That surfaces as EntryPointNotFoundException rather than DllNotFoundException.
The dll loads, it is just the wrong dll. Nothing else in our Sentry setup is wrong.

We never saw it because sentry-unity 2.4.0 stopped sending native init failures
as events and started logging them to a logger that is null unless Debug is on.

Renaming the dll fixes the issue.

SDK 4.9.0 has the same bug, the
experimental Windows backend ships the same filename to the same folder, Mono
dllmap is ignored, and preloading the native dll with LoadLibraryW does not stop
Mono opening the managed one.

### Expected vs Actual Result

Expected: crash report
Result: no crash report

lgarczyn:
https://github.com/getsentry/sentry-unity/issues/1864

Seems to be the same issue

bitsandfoxes:
Thanks for reaching out and the detailed description. Let me take a look at the PR.

PR #2819, opened 2026-08-24
Windows mono native dll clash

lgarczyn:
Fixed native crash capture on Windows with the Mono scripting backend. The native library is now copied into the player as `sentry-native`, so `DllImport` can no longer resolve to the managed `Sentry.dll` sitting beside it in `Managed/` https://github.com/getsentry/sentry-unity/issues/2818

Built with fable, if you have a better idea of how to fix this, I would not complain

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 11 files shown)


bitsandfoxes:
Hey @lgarczyn, thanks for the contribution. I'll need to figure out a way to make sure this also works with our console support but other than that it mostly looks good to me.

lgarczyn:
> Hey @lgarczyn, thanks for the contribution. I'll need to figure out a way to make sure this also works with our console support but other than that it mostly looks good to me. 

No worries! We are using our fork for now, hasn't reoccured yet

bitsandfoxes:
I think this will work great. I'll continue in https://github.com/getsentry/sentry-unity/pull/2842
Thanks again for this, much appreciated! 🙏

PR #2831, opened 2026-09-04
feat(switch2): Nintendo Switch 2 support, and stop the offline transport raising system dialogs

bitsandfoxes:
Brings up Nintendo Switch 2, and fixes three defects found while getting it working on a devkit. Each commit stands alone.

## The reported problem

A Switch game that is offline showed the system "connect to the internet" dialog every couple of seconds, indefinitely.

`UnityWebRequestTransport` opens a connection per envelope, and on Switch the attempt itself is what raises the dialog. With logs and metrics both flushing on a 5s timer, that is a dialog every ~2.5s. A connection error only logged a warning and returned `null`, which also skipped the rate-limit handling in `HttpTransportBase` — so nothing anywhere recorded that the network was down, and the next envelope tried again immediately.

## What's here

**`fix(switch)`: transport backoff** — skip sending while the platform reports no network, and back off exponentially (1s→60s) after a connection error, resetting on the first success. Both paths record `DiscardReason.NetworkError`, so drops still show up in client reports instead of vanishing.

**`fix(switch)`: diagnostic logger no longer throws while logging** — this one bit hard. With `DisableFileWrite`, resolving the installation ID falls through to `NetworkInterface.GetAllNetworkInterfaces()`, whose static initializer throws on Switch. `Hub.ConfigureScope` caught it and logged it — but stringifying *that* exception threw again, escaping the handler and aborting SDK init partway through registering integrations. Every Unity integration after `GlobalRootScopeIntegration` was left unregistered, so nothing was captured and no envelope was ever produced. A logger must never turn a handled error into a fatal one.

**`fix(switch2)` + `feat(switch2)`: Switch 2 support** — `UNITY_SWITCH2` selects `SentryNativeSwitch`, both asmdefs include the platform, and it gets its own plugin directory with no-op stubs. `IsNativeSupportEnabled` had no `Switch2` case, so it fell through to the default and returned `false` — native support could never initialize regardless of configuration. Switch 2 shares `SwitchNativeSupportEnabled` and the existing Switch assembly, which is shipped to it via plugin metadata rather than by building a second assembly.

**`fix(switch)`: ask the platform for network availability** — see below.

## Why reachability isn't enough

`Application.internetReachability` **reports a Switch console as reachable while it is offline**. From the device:

```
Failed to send request: No Internet Connection.
Reachability reported as ReachableViaLocalAreaNetwork.
```

So the reachability gate was inert on the one platform it was written for, and the backoff absorbed everything — which reduced the dialogs to one per minute but could not remove them, because each retry is still an attempt.

sentry-switch now exposes `sentry_switch_utils_is_network_available()`, which asks the console's network interface manager **without submitting a network request** — so asking costs nothing and shows no dialog. `SentryNativeSwitch` supplies it as `NetworkAvailabilityProbe` and the transport prefers it, falling back to reachability when no probe is set, leaving WebGL and the unknown-platform path unchanged.

## Verified on a Switch 2 devkit

Offline, with the probe in place:

```
Sentry: (Debug) Using the native SDK to determine network availability.
Sentry: (Debug) No network available. Dropping envelope instead of attempting to send.
```

No `Failed to send request`, no backoff, **no dialogs** — no `UnityWebRequest` is ever constructed, so nothing can raise one. Native init completes and scope sync is syncing breadcrumbs, user and trace.

## Merge order

Depends on getsentry/sentry-switch#188, which must merge and ship first. The P/Invoke goes through `__Internal` against a statically linked archive, so it resolves at **link time** — a Switch player built against an older `libsentry.a` fails to link rather than falling back. The bundled stubs only cover the no-native-library case.

## Not covered here

Two related issues left deliberately out of scope: `SentryNativeBridge` never sets `network_connect_func`, so the native transport does not start and native crash reports depend on Nintendo's upload path; and `Hub`'s integration loop has no per-integration `try`/`catch`, which is why a single throwing `Register` could decapitate the SDK.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ABHTqM37qGAc9iQZcdbdY8


bitsandfoxes:
Split into three focused PRs, since these are independent changes that happened to be found together:

- #2832 — `UnityLogger` no longer throws while logging
- #2833 — offline sends no longer raise the system connection dialog
- #2834 — Nintendo Switch 2 support

Nothing is dropped; the combined branch is superseded by those three.

PR #2834, opened 2026-09-04
feat: add Nintendo Switch 2 support

bitsandfoxes:
Brings up Nintendo Switch 2 by routing it through the existing Switch implementation rather than duplicating it.

## What

- Bundle `UNITY_SWITCH2` into `SENTRY_NATIVE_SWITCH` as well.
- Flip from platform-include to platform-exclude so old Unity versions without the target don't break.
- `IsNativeSupportEnabled` maps the platform to `SwitchNativeSupportEnabled`, Switch 2 shares the option.
- Enables automatic debug symbol upload
- The existing Switch native assembly and the existing stub are both shipped to Switch 2 via their plugin metadata.
- `SwitchNativePluginBuildPreProcess` now runs for both targets, resolves required libraries per platform (`Assets/Plugins/Sentry/{Switch,Switch2}/`), and toggles stub compatibility for whichever target is building.

## Why

Before the `IsNativeSupportEnabled` mapping, `SentryNativeSwitch.Configure` was reached but bailed out of immediately:

```
Sentry: (Debug) Native support is disabled for 'Switch2'.
```

The switch expression had no `Switch2` case, so it fell through to the default and returned `false`. Native support could never initialize regardless of how the project was configured. After the fix the full sequence runs — temporary storage mounts, the native SDK initializes, and scope sync starts syncing breadcrumbs, user and trace.

## No duplicated assembly or stub

Both are gated by plugin metadata, which is what decides where they ship. Both utilize the same stub in case of no native SDK being found.

## Matching by name

`RuntimePlatform.Switch2` and `BuildTarget.Switch2` only exist in Unity **6000.3 and newer**. So instead of relying on the enum being preset we do a comparison by name for `Switch2` instead.

## Verified on a Switch 2 devkit

Native initialization completes and scope sync is live.


JoshuaMoelans:
we should release switch before this merges probably 🤔

JoshuaMoelans (approved):

Issue #2845, opened 2026-09-11
Unable to load DLL 'sentry' on Windows x86


kkopanski:
### SDK Version

4.10.0

### Unity Version

6000.0.83

### Platform

- [ ] Editor
- [ ] Mobile (Android, iOS)
- [x] Desktop (Windows, macOS, Linux)
- [ ] WebGL
- [ ] Console

### Sentry Environment

Sentry SaaS (sentry.io)

### Other Error Monitoring Solution

No

### Steps to Reproduce

1. Build project for WindowsStandalone (x86 version)
2. Run app

### Expected vs Actual Result

Expected: App starts without any errors
Actual: Immediately after start you can observe an error message in log

> Sentry: (Error) Sentry native initialization failed - native crashes are not captured. System.DllNotFoundException: Unable to load DLL 'sentry'. Tried the load the following dynamic libraries: Unable to load dynamic library 'sentry' because of 'C:\build\output\unity\unity\External\il2cpp\builds\external\baselib\Source\WinApi\Baselib_DynamicLibrary_WinApi.inl.h(36):Baselib_DynamicLibrary_OpenUtf16: Failed to open the requested dynamic library (0x06000000) - %1 nie jest prawidłową aplikacją systemu Win32.  (WinError:000000c1)
>   at Sentry.Unity.Native.SentryNativeBridge.Init (Sentry.Unity.SentryUnityOptions options) [0x00088] in \sentry-unity\src\Sentry.Unity.Native\SentryNativeBridge.cs:50 
>   at Sentry.Unity.Native.SentryNative.Configure (Sentry.Unity.SentryUnityOptions options, UnityEngine.RuntimePlatform platform) [0x00069] in \sentry-unity\src\Sentry.Unity.Native\SentryNative.cs:48 
>   at Sentry.Unity.Native.SentryNative.Configure (Sentry.Unity.SentryUnityOptions options) [0x0000b] in \sentry-unity\src\Sentry.Unity.Native\SentryNative.cs:25 
>   at Sentry.Unity.SentrySdk.Init (Sentry.Unity.SentryUnityOptions options) [0x00020] in \sentry-unity\src\Sentry.Unity\SentrySdk.cs:42 
>   at Sentry.Unity.SentryInitialization.Init () [0x0001f] in .\Library\PackageCache\io.sentry.unity@85c72ef4a503\Runtime\SentryInitialization.cs:84 

I have already tried both backends with:
[...]

PR #2847, opened 2026-09-11
fix: x86 support for Windows

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-unity/issues/2845 and is somewhat of a follow-up on https://github.com/getsentry/sentry-unity/pull/2514
This adds the missing x86 native support for Windows by extending the package with an x86 build of sentry-native.

We're going for Windows-only here. Unity marked the Linux x86 and Linux Universal build targets as deprecated in `2018.3`.


mujacica (approved):

src/Sentry.Unity.Native/CFunctions.cs
```diff
@@ -87,90 +87,90 @@ internal static void SetValueIfNotNull(sentry_value_t obj, string key, long? val
...
+    [DllImport(SentryLib, CallingConvention = CallingConvention.Cdecl)]
     internal static extern void sentry_set_environment(string? environment);
 
-    [DllImport(SentryLib)]
+    [DllImport(SentryLib, CallingConvention = CallingConvention.Cdecl)]
     internal static extern IntPtr sentry_attach_file(string path);
 
-    [DllImport(SentryLib)]
+    [DllImport(SentryLib, CallingConvention = CallingConvention.Cdecl)]
     internal static extern IntPtr sentry_attach_bytes(byte[] buf, UIntPtr buf_len, string filename);
```
tustanivsky: In sentry-native 0.17.0 `sentry_attach_file/bytes` return `sentry_uuid_t` instead of a pointer (https://github.com/getsentry/sentry-native/pull/1974) so declaring `IntPtr` here  no longer matches.

src/Sentry.Unity.Native/CFunctions.cs
```diff
@@ -87,90 +87,90 @@ internal static void SetValueIfNotNull(sentry_value_t obj, string key, long? val
...
 
-    [DllImport(SentryLib)]
-    internal static extern IntPtr sentry_attach_file(string path);
+    [DllImport(SentryLib, CallingConvention = CallingConvention.Cdecl)]
+    internal static extern sentry_uuid_t sentry_attach_file(string path);
 
-    [DllImport(SentryLib)]
-    internal static extern IntPtr sentry_attach_bytes(byte[] buf, UIntPtr buf_len, string filename);
+    [DllImport(SentryLib, CallingConvention = CallingConvention.Cdecl)]
+    internal static extern sentry_uuid_t sentry_attach_bytes(byte[] buf, UIntPtr buf_len, string filename);
```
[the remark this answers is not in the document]
tustanivsky: sentry-native 0.17.0 changed `sentry_attach_file` and `sentry_attach_bytes` to return `sentry_uuid_t` by value instead of a pointer so the struct return type is correct.

src/Sentry.Unity.Native/CFunctions.cs
tustanivsky: Should be addressed in https://github.com/getsentry/sentry-unity/pull/2847/commits/ae3ef44f1c9e1faa43dc473ec70de209184303de

tustanivsky (approved):

PR #2848, opened 2026-09-11
chore: pin platform target for all plugins

bitsandfoxes:
We keep running into issues where plugins are enabled for platforms where they have no business running at. Or where they should be running but are not. This is partly due to the split between `package-dev/` and `package/`.

This has me doubly nervous since https://github.com/getsentry/sentry-unity/pull/2834. We had to flip the explicit include into an explicit exclude to make it work for older Unity versions that do not have the Switch 2 as an valid platform. 

So with this we have a hardcoded table setting the expectations of every plugin within the SDK, preventing future regressions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,10 @@ jobs:
         with:
           name: package-release
 
+      - name: Check plugin platform scopes
+        shell: pwsh
+        run: ./test/Scripts.Tests/test-plugin-platforms.ps1
+
       - name: Check snapshot
         id: snapshot-check
         shell: pwsh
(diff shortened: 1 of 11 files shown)


package-dev/Plugins/Switch/SentryStub~/sentry_native_stubs.c
```diff
@@ -352,9 +348,10 @@ int sentry_clear_crashed_last_run(void)
     return 0;
 }
 
-void sentry_reinstall_backend(void)
```
tustanivsky, line 355: Could it be a problem that [SentryNativeBridge's](https://github.com/getsentry/sentry-unity/blob/ffc8b6d4dfd09151f74f2a78c7d06e351e62a8dd/src/Sentry.Unity.Native/SentryNativeBridge.cs#L380) declaration of `sentry_reinstall_backend` still returns `void` instead of `int`?

tustanivsky (approved):

PR #2853, opened 2026-09-16
feat: make the crashpad handler path configurable

mollstam:
sentry-unity never calls `sentry_options_set_handler_path`, so sentry-native falls back to looking for `crashpad_handler.exe` next to the game executable.

Player use eg MelonLoader to load mods into the game, it hooks by placing itself as `version.dll` in game root directory which causes the Unity runtime to load it on launch (it's looking for `version.dll` in the Windows system dir but local copy takes precedence). This is naturally risky code and may crash the game, at which point the Sentry crashpad triggers. If `crashpad_handler.exe` is in the game root directory (which it currently has to be) it will also load the injected `version.dll` into the handler itself. It causes crashpad to crash leaving the process suspended and no crash report delivered.

Binding `handler_path` lets the handler be staged outside the game root. Moving ours to `<Game>_Data/Plugins/x86_64/` restored native crash capture on modded installs.

Not sure about the changelog additions etc, let me know if you need anything changed. 🙇🏻


src/Sentry.Unity.Native/SentryNativeBridge.cs
```diff
@@ -101,6 +101,24 @@ is RuntimePlatform.WindowsPlayer or RuntimePlatform.WindowsServer
         }
 #endif
 
+#if !SENTRY_NATIVE_SWITCH
+        // Left unset, sentry-native looks for the handler next to the running executable.
+        // Switch does not run an out-of-process handler, so the option does not apply there.
+        if (options.NativeHandlerPath is not null)
```
bitsandfoxes, line 107:
```suggestion
                if (!string.IsNullOrWhiteSpace(options.NativeHandlerPath))
```
Ideally, we'd actually check for the file to be there. But that would require a change to the underlying submodule so it's fine for now.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Added the `NativeHandlerPath` option. It sets `handler_path` on the native SDK, so the out-of-process crash handler can be placed somewhere other than next to the player executable. Left unset, sentry-native keeps looking for the handler next to the executable as before.
```
```suggestion
- Added the `NativeHandlerPath` option. It sets `handler_path` on the native SDK, so the out-of-process crash handler can be placed somewhere other than next to the player executable. Left unset, sentry-native keeps looking for the handler next to the executable as before. ([#2853](https://github.com/getsentry/sentry-unity/pull/2853))
```

bitsandfoxes (changes requested):
Thanks for the contribution! Other than those two minor tweaks this looks like it's good to go!

